PR #3, opened 2022-11-11
integrate mask in flowmatch loss

nikita-0209:


gflownet.py
```diff
@@ -782,6 +805,7 @@ def flowmatch_loss(self, it, batch):
         else:
             # TODO: potentially mask invalid actions next_q
             next_q = self.model(sp)
+        next_q = torch.where(masks == 0, next_q, -loginf)
```
alexhernandezgarcia, line 808: Would it not be `masks == 1`? We want to make `-loginf` the invalid (mask == True) actions.

gflownet.py
```diff
@@ -1238,13 +1262,21 @@ def logq(path_list, actions_list, model, env):
...
+        masks = tf(
+            [
+                env.get_mask_invalid_actions(path[idx], done[idx])
+                for idx in range(len(path))
+            ]
+        )
         with torch.no_grad():
-            # TODO: potentially mask invalid actions next_q
             logits_path = model(tf(path_obs))
+        logits_path = torch.where(masks == 0, logits_path, -loginf)
```
alexhernandezgarcia, line 1279: Would it not be `masks == 1`? We want to make `-loginf` the invalid (mask == True) actions.

alexhernandezgarcia: Sorry, got `torch.where` wrong. All good!

gflownet.py
alexhernandezgarcia: Sorry, got `torch.where` wrong. All good!

gflownet.py
```diff
@@ -1238,13 +1262,21 @@ def logq(path_list, actions_list, model, env):
     # TODO: this method is probably suboptimal, since it may repeat forward calls for
     # the same nodes.
     log_q = torch.tensor(1.0)
+    loginf = 1e3
     for path, actions in zip(path_list, actions_list):
         path = path[::-1]
         actions = actions[::-1]
         path_obs = np.asarray([env.state2obs(state) for state in path])
+        done = [0] * len(path)
```
alexhernandezgarcia, line 1270: This should be `done = [0 for _ in  range(len(path))]` because it is safer. Multiplying a list doesn't make copies.

gflownet.py
```diff
@@ -1238,13 +1262,21 @@ def logq(path_list, actions_list, model, env):
...
+        masks = tf(
+            [
+                env.get_mask_invalid_actions(path[idx], done[idx])
```
alexhernandezgarcia, line 1273: Why do we need to pass `done[idx]`?

nikita-0209: The function prototype is `get_mask_invalid_actions(self, state=None, done=None)` in the `grid` and `aptamers`. I just noticed it is different in the abstract method declaration in class `GFlowNetEnv`

nikita-0209: I changed it in class `GflownetEnv`

alexhernandezgarcia (commented):
@nikita-0209 I have changed a bunch of things. Please see my commits and let me know what you think.

gflownet.py
```diff
@@ -700,23 +705,41 @@ def sample_batch(
...
+                    state_id,
+                    masks,
+                )
             )
         t1_all = time.time()
         times["all"] += t1_all - t0_all
         return batch, times
 
-    def flowmatch_loss(self, it, batch):
+    def flowmatch_loss(self, it, batch, loginf=1000):
```
nikita-0209, line 742: do you think it would be better to make loginf a member variable of the `class GflownetAgent`?

gflownet.py
```diff
@@ -780,8 +803,8 @@ def flowmatch_loss(self, it, batch):
             with torch.no_grad():
                 next_q = self.target(sp)
         else:
-            # TODO: potentially mask invalid actions next_q
             next_q = self.model(sp)
+        next_q[masks] = -loginf
```
nikita-0209, line 807: `next_q[masks==1] = -loginf` is what we want. Elaborated in the `logq` comment

gflownet.py
```diff
@@ -673,7 +677,7 @@ def sample_batch(
                 )
             t0_a_envs = time.time()
             # Add to batch
-            for env, action, valid in zip(envs, actions, valids):
+            for env, action, mask, valid in zip(envs, actions, masks, valids):
```
nikita-0209, line 680: I don't think we want to return the `masks` from `forward_sample()`. Reason: 
We enter `forward_sample()` at `t=0` with `env.state = s0` and get invalid mask `m0`. Then we perform the `step()` function (inside `forward_sample` so `env.state` gets updated to `s1`. But we don't find the new mask `m1`, so what gets returned is `env` corresponding to `s1` but `mask` corresponding to `s0`, i.e., `m0`. I think we should compute the masks in the `sample_batch()` after getting the updated `env` from `forward_sample()`

gflownet.py
```diff
@@ -1234,17 +1257,19 @@ def empirical_distribution_error(env, visited):
...
+    loginf = tf([loginf])
     for path, actions in zip(path_list, actions_list):
         path = path[::-1]
         actions = actions[::-1]
         path_obs = np.asarray([env.state2obs(state) for state in path])
+        masks = tl([env.get_mask_invalid_actions(state, 0) for state in path])
         with torch.no_grad():
-            # TODO: potentially mask invalid actions next_q
             logits_path = model(tf(path_obs))
+        logits_path[masks] = -loginf
```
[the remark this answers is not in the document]
alexhernandezgarcia: Are you sure? If `masks` is just zeros or ones it then behaves as a boolean where 1 = True. I'll take a look too.

nikita-0209:
just ran the code, the `loss` and `logq` converge as expected. Ready to be merged imo

PR #4, opened 2022-11-11
modify traj balance

nikita-0209:
My understanding of the previous implementation was that it was considering all parents instead of just the parent that led to that state. So I modified the forward policy as well to just consider that particular parent that led to the state in that trajectory.
If we just have the forward policy in place, 
a. the previous code gives a log corr score of -0.55.
b. the modified forward policy implemented below gives a log corr score of -0.8

With the backward policy, the log corr score is -0.89. I am guessing -1 is expected so there might still be a bug in this code. I wasn't able to find anything after debugging for a few hours, but I will try over the weekend once again.

The policy currently outputs a vector of dim `len(action_space*2)`, but this could be modified to a neater implementation wherein the model outputs `P_f` and `P_b` (so that we don't have to slice the vector in the code of `traj_balance()`). Currently the policy is created with the help of a function. I am guessing we might want to use a class instead in the future. So then we could modify to returning `P_f, P_b`


gflownet.py
```diff
@@ -486,6 +489,7 @@ def __init__(
         self.oracle_period = args.gflownet.oracle.period
         self.oracle_nsamples = args.gflownet.oracle.nsamples
         self.oracle_k = args.gflownet.oracle.k
+        self.reward_clip_min = 1e-5
```
alexhernandezgarcia, line 492: I believe this is not used

gflownet.py
```diff
@@ -447,7 +450,7 @@ def __init__(
         self.model = make_mlp(
             [self.env.obs_dim]
             + [args.gflownet.n_hid] * args.gflownet.n_layers
-            + [len(self.env.action_space) + 1]
+            + [(len(self.env.action_space) + 1) * 2]
```
alexhernandezgarcia, line 453: Fine for now, but not to self and @nikita-0209: let's aim for having separate models for the forward and backward policies (optionally sharing weights) for more flexibility. The policies should optionally be fixed distributions (e.g. uniform) instead of a neural network, which is practical for the backward policy in some cases.

gflownet.py
```diff
@@ -520,7 +524,7 @@ def forward_sample(self, envs, times, policy="model", model=None, temperature=1.
         t0_a_model = time.time()
         if policy == "model":
             with torch.no_grad():
-                action_logits = model(tf(states))
+                action_logits = model(tf(states))[..., : len(self.env.action_space) + 1]
```
alexhernandezgarcia, line 527: Did you use the three dots notation for a particular reason? Since it is just one dimension that we replace by the ellipsis, I would use colon notation to avoid confusion (`[:, : len(self.env.action_space) + 1]`

gflownet.py
alexhernandezgarcia: But rewards come clipped from the environment (see [here](https://github.com/alexhernandezgarcia/gflownet/blob/main/gflownetenv.py#L110)). Do we need double clipping?

nikita-0209: Oh right, sorry missed this. Yeah no, double clipping does not make sense. I'll delete this

gflownet.py
```diff
@@ -826,23 +832,52 @@ def trajectorybalance_loss(self, it, batch):
...
             torch.zeros(len(torch.unique(path_id, sorted=True)))
-        ).index_add_(0, path_id_parents, logprobs)
+        ).index_add_(0, path_id, logprobs_b)
         # Sort rewards of done states by ascending path id
-        rewards = rewards[done.eq(1)][torch.argsort(path_id[done.eq(1)])]
-        # Trajectory balance loss
-        loss = (self.Z.sum() + sumlogprobs - torch.log((rewards))).pow(2).mean()
+        reward = reward[done.eq(1)][torch.argsort(path_id[done.eq(1)])]
+        # Trajectory balance loss sumlogprobs_b
+        loss = (self.Z.sum() + sumlogprobs_f - torch.log(reward)).pow(2).mean()
```
alexhernandezgarcia, line 880: `sumlogprobs_b` is never added (actually subtracted) to the final loss!

nikita-0209: Oh yeah, sorry I was testing a bunch of stuff because the loss wasn't converging as expected. I had removed that for one of the experiments. Forgot to add it back

PR #6, opened 2022-11-22
Simplify computation of trajectory balance loss

alexhernandezgarcia:
I have not carefully tested all yet.


nikita-0209:
looks good to me

PR #8, opened 2022-11-26
forward and backward policies

nikita-0209:
i have tested the code. Below are the stats
With flowmatch loss, logq=-1 [[comet](https://www.comet.com/nikita-0209/gfn-grid/b3f493ed93c94988bd38935f724943e8?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step)]
With tb loss,
- shared_network for forward and backward policies =-0.89 [[comet](https://www.comet.com/nikita-0209/gfn-grid/a5284111131b4fb08358bb13551da069?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step)]
- individual mlpsfor forward and backward = 0.83 [[comet](https://www.comet.com/nikita-0209/gfn-grid/644f9c228d894bfd87079f3d0dc0a9bf?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step)]
- uniform backward policy and mlp  as forward policy = 0.85 [[comet](https://www.comet.com/nikita-0209/gfn-grid/c784f7ed92fa4c649bbead7662920217?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step)]

To Do:
1. add masks for parents and then update backward_sample()
2. do we need to save the backward mlp (if used)? I don't tthink so because just the forward is used in sampling.
3. Some combinations won't make sense. Like choosing backward policy as mlp when loss is flowmatch. Need to think about the best way to add this check. Perhaps the config files in hydra can help?

### DIFF
--- a/gflownet.py
+++ b/gflownet.py
@@ -1018,17 +1018,6 @@ class GFlowNetAgent:
             .pow(2)
             .mean()
         )
-        for p1, p2 in zip(
-            self.forward_policy.model.parameters(),
-            self.backward_policy.model.parameters(),
-        ):
-            if p1.data.ne(p2.data).sum() > 0:
-                # p1.data.ne(p2.data).sum() > 0:
-                print("\n")
-                print(p1.shape)
-                print(p2.shape)
-                print(p1)
-                print(p2)
         return loss, loss, loss
 
     def unpack_terminal_states(self, batch):
@@ -1347,7 +1336,6 @@ def make_opt(params, Z, args):
     """
     Set up the optimizer
     """
-    # params = list(params)
     params = params
     if not len(params):
         return None
--- a/gflownet.py
+++ b/gflownet.py
@@ -291,6 +291,84 @@ def set_device(dev):
     _dev[0] = dev
 
 
+class Policy:
+    def __init__(self, env, type, args, device, shared_weight=False, base_model=None):
+        self.nn_input_dim = env.obs_dim
+        self.action_space_dim = len(env.action_space)
+        self.shared_weight = shared_weight
+        self.base_model = base_model
+        self.n_hid = args.gflownet.n_hid
+        self.n_layers = args.gflownet.n_layers
+        self.tail = []
+        self.act = nn.LeakyReLU()
+        self.device = device
+        # TODO
+        # Not sure what exactly tail is for. Do we need to pass it as an argument? Or would it be defined here?
+        if type == "uniform":
+            self.model = self.uniform_distribution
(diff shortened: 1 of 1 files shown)


nikita-0209:
Okay there is another way of writing the code for the combined optimizer. Instead of adding the `if` clause in the `parameters()` function, we could add in the make_opt() as is done for `Z`. But we'd need to feed the config or the  variables `forward_policy_type` and`backward_policy_type` into `make_opt()`

gflownet.py
```diff
@@ -462,16 +554,21 @@ def __init__(
...
         else:
             self.model_path = None
         self.ckpt_period = args.gflownet.ckpt_period
         if self.ckpt_period in [None, -1]:
             self.ckpt_period = np.inf
-        self.model.to(self.device_torch)
-        self.target = copy.deepcopy(self.model)
+        self.target = None
+        if self.forward_policy_type is not "uniform":
+            self.target = copy.deepcopy(self.forward_policy.model)
```
alexhernandezgarcia, line 571: @nikita-0209 what is this target variable for?

gflownet.py
```diff
@@ -462,16 +554,21 @@ def __init__(
...
+            ):
+                self.forward_policy.model.load_state_dict(torch.load(self.model_path))
                 print("Reloaded GFN Model Checkpoint")
         else:
             self.model_path = None
         self.ckpt_period = args.gflownet.ckpt_period
         if self.ckpt_period in [None, -1]:
             self.ckpt_period = np.inf
-        self.model.to(self.device_torch)
-        self.target = copy.deepcopy(self.model)
```
nikita-0209, line 474: I did not understand what target was used for. It was in the previous code, and had been used when the condition `self.tau>0` was `True`. Since I wasn't sure what `tau` represents, I kept `self.target`.

gflownet.py
```diff
@@ -823,8 +937,10 @@ def flowmatch_loss(self, it, batch, loginf=1000):
                 (1 - done).sum() + 1e-20
             )
 
-        if self.tau > 0:
-            for a, b in zip(self.model.parameters(), self.target.parameters()):
```
nikita-0209, line 827: Since there seemed to be an operation just on the `model.parameters()`, I initialized `self.target=None` when `forward_policy=="uniform`.

gflownet.py
nikita-0209: replied to this in a new conversation

PR #9, opened 2022-11-26
Changes to new Proxy implementation

alexhernandezgarcia:


alexhernandezgarcia:
Run of 5k iterations with flowmatch: [comet](https://www.comet.com/alexhernandezgarcia/gfn-grid/46ddc804a9fb4f3ea5bbf4fe3a34d991)

nikita-0209:
haven't gone through the code yet but my first feeling is something is up with the rewards. I expected them to be in the range of 0.6, instead of 1.8. Also, I remember `kl` ideally reaches `1e-5` in 2k iterations. Will check today night.

alexhernandezgarcia:
> 

Uhm right, that's odd. Good catch!

src/gflownet/envs/base.py
```diff
@@ -36,9 +36,11 @@ def __init__(
         self.reward_func = reward_func
         self.energies_stats = energies_stats
         self.denorm_proxy = denorm_proxy
-        # TODO: remove oracle as not required in env I think
-        self.oracle = oracle_func
         self.proxy = proxy
+        if oracle is None:
```
nikita-0209, line 40: Why do we need a `oracle` in `class Gflownetenv`?

nikita-0209: Oh wait, just to confirm, this is for `true_density`?

src/gflownet/envs/grid.py
```diff
@@ -64,24 +64,6 @@ def __init__(
...
-        if proxy:
-            self.proxy = proxy
-        else:
-            self.proxy = self.oracle
-        self.reward = (
-            lambda x: [0]
-            if not self.done
-            else self.proxy2reward(self.proxy(self.state2oracle(x)))
-        )
-        self._true_density = None
```
nikita-0209, line 83: Not really imp, but is there any reason we removed the initialization `self._true_density` from the `init()`?

nikita-0209:
Ah sorry my bad

> > 
> 
> ~Uhm right, that's odd. Good catch!~
> 
> The rewards are exp(-beta * proxy) so they can be 1.8. It's the proxy values that can be up to -0.6, which is the case.

Ah right okay, sorry my bad, thought `power` with `reward_beta=reward_norm=1` was being used

nikita-0209 (approved):

PR #12, opened 2022-11-27
correct traj balance

nikita-0209:
We had missed out on including masks while computing the forward and backward log probs. It is necessary because taking into consideration the masks skews the distribution over actions such that a very less probability is assigned on the actions that aren't valid, and a higher probability is assigned to the valid actions
For example, for the terminal states, `done=True` so the only possible backward action is `2`, hence the log prob for this action must be 0, which is enforced by assigning high negative values to the other "possible" but invalid actions.
[[comet](https://www.comet.com/nikita-0209/gfn-grid/f5a969a6c90940adb40d0af7314760d7?experiment-tab=chart&showOutliers=true&smoothing=0&transformY=smoothing&xAxis=step)]

Remarks:
- This is not the cleanest version of code. As discussed in one of our meetings, we could incorporate a more sophisticated way of choosing the forward and backward masks.
- TODO: Implement backward masks for aptamers as well.

### DIFF
--- a/grid.py
+++ b/grid.py
@@ -95,9 +95,9 @@ class Grid(GFlowNetEnv):
             actions += actions_r
         return actions
 
-    def get_mask_invalid_actions(self, state=None, done=None):
+    def get_mask_invalid_actions(self, state=None, done=None, obs2state=False):
         """
-        Returns a vector of length the action space + 1: True if action is invalid
+        Returns a vector of length the action space + 1: True if forward action is invalid
         given the current state, False otherwise.
         """
         if state is None:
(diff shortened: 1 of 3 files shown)


grid.py
```diff
@@ -114,6 +116,29 @@ def get_mask_invalid_actions(self, state=None, done=None):
...
+            mask.append(False)
+            return mask
+        if obs2state == True:
+            state = self.obs2state(state.cpu().numpy())
+        mask = [False for _ in range(len(self.action_space) + 1)]
+        for idx, a in enumerate(self.action_space):
+            for d in a:
+                if state[d] - 1 < 0:
+                    mask[idx] = True
+                    break
```
alexhernandezgarcia, line 139: Just for the records and for potential discussion if necessary, this will always make backward action eos (last in action space) valid, which I would argue is not correct. In any case, this function will not be necessary.

alexhernandezgarcia (approved):
COol

PR #13, opened 2022-12-05
Simplify obtention of forward and backward masks in trajectory balance loss

alexhernandezgarcia:
The main goal of this PR is to avoid the re-computation of invalid masks for the forward and backward logits:
* The masks of the forward logits (for the parents) are obtained from the (states) masks already included in the batch.
* The masks of the backward logits (fo the states) are obtained from the parents actions already included in the batch.

I have removed the legacy code from the previous PR.

### DIFF
--- a/gflownet.py
+++ b/gflownet.py
@@ -829,7 +829,7 @@ class GFlowNetAgent:
 
         return loss, term_loss, flow_loss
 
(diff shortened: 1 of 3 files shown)


nikita-0209:
makes sense to me

PR #19, opened 2022-12-13
Policy changes on hydra branch

alexhernandezgarcia:


src/gflownet/gflownet.py
```diff
@@ -841,18 +897,112 @@ def train(self):
...
+        else:
+            self.tail = []
+        if "type" in config:
+            self.type = config.type
+        elif self.shared_weights:
+            self.type = self.base.type
+        else:
+            raise "Policy type must be defined if shared_weights is False"
+        if self.type == "uniform":
+            self.model = self.uniform_distribution()
```
nikita-0209, line 940: should return the function instead `self.model = self.uniform_distribution()`

config/gflownet/trajectorybalance.yaml
```diff
@@ -8,3 +8,8 @@ optimizer:
   lr_decay_period: 1000000
   lr_decay_gamma: 0.5
 
+policy:
+  backward:
+    shared_weights: True
+    checkpoint: null
+    reload_ckpt: False
```
nikita-0209, line 15: If `shared_weights = False`, it should have `type`,  `n_hid` and `n_layers` as well. Do we want to add those here?

PR #38, opened 2023-01-05
change state2oracle to state2proxy to get rewards

nikita-0209:


nikita-0209:
also I am a little confused about where `self.reward` as defined on line 45 in `envs/base.py` is used

alexhernandezgarcia (approved):

PR #40, opened 2023-01-11
add amp environment

nikita-0209:
yet to be integrated with online training


gflownet/gflownet.py
```diff
@@ -81,8 +81,8 @@ def __init__(
             print("Unkown loss. Using flowmatch as default")
             self.loss = "flowmatch"
             self.Z = None
-        if not sample_only:
-            self.loss_eps = torch.tensor(float(1e-5)).to(self.device)
+        # if not sample_only:
+        self.loss_eps = torch.tensor(float(1e-5)).to(self.device)
```
nikita-0209, line 85: Initially this was declared only if `sample_only` is False. I think `loss_eps` is used irrespective of that?

gflownet/envs/base.py
```diff
@@ -106,7 +106,7 @@ def proxy2reward(self, proxy_vals):
             proxy_vals = proxy_vals * self.energies_stats[3] + self.energies_stats[2]
         if self.reward_func == "power":
             return np.clip(
-                (-1.0 * proxy_vals / self.reward_norm) ** self.reward_beta,
+                (proxy_vals / self.reward_norm) ** self.reward_beta,
```
nikita-0209, line 109: (to discuss): make code compatible with positive energies,
 `assert self.reward_norm > 0` because `np.log(self.reward_norm)` cannot take negative input

config/env/amp.yaml
```diff
@@ -0,0 +1,33 @@
...
+max_seq_length: 50
+# Number of letters in alphabet
+n_alphabet: 20
+# Minimum and maximum number of steps in the action space
+min_word_len: 1
+max_word_len: 1
+reward_func: power
+reward_norm: 1.0
+reward_norm_std_mult: -1.0
+reward_beta: 8.0
```
alexhernandezgarcia, line 18: If the defaults are the same as in base.yaml, then there is no need to add them here too

config/gflownet/gflownet.yaml
```diff
@@ -26,7 +24,7 @@ optimizer:
   # Train to sample ratio
   train_to_sample_ratio: 1
   # Number of training iterations
-  n_train_steps: 5000
+  n_train_steps: 2
```
alexhernandezgarcia, line 27: Is this intentional?

config/main.yaml
```diff
@@ -1,9 +1,25 @@
 defaults:
-  - env: grid
+  - _self_
+  - env: amp
   - gflownet: flowmatch
-  - proxy: grid_corners
+  - proxy: amp
   - logger: wandb
+  - user: nikita
```
alexhernandezgarcia, line 7: Are these the defaults we want in `main`? ^^

gflownet/envs/base.py
alexhernandezgarcia: We need to discuss this indeed. The convention in the repository is that oracles return negative values (energies), then the -1 will make the argument of the rewards positive.

gflownet/envs/base.py
```diff
@@ -125,7 +125,7 @@ def reward2proxy(self, reward):
         an oracle.
         """
         if self.reward_func == "power":
-            return -np.exp(
+            return np.exp(
```
alexhernandezgarcia, line 128: same as above

gflownet/gflownet.py
```diff
@@ -394,15 +400,15 @@ def sample_batch(
                 envs, actions, valids = self.forward_sample(
                     envs,
                     times,
-                    sampling_method="policy",
+                    sampling_method="mixt",
```
alexhernandezgarcia, line 403: you most likely do not want to have "mixt" sampling method as default when sampling candidates for test!

gflownet/gflownet.py
```diff
@@ -394,15 +400,15 @@ def sample_batch(
...
                     model=self.forward_policy,
                     temperature=1.0,
                 )
             else:
                 envs, actions, valids = self.forward_sample(
                     envs,
                     times,
-                    sampling_method="policy",
+                    sampling_method="mixt",
```
alexhernandezgarcia, line 411: I think the default should be "policy".

gflownet/proxy/amp.py
```diff
@@ -0,0 +1,44 @@
...
+            norm_constant=medoid_oracle_norm,
+        )
+        self.oracle.to(device)
+
+    def __call__(self, sequences, batch_size=256):
+        """
+        oracle.evaluate_many() returns a dictionary
+            confidence: 2D array (bartch, 2) where confidence[i][k] is the probability of the ith element being part of category "k"
+            prediction: 1D array, essentially argmax(probability)
+            entropy: 1D array, uncertainty in the prediction
```
alexhernandezgarcia, line 32: fix line lengths

gflownet/proxy/grid.py
```diff
@@ -3,7 +3,7 @@
 
 
 class GridCorners(Proxy):
-    def __init__(self):
+    def __init__(self, **kwargs):
```
alexhernandezgarcia, line 6: why?

gflownet/envs/amp.py
```diff
@@ -0,0 +1,662 @@
...
+            Number of train samples.
+
+        seed : int
+            Random seed.
+
+        output_csv: str
+            Optional path to store the test set as CSV.
+        """
+        df_train = pd.read_csv(
+            "/home/mila/n/nikita.saxena/gflownet/logs/train_dataset_{}.csv".format(
```
alexhernandezgarcia, line 436: a path should not be hardcoded in the code. use the config ;)

gflownet/envs/amp.py
```diff
@@ -0,0 +1,662 @@
...
+                split
+            )
+        )
+        df_valid = pd.read_csv(
+            "/home/mila/n/nikita.saxena/gflownet/logs/valid_dataset_{}.csv".format(
```
alexhernandezgarcia, line 441: a path should not be hardcoded in the code. use the config ;)

gflownet/envs/amp.py
```diff
@@ -0,0 +1,662 @@
...
+        #             split
+        #         ),
+        #         self.train_scores,
+        #     )
+        #     np.save(
+        #         "/home/mila/n/nikita.saxena/gflownet/logs/valid_scores_{}.npy".format(
+        #             split
+        #         ),
+        #         self.valid_scores,
+        #     )
```
alexhernandezgarcia, line 501: this should be cleaned up before merging into main

gflownet/envs/amp.py
```diff
@@ -0,0 +1,662 @@
...
+            {
+                "indices": [self.readable2state(s) for s in valid],
+                "samples": valid,
+                "energies": valid_scores,
+            }
+        )
+        return df_train, df_valid
+        # # if output_csv:
+        # df_train.to_csv("/home/mila/n/nikita.saxena/gflownet/logs/train_dataset_{}.csv".format(split))
+        # df_valid.to_csv("/home/mila/n/nikita.saxena/gflownet/logs/valid_dataset_{}.csv".format(split))
```
alexhernandezgarcia, line 527: clean up before merging

config/gflownet/gflownet.yaml
nikita-0209: Yep, because I'm still running some tests

nikita-0209:
working version with no-AL above

nikita-0209:
quick grid test [[results](https://wandb.ai/nikita0209/Scratch/runs/u0u7szsw)] after moving copy to `class GFlowNet' (base env class)

PR #45, opened 2023-01-14
Improvements (hopefully) of Logger

alexhernandezgarcia:


gflownet/gflownet.py
```diff
@@ -100,7 +97,6 @@ def __init__(
         self.progress = progress
```
nikita-0209, line 97: Do we want to move this to 'config/logger.yaml` as we discussed yesterday? And then do `self.progress = self.logger.progress`?

alexhernandezgarcia: Yes, we do

nikita-0209 (approved):

PR #59, opened 2023-01-25
add encoding of the states

AlexandraVolokhova:


alexhernandezgarcia (approved):
Thank you! :)

I have
- added it to ctorusmixture too
- corrected typo in variable name
- added condition that the dim >= 2
- change some variable names

PR #62, opened 2023-01-30
Fixed wandb config mess

AlexandraVolokhova:


config/experiments/debug/ctorus.yaml
```diff
@@ -0,0 +1,62 @@
...
+  test:
+    period: 25
+    n: 1000
+  checkpoints:
+    period: 500
+
+# Hydra
+hydra:
+  run:
+    dir: ${user.logdir.root}/debug/ctorus/${now:%Y-%m-%d_%H-%M-%S}
```
alexhernandezgarcia, line 62: @AlexandraVolokhova would you be able to remove this config file from the PR, since it is not related to it? Thanks!

alexhernandezgarcia (changes requested):
Thanks, it looks good! Except that the PR includes an unrelated file

PR #63, opened 2023-01-31
Changes for mask and step class methods

sh-divya:
(1) Function "get_mask_invalid_action" modified to crystals
    invalid actions:
        old elements not added again
        new element not added if adding leads > max_diff_elem
        conditions for eos action to be invalid

(2) Function "reset" changed according to new state representation

(3) Function "step" modified for crystals
    charge sum condition added for eos action

### DIFF
--- a/gflownet/envs/crystals.py
+++ b/gflownet/envs/crystals.py
@@ -6,8 +6,7 @@ from typing import List, Optional
 
 import numpy as np
 
-# from gflownet.envs.base import GFlowNetEnv
-from .base import GFlowNetEnv
+from gflownet.envs.base import GFlowNetEnv
 
 
 class Crystal(GFlowNetEnv):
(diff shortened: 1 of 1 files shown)


gflownet/envs/crystals.py
```diff
@@ -325,10 +330,23 @@ def step(self, action_idx):
...
-                self.n_actions += 1
+                nums_charges = [(num, self.oxidation_states[i]) for i, num in enumerate(self.state) if num > 0]
+                sum_diff_elem = []
+                for n, c in nums_charges:
+                    charges = []
+                    for c in itertools.product(c, repeat=n):
+                        sc = sorted(c)
+                        if sc not in charges:
+                            charges.append(sc)
+                    sum_diff_elem.append([sum(ci) for ci in charges])
```
michalkoziarski, line 344:
```suggestion
                    charge_sums = []
                    for c_i in itertools.product(c, repeat=n):
                        charge_sums.append(sum(c_i))
                    sum_diff_elem.append(np.unique(charge_sums))
```
Do we need to actually consider all charge configurations? Can't we simplify that by considering unique sums (something like suggestion)?

alexhernandezgarcia, line 344: Complementary comment to Michał's: I think it would be worth having a separate function to check the charge neutrality, like `has_neutral_charge(state)`

PR #65, opened 2023-02-01
add copy() to base and load data option to buffer

nikita-0209:
- Need to Discuss: how to adapt `reward2proxy` and `proxy2reward` if `proxy_val` is positive
- modified buffer to check same states before updating
- check whether there exists `idx_norandom=True` before feeding to model else `state2policy` runs into empty array error
- replace `deepcopy()` with `copy()`
- added some additional time logging (helpful in my AMP experiments), we can remove them if need be
- TODO: need to remove step from `log_metrics()` as that leads to issue with activelearning steps (unless there is a better way to do this)

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -236,20 +236,19 @@ class GFlowNetAgent:
             .to(bool)
         )
         # check for at least one non-random action
-        if idx_norandom.sum() > 0:
-            if sampling_method == "policy":
-                policy_outputs[idx_norandom, :] = model(
-                    self._tfloat(
-                        self.env.statebatch2policy(
-                            [s for s, do in zip(states, idx_norandom) if do]
-                        )
+        if sampling_method == "policy" and idx_norandom.sum() > 0:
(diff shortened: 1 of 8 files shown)


alexhernandezgarcia (approved):
Looks good! I made a few minor changes and will merge!

gflownet/gflownet.py
```diff
@@ -236,20 +236,19 @@ def forward_sample(
...
         )
         # check for at least one non-random action
-        if idx_norandom.sum() > 0:
-            if sampling_method == "policy":
-                policy_outputs[idx_norandom, :] = model(
-                    self._tfloat(
-                        self.env.statebatch2policy(
-                            [s for s, do in zip(states, idx_norandom) if do]
-                        )
+        if sampling_method == "policy" and idx_norandom.sum() > 0:
```
nikita-0209, line 239: ah I thought we might want to nest both conditions `policy` and `uniform` under the `idx_norrandom` check. Or we can add a `#TODO` and be mindful while writing the logic for uniform sampling

PR #66, opened 2023-02-01
add cont compatible amp

nikita-0209:
todo:
- fix `step` in logger
- discuss `load_dataset`
- proxy2reward for positive rewards
- handling var length in tb looks good?
- `env_id` required?


main.py
```diff
@@ -54,17 +55,22 @@ def main(config):
...
+        states, times = gflownet.sample_batch(env, config.n_samples, train=False)
+        samples = env.statebatch2oracle(states)
+        energies = env.oracle(samples)
+        gflownet.evaluate(samples, energies)
         df = pd.DataFrame(
             {
-                "readable": [env.state2readable(s) for s in samples],
-                "energies": energies.tolist(),
+                "readable": [env.state2readable(s) for s in states],
+                "energies": energies.cpu().numpy(),
```
alexhernandezgarcia, line 65: The energies will always be a one dimensional "array", that is a list. So maybe it's better to change the output of the oracle to make it work with `tensor.tolist()`? What do you think?

nikita-0209: Ah yes makes sense. The output of the oracle is a tensor itself. I don't remember what made me change it to numpy format before saving here, I will revert this to `tolist()`

alexhernandezgarcia (approved):
Done reviewing, thanks for the changes!

PR #67, opened 2023-02-01
toggle between -1 and 1 while converting proxy_vals to rewards

nikita-0209:
can potentially create `proxy/base.yaml` and move `higher_is_better` there (default value)


alexhernandezgarcia (approved):
Looks good, thanks!

PR #69, opened 2023-02-07
Fixes after merging PRs to make torus work

alexhernandezgarcia:


gflownet/envs/htorus.py
```diff
@@ -71,6 +72,9 @@ def __init__(
         # Setup proxy
         self.proxy.set_n_dim(self.n_dim)
 
+    def copy(self):
```
nikita-0209, line 75: Does the `copy()` in the base env class break here?

PR #71, opened 2023-02-07
Continuous fix logger

AlexandraVolokhova:


alexhernandezgarcia (approved):
Looks good to me!

nikita-0209:
When we log the steps of the gflownet as iteration numbers, it breaks the `activelearning` code as in activelearning, step number 1 (2, 3,.. and so on) is taken by the proxy train/val loss and then it proceeds on to the gflownet. I haven't found a way to log step 1 for the proxy and gflownet separately, so in my branch I have just removed the step parameter and wandb takes care of it with something of its own.
But if anyone has ideas, let me know and I will try implementing that

PR #72, opened 2023-02-08
Cleaned up active learning context arguments

AlexandraVolokhova:
Several changes to make code more flexible and readable:
1. renamed flag `al` to `active_learning` in the gfn init method to make it more readable and added it into gfn's config (@nikita-0209 that may break something in your code, pls review)
2. added `context` to parameters of the logger init method, added it to the config of the logger too


alexhernandezgarcia (approved):
Looks good to me, thanks!

PR #73, opened 2023-02-08
Added lr and lr_logZ logging

AlexandraVolokhova:
I've added lr and lr_logZ logging for keeping track of lr scheduling


alexhernandezgarcia (approved):
Looks good!

PR #75, opened 2023-02-08
Refactored Crystal env

michalkoziarski:


gflownet/envs/crystals.py
```diff
@@ -2,112 +2,118 @@
...
+            Mapping from ints (representing elements) to lists of different oxidation states
 
-        self.state = [0 for _ in range(periodic_table)]  # atom counts for each element
+        alphabet : (optional) dict
+            Mapping from ints (representing elements) to strings containing human-readable elements' names
+
+        required_elements : (optional) list
+            List of elements that must be present in a crystal for it to represent a valid end state
+        """
+        super(Crystal, self).__init__(**kwargs)
```
alexhernandezgarcia, line 68: A question just for the sake of learning from you @michalkoziarski: I thought that in recent Python versions (from 3+?) `super()` is equivalent to `super(ClassName, self)`. Is there a good reason to keep using `super(ClassName, self)`? (is it just because "explicit is better than implicit¨?)

michalkoziarski: Like we discussed :) but just for posterity, I don't see any reason other than backward compatibility to Python 2 (which doesn't matter in our case because the rest of code isn't compatible anyway), it was like that before and I didn't change it.

gflownet/envs/crystals.py
```diff
@@ -276,76 +258,102 @@ def get_parents(self, state=None, done=None, actions=None):
...
                 valid = False
             else:
                 self.state = state_next
                 valid = True
                 self.n_actions += 1
-            return self.state, action_idx, valid
+            return self.state, action, valid
         # If action is eos, then perform eos
         else:
             if sum(self.state) < self.min_atoms:
```
alexhernandezgarcia, line 323:
```suggestion
           self.get_mask_invalid_actions()[self.eos]:
```

alexhernandezgarcia (commented):
Nice work and again, I love the tests! Just suggested a few cosmetic changes.

gflownet/envs/crystals.py
```diff
@@ -139,68 +152,33 @@ def get_mask_invalid_actions(self, state=None, done=None):
...
-
-        # z = np.zeros(self.obs_dim, dtype=np.float32)
+            state = self.state
 
-        if len(state) > 0:
-            if hasattr(
-                state[0], "device"
-            ):  # if it has a device at all, it will be cuda (CPU numpy array has no dev
-                state = [subseq.cpu().detach().numpy() for subseq in state]
+        li_idx = self.elem2idx.get(3)
```
sh-divya, line 172: Just wanted to confirm that state2oracle will be Li or "current task" specific?

gflownet/envs/crystals.py
```diff
@@ -118,17 +124,24 @@ def get_mask_invalid_actions(self, state=None, done=None):
             state = self.state.copy()
         if done is None:
             done = self.done
-        state_elem = [i for i, e in enumerate(state) if e > 0]
-        state_atoms = sum(state)
+
         if done:
-            return [True for _ in range(len(self.action_space) + 1)]
-        mask = [False for _ in range(len(self.action_space) + 1)]
+            return [True for _ in self.action_space[:-1]] + [False]
```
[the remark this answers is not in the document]
michalkoziarski: Good catch, my bad.

gflownet/envs/crystals.py
```diff
@@ -118,17 +124,24 @@ def get_mask_invalid_actions(self, state=None, done=None):
...
+
+        mask = [False for _ in self.action_space]
+        state_elem = [self.idx2elem[i] for i, e in enumerate(state) if e > 0]
+        state_atoms = sum(state)
```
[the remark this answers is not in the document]
michalkoziarski: Agreed.

gflownet/envs/crystals.py
```diff
@@ -118,17 +124,24 @@ def get_mask_invalid_actions(self, state=None, done=None):
...
+
         if state_atoms < self.min_atoms:
             mask[self.eos] = True
         if len(state_elem) < self.min_diff_elem:
             mask[self.eos] = True
+        if any(r not in state_elem for r in self.required_elements):
+            mask[self.eos] = True
 
-        for idx, a in enumerate(self.action_space):
+        for idx, a in enumerate(self.action_space[:-1]):
```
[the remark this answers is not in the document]
michalkoziarski: Agreed, changed.

gflownet/envs/crystals.py
michalkoziarski: For now that's the case (I thought that's what we agreed on? But maybe I remembered wrong, let me know), the alternative would be to just return the state and do this conversion in a proxy/oracle wrapper.

alexhernandezgarcia: I'm happy to keep it simple like this for now. We can always update it later when/if we move to non-Li compounds.

PR #76, opened 2023-02-13
Remove config flatten code

alexhernandezgarcia:
It's not very important and the code crashes with some configurations (crystals)

### DIFF
--- a/main.py
+++ b/main.py
@@ -20,12 +20,6 @@ def main(config):
     random.seed(None)
     # Set other random seeds
     set_seeds(config.seed)
-    # Log config
-    # TODO: Move log config to Logger
-    log_config = flatten_config(OmegaConf.to_container(config, resolve=True), sep="/")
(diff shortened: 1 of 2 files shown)


main.py
```diff
@@ -20,12 +20,6 @@ def main(config):
     random.seed(None)
     # Set other random seeds
     set_seeds(config.seed)
-    # Log config
-    # TODO: Move log config to Logger
-    log_config = flatten_config(OmegaConf.to_container(config, resolve=True), sep="/")
```
michalkoziarski, line 25: Best to remove unused import `from gflownet.utils.common import flatten_config`.

main.py
```diff
@@ -20,12 +20,6 @@ def main(config):
...
-    log_config = {"/".join(("config", key)): val for key, val in log_config.items()}
```
michalkoziarski, line 26: Are you planning to revert / use `flatten_config` in the future? If not how about deleting it to clean things up?

michalkoziarski (approved):

PR #78, opened 2023-02-13
Add default values for element names and oxidation states

michalkoziarski:
Oxidation states received from @sh-divya, with missing ones filled with [0]. Possibly not final and might change in the future.


alexhernandezgarcia (approved):
Looks perfect to me! Thanks

PR #68, opened 2023-02-02
minor changes to adapt to mf

nikita-0209:
creating a draft PR because I keep forgetting the names of branches I am working on

PR #80, opened 2023-02-13
Update flowmatch loss code to latest changes in the code base

alexhernandezgarcia:
Note: this PR includes the fixes in [PR 79](https://github.com/alexhernandezgarcia/gflownet/pull/79) - changes in the buffer to prevent an infinite loop.

* [wandb](https://wandb.ai/alexhg/public/runs/7x9d0vtq) of `python main.py gflownet=flowmatch env=grid proxy=corners`
* [wandb](https://wandb.ai/alexhg/public/runs/w0o1q8fl) of `python main.py gflownet=trajectorybalance env=grid proxy=corners`

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -524,67 +524,59 @@ class GFlowNetAgent:
             Loss of the intermediate nodes only
         """
         loginf = self._tfloat([loginf])
-        batch_idxs = self._tlong(
-            sum(
-                [
-                    [i] * len(parents)
-                    for i, (_, _, _, parents, _, _, _, _, _) in enumerate(batch)
-                ],
-                [],
-            )
+        # Unpack batch
+        (
+            states,
+            actions,
+            parents,
+            parents_a,
+            done,
+            traj_id_parents,
+            state_id,
+            masks_sf,
+            masks_b,
+        ) = zip(*batch)
+        # Get state/batch id
+        parents_batch_id = self._tlong(
+            sum([[idx] * len(p) for idx, p in enumerate(parents)], [])
         )
-        sp, _, r, parents, actions, done, _, _, masks = map(torch.cat, zip(*batch))
-        # Sanity check if negative rewards
-        if self.logger.debug and torch.any(r < 0):
-            neg_r_idx = torch.where(r < 0)[0].tolist()
-            for idx in neg_r_idx:
-                state_oracle = self.env.state2oracle(sp)
-                output_proxy = self.env.proxy(state_oracle)
-                reward = self.env.proxy2reward(output_proxy)
-                import ipdb
-
-                ipdb.set_trace()
-
-        # Q(s,a)
-        parents_Qsa = self.forward_policy(parents)[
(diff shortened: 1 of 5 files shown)


gflownet/gflownet.py
```diff
@@ -57,8 +57,16 @@ def __init__(
...
+            print(
+                """
+            Flow matching loss is not available for continuous environments.
+            Trajectory balance will be used instead
+            """
+            )
+            optimizer.loss = "tb"
         # Loss
-        if optimizer.loss in ["flowmatch"]:
+        if optimizer.loss in ["flowmatch", "flowmatching"]:
```
nikita-0209, line 69: I'm sorry, but what is `flowmatching`? I mean why would it be an option in the config when we already have `flowmatch`?

gflownet/gflownet.py
```diff
@@ -524,68 +532,61 @@ def flowmatch_loss(self, it, batch, loginf=1000):
...
-        ]
-
-        # log(eps + exp(log(Q(s,a)))) : qsa
-        in_flow = torch.log(
-            self.loss_eps
-            + self._tfloat(torch.zeros((sp.shape[0],))).index_add_(
-                0, batch_idxs, torch.exp(parents_Qsa)
-            )
+        # Concatenate lists of tensors
+        states, parents, parents_a, done, masks_sf = map(
```
nikita-0209, line 552: I think we might have to deal with `states` and `parents` separately for variable length sequences as we did for `tb`

gflownet/gflownet.py
alexhernandezgarcia: of this is just to allow for both names ("flowmatch" or "flowmatching"), like for the trajectory balance loss we allow both "trajectorybalance" and "tb". Maybe we should add "fm" to the flowmatch loss too :P

gflownet/gflownet.py
alexhernandezgarcia: uhm but that actually broke a bunch of things :sweat_smile: let's deal with it when the need arises :grimacing:

gflownet/gflownet.py
nikita-0209: lol okay

gflownet/gflownet.py
nikita-0209: Yeah so in [PR#68](https://github.com/alexhernandezgarcia/gflownet/pull/68/files#diff-700050cd4007666a2c357edc47c03336bdbaddab3e135d60a3f382718580780b), I slightly modify that implementation by introducing a `do_padding` variable in the env, which is `False` by default (so that it does not break the grid/torus envs) and can be explicitly set to `True` for AMP

AlexandraVolokhova (approved):
looks good to me!

gflownet/gflownet.py
alexhernandezgarcia: Ah that sounds better. I think we should merge this for now and you can add that part to the flowmatch loss if you ever need to use it?

PR #81, opened 2023-02-15
Model specific Crystal MLP proxy and wrapper

sh-divya:
Proxy model specific to MLP trained on Sendek LR model. Currently accepts path to saved state_dict and standardization tensors, which are available in the data directory.

Class that wraps as well the the MLP module class are available

### DIFF
--- a/gflownet/envs/crystals.py
+++ b/gflownet/envs/crystals.py
@@ -233,7 +233,7 @@ class Crystal(GFlowNetEnv):
         self.id = env_id
         return self
 
-    def get_parents(self, state=None, done=None, actions=None):
+    def get_parents(self, state=None, done=None, action=None):
         """
         Determines all parents and actions that lead to a state.
 


gflownet/proxy/crystals.py
```diff
@@ -0,0 +1,111 @@
...
+            if l % 2 == 1:
+                x = self.hidden_act(x)
+
+        return x
+
+
+if __name__ == "__main__":
+    tmp = SendekMLPWrapper(
+        "all",
+        "/Users/divya-sh/Documents/gflownet/data/crystals",
```
alexhernandezgarcia, line 108: Ideally this should be taken from the config, specifically from the user/<username>.yaml config file.

gflownet/proxy/crystals.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def __call__(self, crystals):
+        """
+        Returns a vector of size [batch_size] that are calculated
+        ionic conductivity values between 0 and 1
+        """
+
+        if self.scale is not None:
+            crystals = (crystals - self.scale["mean"]) / self.scale["std"]
+        crystals = torch.nan_to_num(crystals, nan=0.0)
```
alexhernandezgarcia, line 59: Why would some values be nan?

gflownet/proxy/crystals.py
```diff
@@ -0,0 +1,111 @@
...
+        with torch.no_grad():
+            scaled_ionic_conductivity = self.oracle(crystals.to(torch.float32))
```
alexhernandezgarcia, line 61:
```suggestion
            scaled_ionic_conductivity = self.oracle(crystals.to(device=self.device, dtype=self.float))
```
The class proxy has attributes device and float to be used for this :)

alexhernandezgarcia (changes requested):
Looks great, thanks! I have suggested one (minor) change and commented about taking the path from the config instead of being hard coded. Not crucial because it won't break anything, but it would be nicer.

gflownet/proxy/crystals.py
sh-divya: the std of some features is 0, so division by zero leads to nans

alexhernandezgarcia: Oh I see, then I would catch that directly at the normalisation line, by adding a small constant. I'll suggest a change.

alexhernandezgarcia: Actually, if we want those values to become 0.0, then my suggestion would not work, obviously. In that case we should explicitly make those values 0.

config/proxy/crystals.yaml
```diff
@@ -0,0 +1,5 @@
+_target_: gflownet.proxy.crystals.SendekMLPWrapper
```
sh-divya, line 1: Proxy and user config file for crystal runs

gflownet/proxy/crystals.py
```diff
@@ -100,12 +100,3 @@ def forward(self, x):
                 x = self.hidden_act(x)
 
         return x
-
-
-if __name__ == "__main__":
```
sh-divya: removed hardcoding of paths

PR #82, opened 2023-02-16
Update and improve backward sampling 

alexhernandezgarcia:


gflownet/gflownet.py
```diff
@@ -256,112 +269,50 @@ def forward_sample(
...
-        action = parents_a[action_idx]
-        env.set_state((parents)[action_idx], done=False)
-        return env, env.state, action, parents, parents_a
+            valids = []
+            for env, action in zip(envs, actions):
+                parents, parents_a = env.get_parents(action=action)
+                if action in parents_a:
+                    state_next = parents[parents_a.index(action)]
+                    env.set_state(state_next, done=False)
+                    env.n_actions -= 1
```
nikita-0209, line 311: Not sure but wouldn;t we need to update `env.done` as well?

gflownet/gflownet.py
```diff
@@ -388,6 +339,34 @@ def sample_batch(
         Args
         ----
         """
+
+        def _add_to_batch(batch, envs, actions, valids, train=True):
```
nikita-0209, line 343: This is amazing!

gflownet/gflownet.py
alexhernandezgarcia: Yes, but that is already done by `env.set_state(state_next, done=False)` (`done` will be set to `False` inside `set_state`)

gflownet/gflownet.py
```diff
@@ -984,6 +947,10 @@ def log_iter(
 
 class Policy:
     def __init__(self, config, env, device, float_precision, base=None):
+        # If config is null, default to uniform
+        if config is None:
+            config = OmegaConf.create()
+            config.type = "uniform"
```
AlexandraVolokhova, line 953: That's a minor concern, but it seems to me it would be better to set it to "fixed" and let it be defined by the environment instead of being "uniform" because for some envs "uniform" may not exists (e.g. unbounded continuous spaces). Related thought: "uniform" is essentially a specific case of "fixed", maybe it should be moved into corresponding environments?

gflownet/gflownet.py
nikita-0209: Right, my bad, got it, thanks!

gflownet/gflownet.py
alexhernandezgarcia: Yeah I had thought of whether it should be uniform or fixed... For the purposes here it's better if it's random (i.e. uniform) than fixed and I agree it might not be defined in some special cases, but we don't need to deal with that at the moment.

PR #85, opened 2023-02-20
Alanine dipeptide

alexhernandezgarcia:


config/proxy/molecule.yaml
```diff
@@ -1 +1,3 @@
-_target_: gflownet.proxy.molecule.MoleculeEnergy
+_target_: gflownet.proxy.molecule.MoleculeEnergyProxy
+
+path_to_model: '/home/mila/a/alexandra.volokhova/projects/gflownet/data/random_forest_reward_100.pkl'
```
alexhernandezgarcia, line 3: This should be handled differently - not user-specific

PR #86, opened 2023-02-21
Modify molecule oracle to use TorchANI instead of XTB

michalkoziarski:


AlexandraVolokhova:
I've fixed a small bug, I've updated the class name in the config for proxy (before it referred to the old class name and couldn't find it). Now it looks good to me

AlexandraVolokhova (approved):

michalkoziarski:
> I've fixed a small bug, I've updated the class name in the config for proxy (before it referred to the old class name and couldn't find it). Now it looks good to me

Thank you for fixing that!

gflownet/proxy/base.py
```diff
@@ -11,13 +11,20 @@ class Proxy:
...
+        self.set_float_precision(float_precision)
         # Reward2Proxy multiplicative factor (1 or -1)
-        self.higher_is_better=higher_is_better
+        self.higher_is_better = higher_is_better
+
+    def set_device(self, device):
+        self.device = set_device(device)
+
+    def set_float_precision(self, dtype):
+        self.float = set_float_precision(dtype)
```
alexhernandezgarcia, line 27: Sorry, I am late to the party, but why is this necessary? I recently updated the whole codebase to make all classes use the same device and float precision and set it via `set_device, set_float_precision` from `gflownet.utils.common`.

Is it to use different device and float precision in the XTB proxy? If so, what is the reason?

michalkoziarski: That's a very good question - to be honest I just saw that in the previous version both XTB and RF proxy had this methods present, and assumed that it's used somewhere (but didn't really inquire where); so I simplified it under this assumption. TorchANI proxy actually uses it (to cast underlying model to the specific device).

So the question is, do we need (re)setting device/float precision past the initialization? If yes, then it should work just like before (and use your `set_device` and `set_float_precision` functions from `utils`); if not, it can be removed altogether.

PR #87, opened 2023-02-22
Cleanup analinedipeptide

AlexandraVolokhova:
Two changes:
1. I added automatic downloading of the pretrained proxy model and of a small dataset needed for defining conformer. I put the downloading function in `utils.common` because it can be potentially helpful for other environments. It uses [gdown](https://github.com/wkentaro/gdown) library, so you will need to install it.
2. Updated parameters in `experiments/icml23/molecule_mixture` according to the paper

@michalkoziarski now you should be able to run the icml experiment with molecule just via `python main.py +experiments=icml23/molecule_mixture user=your_user`

### DIFF
--- a/gflownet/utils/common.py
+++ b/gflownet/utils/common.py
@@ -1,7 +1,10 @@
-from collections.abc import MutableMapping
-import torch
+import gdown
 import numpy as np
+import torch
 
+from collections.abc import MutableMapping
+from hydra.utils import get_original_cwd
+from pathlib import Path
 
 def set_device(device: str):
     if device.lower() == "cuda" and torch.cuda.is_available():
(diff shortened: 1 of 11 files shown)


alexhernandezgarcia (approved):
Looks good to me, thanks.

gflownet/utils/common.py
```diff
@@ -42,3 +45,18 @@ def handle_logdir():
             print(f"logdir {config.logdir} already exists! - Ending run...")
     else:
         print(f"working directory not defined - Ending run...")
+
+def download_file_if_not_exists(path: str, url: str):
+    """
+    Download a file from google drive if path doestn't exist.
```
michalkoziarski, line 51:
```suggestion
    Download a file from google drive if path doesn't exist.
```

michalkoziarski (approved):
Looks good to me!

PR #88, opened 2023-02-22
Tiny but important change: update name of method

alexhernandezgarcia:
* `get_mask_invalid_actions_forward` <- `get_mask_invalid_actions`
* unpack action in for loop

- [x] Did you **test your PR locally** with `pytest` command?


michalkoziarski (approved):

michalkoziarski (changes requested):
There are multiple calls to this method in both this class and the tests that need to use changed name.

alexhernandezgarcia:
> There are multiple calls to this method in both this class and the tests that need to use changed name.

My bad :woman_facepalming: - I had not merged into `crystals_ahg` the latest changes in `crystals`... - and thanks for the warning! 

I have added another small change by the way.

gflownet/envs/crystals.py
```diff
@@ -229,7 +229,7 @@ def reset(self, env_id=None):
         self.id = env_id
         return self
 
-    def get_parents(self, state=None, done=None, actions=None):
+    def get_parents(self, state=None, done=None, action=None):
```
michalkoziarski, line 232: This is super minor, but it would be good to update the name in docstring as well (actions -> action).

gflownet/envs/crystals.py
```diff
@@ -263,8 +263,7 @@ def get_parents(self, state=None, done=None, actions=None):
         else:
             parents = []
             actions = []
-            for idx, action in enumerate(self.action_space[:-1]):
-                element, n = action
+            for idx, (element, n) in enumerate(zip(self.action_space[:-1])):
```
michalkoziarski, line 266: This breaks the tests, I believe zip is unnecessary.

michalkoziarski (changes requested):

gflownet/envs/crystals.py
alexhernandezgarcia: absolutely!

gflownet/envs/crystals.py
alexhernandezgarcia: Ugh, it's not that it breaks the tests, the zip is just wrong... Sorry! Fixed.

michalkoziarski: Well, it's both, I was subtly suggesting to use tests before pushing changes, haha.

michalkoziarski (approved):

PR #91, opened 2023-02-24
tensor compatible amp

nikita-0209:


gflownet/envs/amp.py
```diff
@@ -55,52 +79,36 @@ def __init__(
         self,
         max_seq_length=50,
         min_seq_length=1,
+        # Not required in env. But used in config_env in MLP. TODO: Find a way out
```
alexhernandezgarcia, line 82: Is it used to determine the dimensions of the inputs/outputs? We may want to use something like what is used in the continuous environments to determine the dimensions of the policy inputs/outputs. See [here](https://github.com/alexhernandezgarcia/gflownet/blob/continuous/gflownet/envs/htorus.py#L68)

gflownet/envs/amp.py
```diff
@@ -55,52 +79,36 @@ def __init__(
...
         n_alphabet=20,
```
alexhernandezgarcia, line 83: Since `AMINO_ACIDS` is used instead to determine the alphabet and there is no alternatives implemented, this should not be an argument of the environment.

gflownet/envs/amp.py
alexhernandezgarcia: I see you also used these variables here, so ignore the reference. But I do think the MLP should take the dimensions in a different way.

gflownet/envs/amp.py
nikita-0209: Yeah I agree. I need to change the activelearning MLP proxy implementation so that it's compatible with the AMP environment without `n_alphabet`. I was planning to create something similar to `policy_fixed_input` but that would depend on the `proxy_state_format` -- it is a slightly involved change so I was planning to do it in a separate PR

gflownet/envs/amp.py
```diff
@@ -202,34 +211,38 @@ def true_density(self, max_states=1e6):
...
-        state_oracle = [self.state2oracle(state) for state in state_batch]
-        return state_oracle
-
     def get_max_traj_len(
         self,
     ):
         return self.max_seq_length / self.min_word_len + 1
 
+    def statebatch2oracle(
+        self, states: List[TensorType["max_seq_length"]], bos_idx=None
```
alexhernandezgarcia, line 220: Is the input `states` a list of tensors as the typing indicates? Is this ever used?

gflownet/envs/amp.py
```diff
@@ -202,34 +211,38 @@ def true_density(self, max_states=1e6):
...
         for state in states:
-            if state[-1] == self.invalid_state_element:
-                state = state[: torch.where(state == self.invalid_state_element)[0][0]]
+            if state[-1] == self.padding_idx:
+                state = state[: torch.where(state == self.padding_idx)[0][0]]
+            if bos_idx is not None and state[0] == bos_idx:
+                state = state[1:-1]
             state_numpy = state.detach().cpu().numpy()
             state_oracle.append(self.state2oracle(state_numpy))
         return state_oracle
```
alexhernandezgarcia, line 243: This is nearly identical to `statebatch2oracle` so I would make one use the other.

gflownet/envs/amp.py
```diff
@@ -248,102 +261,68 @@ def state2policy(self, state=None):
...
         state_policy = torch.zeros(
-            (len(states), self.n_alphabet * self.max_seq_length),
-            dtype=self.float,
-            device=self.device,
+            states.shape[0], self.max_seq_length, self.n_alphabet
         )
-        state_policy[rows, cols] = 1.0
-        return state_policy
+        state_policy[:, : state_onehot.shape[1], :] = state_onehot
+        return state_policy.reshape(states.shape[0], -1).to(self.device).to(self.float)
```
alexhernandezgarcia, line 310: Minor, but it is better to set `device` and `dtype` directly when creating the tensor (in `zeros` here) to avoid making unnecessary copies from cpu to gpu.

gflownet/envs/amp.py
```diff
@@ -248,102 +261,68 @@ def state2policy(self, state=None):
...
+
+    def statebatch2state(
```
alexhernandezgarcia, line 312: This is new, right? What is this for?

gflownet/envs/amp.py
```diff
@@ -248,102 +261,68 @@ def state2policy(self, state=None):
...
+        self, states: List[TensorType["1", "state_dim"]]
+    ) -> TensorType["batch", "state_dim"]:
+        if self.tokenizer is not None:
+            states = torch.vstack(states)
+            states = self.tokenizer.transform(states)
+        return states.to(self.device)
 
     def statetorch2state(
```
alexhernandezgarcia, line 320: This is new, right? What is this for?

gflownet/envs/amp.py
```diff
@@ -481,24 +501,27 @@ def step(self, action: Tuple[int]) -> Tuple[List[int], Tuple[int, int], bool]:
...
+                    state_last_element : state_last_element + len(action)
+                ] = torch.LongTensor(action)
                 self.state = state_next
                 valid = True
                 self.n_actions += 1
             return self.state, action, valid
-        # If action is eos, then perform eos
         else:
-            if len(self.state) < self.min_seq_length:
+            state_last_element = int(torch.where(self.state == self.padding_idx)[0][0])
```
alexhernandezgarcia, line 523: This line is repeated, can we have it just once before the first use?

gflownet/envs/amp.py
```diff
@@ -590,7 +613,7 @@ def plot_reward_distribution(self, states=None, scores=None, ax=None, title=None
             plt.close()
         return ax
 
-    # TODO: do we need this?
+    # TODO: Deprecate as never used.
```
alexhernandezgarcia, line 616: You can remove this safely I think.

gflownet/gflownet.py
```diff
@@ -168,7 +168,13 @@ def __init__(
         self.jsd = -1.0
 
     def _tfloat(self, x):
-        return torch.tensor(x, dtype=self.float, device=self.device)
+        if isinstance(x, torch.Tensor):  # logq
+            return x.to(self.device).type(self.float)
+        elif isinstance(x[0], torch.Tensor):  # state is already a tensor
+            x = [x_element.unsqueeze(0) for x_element in x]
+            return torch.cat(x).to(self.device).type(self.float)
```
alexhernandezgarcia, line 175: Minor change to simplify:
```suggestion
            return torch.cat(x).to(self.device, self.float)
```

gflownet/gflownet.py
```diff
@@ -168,7 +168,13 @@ def __init__(
         self.jsd = -1.0
 
     def _tfloat(self, x):
-        return torch.tensor(x, dtype=self.float, device=self.device)
+        if isinstance(x, torch.Tensor):  # logq
+            return x.to(self.device).type(self.float)
```
alexhernandezgarcia, line 172: Minor change to simplify:
```suggestion
            return x.to(self.device, self.float)
```

gflownet/gflownet.py
```diff
@@ -631,48 +636,17 @@ def trajectorybalance_loss(self, it, batch, loginf=1000):
...
-                states = torch.cat([states, torch.cat(fid)], dim=1)
-        else:
-            parents, states = map(
-                torch.cat,
-                [
-                    parents,
-                    states,
-                ],
-            )
-
```
alexhernandezgarcia, line 666: Yay! :)

gflownet/gflownet.py
```diff
@@ -890,10 +864,10 @@ def test(self):
             z_pred = sum([hist[tuple(x)] for x in x_tt]) + 1e-9
             density_pred = np.array([hist[tuple(x)] / z_pred for x in x_tt])
             # corr = np.corrcoef(density_pred, density_true)[0, 1]
-            corr = 0.0
+            # corr = 0.0
```
alexhernandezgarcia, line 867: Remove?

gflownet/gflownet.py
```diff
@@ -890,10 +864,10 @@ def test(self):
             z_pred = sum([hist[tuple(x)] for x in x_tt]) + 1e-9
             density_pred = np.array([hist[tuple(x)] / z_pred for x in x_tt])
             # corr = np.corrcoef(density_pred, density_true)[0, 1]
```
alexhernandezgarcia, line 866: Remove?

alexhernandezgarcia (changes requested):
Looks good in general, relatively minor comments/suggestions only.

gflownet/envs/amp.py
nikita-0209: It is for the input to the deep kernel

gflownet/gflownet.py
nikita-0209: Ah no, sometimes I do not want to calculate the correlation (multi-fidelity grid) as it takes too long. I will implement a sophisticated way of toggling later on, but for now I just comment and uncomment whenever necessary

gflownet/gflownet.py
nikita-0209: Same as above

gflownet/envs/amp.py
nikita-0209: Yeah, statebatch is a list of tensors for AMP (as in other envs it is a list of lists). Yes it is used to get the true energies after the AL round as the trained gflownet samples a list of states (tensors)

alexhernandezgarcia (approved):
Thanks, looks good! (but don't forget about fixing the corr comments thing soon ;P)

PR #93, opened 2023-02-24
Eliminate difference between mixture and not

alexhernandezgarcia:
Instead of having different versions of the ctorus/alaninedipeptide for mixture and "normal", now there is only one version, which includes mixture capabilities but is named simply ctorus/alaninedipeptide. The formar "normal" version should be equivalent to the mixture with `n_comp=1`


AlexandraVolokhova (approved):
Looks good to me!

AlexandraVolokhova:
Oh, I've just noticed there're conflicts, they seem to be related to what I've done recently. @alexhernandezgarcia let me know if you need any help to resolve them

PR #95, opened 2023-02-28
Import gdown in function where it is used

alexhernandezgarcia:
Tiny PR: just to avoid the need of having `gdown` installed if it will not be used.


AlexandraVolokhova (approved):
Makes sense!

PR #97, opened 2023-03-03
Do not add to batch if step() returns invalid action

alexhernandezgarcia:
This is a _bug_ I introduced in when I re-implemented `sample_batch()`. I forgot to check if `valid is False` and not add to the batch if so. After some checks, because of the multiple layers of "security", I have not seen cases where valid is False, but it's better to double/triple/... check for now.


nikita-0209:
makes sense!

PR #98, opened 2023-03-06
add amp env along with few other changes for activelearning pipeline

nikita-0209:
Sorry, this is a slightly big PR -- it modifies trivial stuff in lots of files.

Summary of Changes:
1. correlation can be calculated in two ways: 
     a. ratio of the true and predicted density (for small sample spaces like a 10x10 grid) OR 
     b. sampling all possible trajectories to states in the test set. 
 The way correlation should be calculated is specified as a member variable of the environment (can be specified elsewhere as well, open to discussion), `corr_type.` If not specified, `corr=0.0`.
2. This change might affect all other proxies/oracles. Each proxy/oracle must be updated to have a member variable `maximize`. I didn't set a default value for `maximize` for all oracles/proxies as I feel that can lead to silent errors. `maximize=True` if the aim is to find samples that maximize the score given by the oracle. (Note: `higher_is_better` has been refactored to maximize.) 
3. Frequency of logging plots is now different from the frequency of testing (hence, variable `do_plot` is introduced)
4. `step` is not a required parameter in `logger.log_metric` because it messes up with the AL setup wherein step=1 corresponds to proxy's first training step, not the gflownet's first training step.
5. Introduces a linear shift reward2proxy transformation. Required when proxy values can be positive/negative simultaneously. 
6. Introduced `statetorch2state` and `statebatch2state` transformations in the envs: required for Branin and AMP DKL experiments


gflownet/gflownet.py
```diff
@@ -435,7 +442,11 @@ def _add_to_batch(batch, envs, actions, valids, train=True):
             )
             assert all(valids)
             # Filter out finished trajectories
-            envs_offline = [env for env in envs_offline if env.state != env.source]
+            if isinstance(env.state, list):
+                state_source_eq = env.state[0] != env.source[0]
+            elif isinstance(env.state, TensorType):
+                state_source_eq = torch.eq(env.state, env.source).all()
```
nikita-0209, line 448: change to inequality

PR #99, opened 2023-03-09
Update format of grid actions

alexhernandezgarcia:
Note that the base branch is `tests-and-simplify-envs` because it is the most up to date with the grid environment although it has not been merged into continuous yet.

This PR updates the format of the actions in the grid environment. It used to be a tuple of dimension indices, with variable length, and for each element d, the dimension d would be incremented by 1. With the updated format, the action space is the increment to be applied to each dimension. For instance, (0, 0, 1) will increment dimension 2 by 1 and the action that goes from [1, 1, 1] to [2, 3, 1] is (1, 2, 0). This has a number of advantages related to the actions having fixed length, plus it is more intuitive.

Besides the main change about the update of the action format, the PR include the following minor changes:
* Extended docstring
* Change of variable name: `max_step_len` is now `max_increment`
* `min_step_len` has been removed assuming the minimum should be 1.
* `max_dim_per_action` has been introduced to control how many dimensions can be incremented at a time.
* self.eos is now an action (a tuple) instead of an index. This should be done in other environments too.
* Assertions in `__init__` for more robustness.
* The grid tests include an environment with _extended action space_ (`max_increment` larger than 1 and `max_dim_per_action` larger than 1)

- [X] `python -m pytest tests/gflownet/envs/test_grid.py`
- [X] black

### DIFF
--- a/config/env/grid.yaml
+++ b/config/env/grid.yaml
@@ -9,9 +9,10 @@ func: corners
 n_dim: 2
 # Number of cells per dimension
 length: 3
-# Minimum and maximum number of steps in the action space
-min_step_len: 1
-max_step_len: 1
+# Maximum number of steps in the action space
+max_increment: 1
+# Maximum number of dimensions that can be incremented by one action
+max_dim_per_action: 1
 # Mapping coordinates
 cell_min: -1
 cell_max: 1
--- a/config/env/grid.yaml
+++ b/config/env/grid.yaml
@@ -9,7 +9,7 @@ func: corners
 n_dim: 2
 # Number of cells per dimension
 length: 3
-# Maximum number of steps in the action space
+# Maximum increment per each dimension that can be done by one action
 max_increment: 1
 # Maximum number of dimensions that can be incremented by one action
 max_dim_per_action: 1
--- a/gflownet/envs/grid.py
+++ b/gflownet/envs/grid.py
@@ -34,9 +34,6 @@ class Grid(GFlowNetEnv):
     length : int
         Size of the grid (cells per dimension)
 
-    min_increment : int
-        Minimum increment of each dimension by the actions.
-
     max_increment : int
         Maximum increment of each dimension by the actions.
 
@@ -54,7 +51,6 @@ class Grid(GFlowNetEnv):
         self,
         n_dim: int = 2,
         length: int = 3,
-        min_increment: int = 1,
         max_increment: int = 1,
         max_dim_per_action: int = 1,
         cell_min: float = -1,
@@ -64,17 +60,17 @@ class Grid(GFlowNetEnv):
         # Constants
         assert n_dim > 0
         self.n_dim = n_dim
-        self.eos = self.n_dim
         assert length > 1
         self.length = length
-        assert min_increment > 0
         assert max_increment > 0
-        assert min_increment <= max_increment
-        self.min_increment = min_increment
         self.max_increment = max_increment
+        assert max_dim_per_action > 0
+        self.max_dim_per_action = max_dim_per_action
         self.cells = np.linspace(cell_min, cell_max, length)
         # Source state: position 0 at all dimensions
         self.source = [0 for _ in range(self.n_dim)]
+        # End-of-sequence action
+        self.eos = tuple([0 for _ in range(self.n_dim)])
         # Base class init
         super().__init__(**kwargs)
         # Proxy format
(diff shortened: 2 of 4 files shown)


gflownet/envs/grid.py
```diff
@@ -62,15 +83,21 @@ def __init__(
...
+        increment to apply to dimension d of the hyper-grid.
         """
-        valid_steplens = np.arange(self.min_step_len, self.max_step_len + 1)
-        dims = [a for a in range(self.n_dim)]
         actions = []
-        for r in valid_steplens:
-            actions_r = [el for el in itertools.product(dims, repeat=r)]
-            actions += actions_r
-        actions += [(self.eos,)]
+        for incr in range(self.max_increment + 1):
```
michalkoziarski, line 91: I don't think this works for `max_dim_per_action > 1`, if I understand correctly, for example for `n_dim = 2` and `max_dim_per_action = 2`, the expected result should be `[(1, 0), (0, 1), (1, 1)]`, and it's `[(1, 0), (0, 1)]`?

If I understood the intended behavior correctly I would recommend writing a test for that with some expected, predefined values.

tests/gflownet/envs/test_grid.py
```diff
@@ -60,3 +72,7 @@ def test__statebatch2oracle__returns_expected(env, states, statebatch2oracle):
 
 def test__all_env_common(env):
     return common.test__all_env_common(env)
+
+
+def test__all_env_common(env_extended_action_space):
```
michalkoziarski, line 77: Common tests are great, but I would suggest also having some manual ones for `step`/`get_action_space`/`get_mask_invalid_actions_forward` functions where there is a couple of edge cases and expected output (for instance, for the bug in the previous version where invalid actions were not caught if there were multiple increments of the same dimension).

config/env/grid.yaml
```diff
@@ -9,9 +9,10 @@ func: corners
 n_dim: 2
 # Number of cells per dimension
 length: 3
-# Minimum and maximum number of steps in the action space
-min_step_len: 1
-max_step_len: 1
+# Maximum number of steps in the action space
```
michalkoziarski, line 12:
```suggestion
# Maximum increment per each dimension that can be done by one action
```
Maybe that's slightly more clear?

gflownet/envs/grid.py
```diff
@@ -14,16 +14,32 @@
...
+    dimensions.
+
+    The state space is the entire grid and each state is represented by the vector of
+    coordinates of each dimensions. For example, in 3D, the origin will be at [0, 0, 0]
+    and after incrementing dimension 0 by 2, dimension 1 by 3 and dimension 3 by 1, the
+    state would be [2, 3, 1].
+
+    The action space is the increment to be applied to each dimension. For instance,
+    (0, 0, 1) will increment dimension 2 by 1 and the action that goes from [1, 1, 1]
+    to [2, 3, 1] is (1, 2, 0).
```
michalkoziarski, line 27: Nice examples!

michalkoziarski:
By the way, since there is a `python -m pytest tests/gflownet/envs/test_grid.py` checkbox in the description - just to confirm, you are running all the tests, right? Not just the ones specific to the particular environment that is a focus of PR?

gflownet/envs/grid.py
alexhernandezgarcia: Thanks for the catch, my bad :( I have re-implemented this properly and added an explicit test for it.

gflownet/envs/grid.py
```diff
@@ -35,21 +51,26 @@ def __init__(
...
-        min_step_len: int = 1,
-        max_step_len: int = 1,
+        max_increment: int = 1,
+        max_dim_per_action: int = 1,
         cell_min: float = -1,
         cell_max: float = 1,
         **kwargs,
     ):
         # Constants
+        assert n_dim > 0
```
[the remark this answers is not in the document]
alexhernandezgarcia: Fine with me - done!

alexhernandezgarcia:
> By the way, since there is a `python -m pytest tests/gflownet/envs/test_grid.py` checkbox in the description - just to confirm, you are running all the tests, right? Not just the ones specific to the particular environment that is a focus of PR?

In this case I have not run all tests because this branch stemmed from `tests-and-simplify-envs` which is work in progress and some tests will fail for reasons other than the content of this PR.

michalkoziarski (approved):

nikita-0209 (approved):

PR #100, opened 2023-03-15
Small fix alaninedipeptide

AlexandraVolokhova:
Some of my previous changes in config files were lost (idk how it happened, probably some sloppy merge conflicts with other PRs), adding them again

### DIFF
--- a/config/env/alaninedipeptide.yaml
+++ b/config/env/alaninedipeptide.yaml
@@ -3,7 +3,8 @@ defaults:
 
 _target_: gflownet.envs.alaninedipeptide.AlanineDipeptide
 
-path_to_dataset: ${user.data.alanine_dipeptide}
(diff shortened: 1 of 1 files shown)


config/env/alaninedipeptide.yaml
```diff
@@ -3,7 +3,8 @@ defaults:
 
 _target_: gflownet.envs.alaninedipeptide.AlanineDipeptide
 
-path_to_dataset: ${user.data.alanine_dipeptide}
+path_to_dataset: './data/alanine_dipeptide_conformers_1.npy'
```
alexhernandezgarcia, line 6: Isn't it better to have the previous, user-dependent path? (`${user.data.alanine_dipeptide}`)

AlexandraVolokhova:
I think it's better to have this relative path so a user doesn't need to specify anything and can run the experiment right away. And if they want they can always redefine this path in their user config

PR #101, opened 2023-03-26
BIG PR: Simplification and homegenisation of environments, addition of tests, other clean up changes

alexhernandezgarcia:
- [x] `python -m pytest tests/gflownet/envs/`
- [x] `black .`
- [x] `isort .`

It's quite a massive PR and it's hard to mention all changes, but here is a non-exhaustive list of things:

- I have simplified the init method of the environments, such that everything that is common to all environments is now in the init of the parent class (`base.py`) and `super()` is called at the end of the init.
- Similar thing for the arguments: the arguments of an environment should only be those that are specific to the environment. Everything else goes in the `kwargs`. 
- And methods that were common or very similar for all environments have been unified into the parent class. A good example is `reset()`. Now it's common for all environments - I think - therefore there is no need to re-implement it.
- All environments now define the `self.source` state in the init, and this is used by `reset()` (in the parent class) to initialise the first `self.state`.
- The end-of-sequence (`self.eos`) in the environments is now an action with the format of other other actions and not an integer or whatever. This is more consistent in my opinion.
- There is a new method in the parent env class, `actions2indices()` that is used by the flowmatch loss.
- All proxies can implement a `setup()` method that receives the environment as argument and sets whatever variables are necessary, like `n_dim`. `setup()` is called in the init of the parent env class.
- In general, the proxies have also been cleaned up, and the uniform proxy is now working, which is useful for debugging.
- The buffer class is now in a separate script in `utils/buffer.py` 
- In the grid and the (discrete) torus, extended actions are now handled via the variables `max_increment` and `max_dim_per_action`, which specify how many positions (e.g. cells in the grid or discrete angles in the torus) can be done in one action and how many dimensions can be updated in one single action.
- I have re-arranged the order of methods in the environment class.
- I have extended and clarified the docstrings.
- I have added more mypy types - now most methods should have it. If missing, we should add them.
- `torus_rounds.py` was obsolete and has been removed.#
- I have added a few more assert statements for extra security.
- There are many new tests (although there should be more!) and since many are common to all environments, these are grouped in `tests/gflownet/env/common.py`
- `self.get_actions_space()` is now `self.get_action_space()`; `get_max_traj_len` is now `get_max_traj_length`; `self.d_action_space` is now `self.action_space_dim`.

Some things remain to be done:

- The aptamers (`aptamers.py`) environment has not been cleaned up and updated because I have the more ambitious plan of creating a Sequence parent class from which the aptamers will be inherited.
- The plane (`plane.py`) environment has not been cleaned up and updated because it is not working well yet.
- We need more tests, especially for the continuous environments, and for the GFlowNet agent itself.
- Docstrings need to keep improving.

:warning: I am not requesting careful reviews of the PR since it is huge, but reviews of things here and there and comments are very welcome. After merging this into `continuous`, my plan is to merge `continuous` into `main` (without further review, obviously) and continue improving things from there.


nikita-0209:
Regarding the aptamers bullet point, I have created a `Sequence` parent class from which child classes `Aptamers` and `AMP` are derived in the `cont_mf` branch. I will merge this branch into `cont-mf` and then create a PR with those changes.

alexhernandezgarcia:
> Regarding the aptamers bullet point, I have created a `Sequence` parent class from which child classes `Aptamers` and `AMP` are derived in the `cont_mf` branch. I will merge this branch into `cont-mf` and then create a PR with those changes.

Oh I didn't know you had done this already - that's great, thanks!

gflownet/gflownet.py
```diff
@@ -409,7 +409,7 @@ def _add_to_batch(batch, envs, actions, valids, train=True):
             for idx in range(n_empirical):
                 env = envs[idx]
                 env = env.set_state(x_tr[idx].tolist(), done=True)
-                env.n_actions = env.get_max_traj_len()
+                env.n_actions = env.get_max_traj_length()
                 envs_offline.append(env)
                 actions.append((env.eos,))
```
nikita-0209, line 414: Should this be `actions.append(env.eos)` as `env.eos` is already a tuple now (if I understand correctly)

alexhernandezgarcia: Indeed, good catch! Updated.

gflownet/envs/spacegroup.py
```diff
@@ -7,27 +7,28 @@
...
+from torchtyping import TensorType
 
 from gflownet.envs.base import GFlowNetEnv
-from gflownet.utils.crystals.constants import (
-    CRYSTAL_SYSTEMS,
-    CRYSTAL_CLASSES,
-    POINT_SYMMETRIES,
-    SPACE_GROUPS,
-)
+from gflownet.utils.crystals.constants import (CRYSTAL_CLASSES,
```
michalkoziarski, line 13: Wait, did black actually format this or is it manual? Mine seems to keep the previous version, and I am worried about it reverting it on next modification.

michalkoziarski (commented):
Doesn't have to be a part of this PR, but I think it's time to group envs related to crystal generation into a single directory.

gflownet/envs/spacegroup.py
alexhernandezgarcia: ugh this is an annoying conflict between `isort` and `black`. I guess I can fix it with some configuration by I haven't got my head around it yet.

gflownet/envs/base.py
```diff
@@ -63,46 +67,227 @@ def __init__(
...
+        # the action_dim dimension are True
+        return torch.where(torch.all(actions == action_space, dim=2))[1]
 
-    def get_max_traj_len(
+    def get_mask_invalid_actions_forward(
         self,
-    ):
-        return 1e3
+        state: Optional[List] = None,
+        done: Optional[bool] = None,
```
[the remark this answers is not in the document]
alexhernandezgarcia: What do you mean by "boilerplate with setting state/done to self values"?

gflownet/envs/base.py
```diff
@@ -20,40 +21,43 @@ class GFlowNetEnv:
...
+        reward_norm_std_mult: float = 0.0,
+        reward_func: str = "identity",
+        energies_stats: List[int] = None,
+        denorm_proxy: bool = False,
         proxy=None,
         oracle=None,
-        proxy_state_format=None,
+        proxy_state_format: str = "oracle",
+        fixed_distribution: dict = None,
+        random_distribution: dict = None,
```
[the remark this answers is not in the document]
alexhernandezgarcia: I'm afraid these are not optional in these continuous environments because they are used by [`get_policy_output()`](https://github.com/alexhernandezgarcia/gflownet/blob/tests-and-simplify-envs/gflownet/envs/htorus.py#L103).

gflownet/envs/base.py
```diff
@@ -20,40 +21,43 @@ class GFlowNetEnv:
...
+        reward_norm: float = 1.0,
+        reward_norm_std_mult: float = 0.0,
+        reward_func: str = "identity",
+        energies_stats: List[int] = None,
+        denorm_proxy: bool = False,
         proxy=None,
         oracle=None,
-        proxy_state_format=None,
+        proxy_state_format: str = "oracle",
+        fixed_distribution: dict = None,
```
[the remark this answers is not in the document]
alexhernandezgarcia: I'm afraid these are not optional in these continuous environments because they are used by [`get_policy_output()`](https://github.com/alexhernandezgarcia/gflownet/blob/tests-and-simplify-envs/gflownet/envs/htorus.py#L103).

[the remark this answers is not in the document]
alexhernandezgarcia: Does that make them optional or not? Is there a conditionally optional? I guess this will be solved with separate continuous and discrete base envs as @AlexandraVolokhova has suggested for other reasons.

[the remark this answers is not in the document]
michalkoziarski: I think that makes them optional in general, and ideally there would be a typing change for the continuous envs (or at least a check in the init, with exception raised if it's None in that context). Anyhow, I think `var_name: dict = None` is wrong either way, and it should either not have a None as default and be required, or have it and be optional. But I am fine with leaving this as is it here is supposed to be continuous base env soon.

gflownet/envs/base.py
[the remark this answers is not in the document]
michalkoziarski: I mean something like:
```
if state is None:
    state = self.state.copy()
if done is None:
    done = self.done
```
Which basically should be the case for every env, right? I was thinking if this can't be somehow extracted so that it doesn't have to be repeated in every child class. One option would be to have two methods, `get_mask_invalid_actions_forward` that actually can have optional arguments, and the only thing it does is sets them to `self.{state, done}` and calls `_get_mask_invalid_actions_forward`, which assumes they are not None and has actual logic - but I might be overthinking this and it might be an overkill.

Anyway, it's not a big issue, and I am ok with it being as is.

gflownet/envs/spacegroup.py
michalkoziarski: Could this help (disclaimer: I didn't test it)? https://github.com/PyCQA/isort/issues/1518

gflownet/envs/base.py
```diff
@@ -241,158 +479,38 @@ def reward2proxy(self, reward):
...
-        """
-        return str(traj).replace("(", "[").replace(")", "]").replace(",", "")
-
-    def reset(self, env_id=None):
+    def reset(self, env_id: Union[int, str] = None):
         """
         Resets the environment.
         """
-        self.state = []
+        self.state = self.source.copy()
```
AlexandraVolokhova, line 488: I'd like to have it as a method, smth like `self.get_source(reset=True)` to accommodate scenario when we start from random state

gflownet/proxy/base.py
```diff
@@ -2,8 +2,10 @@
 Base class of GFlowNet proxies
 """
 from abc import abstractmethod
```
AlexandraVolokhova, line 4:
```suggestion
from abc import abstractmethod, ABC
```
you need to inherit Proxy from ABC to make `@abstractmethod` work (throw error if it isn't implementer in the child class)

gflownet/proxy/base.py
```diff
@@ -2,8 +2,10 @@
...
+
 import numpy as np
 import numpy.typing as npt
+
 from gflownet.utils.common import set_device, set_float_precision
```
AlexandraVolokhova, line 11:
```suggestion
class Proxy(ABC):
```
continuation of the previous suggestion

AlexandraVolokhova: I'm sorry, it got a bit messy but for some reason I couldn't add this suggestion it to the line below

gflownet/proxy/uniform.py
```diff
@@ -1,16 +1,16 @@
-from gflownet.proxy.base import Proxy
 import torch
 from torchtyping import TensorType
 
+from gflownet.proxy.base import Proxy
+
 
 class Uniform(Proxy):
```
AlexandraVolokhova, line 7: Very nice proxy for testing, I didn't know we had it :) Could we add some readme about these things, somewhat like a guide or a bunch of tips "how to develop and test your env"?

gflownet/utils/buffer.py
AlexandraVolokhova, line 1: Big heart to this change ❤️ I'd love to do the same with the Policy class separating it from the gfn agent

config/tests.yaml
AlexandraVolokhova, line 1: It is for testing different envs, right? Could you show me how you do this in more detail, I'll need that for testing changes with Batch class and related things

AlexandraVolokhova (approved):

gflownet/envs/base.py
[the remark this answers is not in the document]
alexhernandezgarcia: Ah I see! I have actually been thinking of something like this too and I was wondering if there is an elegant way to handle this properly through inheritance. Since this is not crucial right now, I'd leave it for later, but it's a very good point and I agree.

gflownet/envs/spacegroup.py
alexhernandezgarcia: Works! The latest commit does
```
black .
```
then
```
isort --profile black .
```

gflownet/envs/base.py
alexhernandezgarcia: Yes, to be done when needed :+1:

gflownet/envs/torus.py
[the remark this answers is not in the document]
alexhernandezgarcia: Yes, torus is discrete. htorus is hybrid (continuous) torus and ctorus is "most" continuous torus (all angles updated at once)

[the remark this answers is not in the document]
alexhernandezgarcia: If you have a suggestion for better names, I welcome suggestions :)

gflownet/proxy/base.py
alexhernandezgarcia: Thanks!

gflownet/proxy/uniform.py
alexhernandezgarcia: Yes, I would like to write up something, but there are so many things to do...

gflownet/utils/buffer.py
alexhernandezgarcia: That's a pretty easy change!

config/tests.yaml
alexhernandezgarcia: This for now is just to load general config for the tests, for all envs. But yeah, happy to chat about how to do it for the Buffer.

PR #107, opened 2023-04-08
Update of LatticeParameters to new Grid format

michalkoziarski:
As in title, discrete version of LatticeParameters finally updated to the new Grid format, with changes to main also included.

The logic should mostly be the same, the biggest change is using new Grid action format (6-dimensional tuple, with i-th element holding the increment of i-th state position, instead of previous version with a variable-length tuple containing dimensions that will by incremented by 1).

### DIFF
--- a/gflownet/envs/lattice_parameters.py
+++ b/gflownet/envs/lattice_parameters.py
@@ -40,8 +40,7 @@ class LatticeParameters(Grid):
         min_angle: float = 30.0,
         max_angle: float = 150.0,
         grid_size: int = 61,
-        min_step_len: int = 1,
-        max_step_len: int = 1,
+        max_increment: int = 1,
         **kwargs,
     ):
         """
@@ -67,6 +66,9 @@ class LatticeParameters(Grid):
             Note that it has to be defined in such a way that 90 and 120 degree angles will be present in the
             mapping from grid cells to angles (np.linspace(min_angle, max_angle, grid_size)).
 
+        max_increment : int
+            Maximum increment of each dimension by the actions.
+
         min_step_len : int
             Minimum value of the step (how many cells can be incremented in a single step).
 
@@ -76,8 +78,8 @@ class LatticeParameters(Grid):
         super().__init__(
             n_dim=6,
             length=grid_size,
-            min_step_len=min_step_len,
-            max_step_len=max_step_len,
+            max_increment=max_increment,
+            max_dim_per_action=3,
             **kwargs,
         )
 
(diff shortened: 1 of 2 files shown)


gflownet/envs/lattice_parameters.py
```diff
@@ -67,17 +66,14 @@ def __init__(
...
+        max_increment : int
+            Maximum increment of each dimension by the actions.
         """
         super().__init__(
             n_dim=6,
             length=grid_size,
-            min_step_len=min_step_len,
-            max_step_len=max_step_len,
+            max_increment=max_increment,
+            max_dim_per_action=3,
```
alexhernandezgarcia, line 76: So `max_dim_per_action` will be fixed and equal to 3?

gflownet/envs/lattice_parameters.py
```diff
@@ -169,23 +165,27 @@ def get_actions_space(self) -> List[Tuple[int]]:
...
+        valid_steplens = [el for el in range(1, self.max_increment + 1)]
         actions = []
 
         # lengths
         for r in valid_steplens:
             for dim in [0, 1, 2]:
-                actions.append((dim,) * r)
-            actions.append((0, 1) * r)
-            actions.append((0, 1, 2) * r)
+                action = [0] * 6
```
alexhernandezgarcia, line 174:
```suggestion
                action = [0 for _ in range(6)
```
Because [list multiplications are evil](https://towardsdatascience.com/dont-use-python-s-list-multiplication-n-n-450c68d673a3)! (It should not matter here though)

gflownet/envs/lattice_parameters.py
```diff
@@ -169,23 +165,27 @@ def get_actions_space(self) -> List[Tuple[int]]:
...
+                actions.append(tuple(action))
+            actions.append((r, r, 0, 0, 0, 0))
+            actions.append((r, r, r, 0, 0, 0))
 
         # angles
         for r in valid_steplens:
             for dim in [3, 4, 5]:
-                actions.append((dim,) * r)
-            actions.append((3, 4, 5) * r)
+                action = [0] * 6
```
alexhernandezgarcia, line 183:
```suggestion
                action = [0 for _ in range(6)]
```
(same as above)

gflownet/envs/lattice_parameters.py
michalkoziarski: In practice, yes, since we only allow to either the angles or the lengths to be changed in a single step, not both. It shouldn't really matter what we set this to, because it's only used in the Grid's `get_action_space`, I set it to 3 because it seemed more consistent to me (and in case of any future changes in the Grid).

alexhernandezgarcia (approved):
Sorry I got interrupted while reviewing this and never approved it :( Approved from my side! And thanks for the work.

PR #109, opened 2023-04-11
plot methods in base env and grid and other minor changes

nikita-0209:
1. `higher_is_better` has been refactored to `maximize`: Each proxy/oracle must be updated to have a member variable `maximize`. I didn't set a default value for maximize for all oracles/proxies as I feel that can lead to silent errors. maximize=True if the aim is to find samples that maximize the score given by the oracle. 
2. remove `state_proxy_format`: In the absence of active learning, the proxy is always the oracle so `state2proxy` is always `state2oracle`
3. add linear shift `proxy2reward` transformation
4. add functionality to plot the reward distribution of sampled states

### DIFF
--- a/gflownet/envs/base.py
+++ b/gflownet/envs/base.py
@@ -35,18 +35,14 @@ class GFlowNetEnv:
         denorm_proxy: bool = False,
         proxy=None,
         oracle=None,
-        fixed_distribution: dict = None,
-        random_distribution: dict = None,
-        corr_type: str = None,
+        fixed_distribution: Optional[dict] = None,
+        random_distribution: Optional[dict] = None,
         **kwargs,
     ):
         # Call reset() to set initial state, done, n_actions
         self.reset()
         # Device
-        if isinstance(device, str):
-            self.device = set_device(device)
-        else:
-            self.device = device
+        self.device = set_device(device)
         # Float precision
         self.float = set_float_precision(float_precision)
         # Reward settings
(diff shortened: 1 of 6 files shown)


gflownet/envs/base.py
```diff
@@ -33,15 +35,18 @@ def __init__(
         denorm_proxy: bool = False,
         proxy=None,
         oracle=None,
-        proxy_state_format: str = "oracle",
-        fixed_distribution: Optional[dict] = None,
-        random_distribution: Optional[dict] = None,
+        fixed_distribution: dict = None,
+        random_distribution: dict = None,
+        corr_type: str = None,
```
alexhernandezgarcia, line 40: Note: to be discussed and possibly updated depending on the decisions from discussions in [PR 108](https://github.com/alexhernandezgarcia/gflownet/pull/108). See also [branch main_misc-ahg](https://github.com/alexhernandezgarcia/gflownet/tree/main_misc-ahg).

gflownet/envs/base.py
```diff
@@ -33,15 +35,18 @@ def __init__(
         denorm_proxy: bool = False,
         proxy=None,
         oracle=None,
-        proxy_state_format: str = "oracle",
-        fixed_distribution: Optional[dict] = None,
-        random_distribution: Optional[dict] = None,
+        fixed_distribution: dict = None,
+        random_distribution: dict = None,
```
alexhernandezgarcia, line 39: These should keep `Optional[dict] = None`

alexhernandezgarcia (approved):
It's been a while (partly my bad) and some of the changes would clash with more recent changes. Therefore, I have removed the stuff I was not sure about. Some of it is addressed in other branches, some will need to be addressed later. I have left the most important contributions, namely the plot methods.

PR #110, opened 2023-04-11
add sequence envs

nikita-0209:


config/env/aptamers.yaml
```diff
@@ -13,4 +12,4 @@ n_alphabet: 4
 # Minimum and maximum number of steps in the action space
 min_word_len: 1
 max_word_len: 1
-
+corr_type: None
```
alexhernandezgarcia, line 15: Note: to be re-assed

gflownet/envs/sequence.py
```diff
@@ -0,0 +1,470 @@
...
+        min_word_len=1,
+        max_word_len=1,
+        special_tokens=None,
+        **kwargs,
+    ):
+        self.min_seq_length = min_seq_length
+        self.max_seq_length = max_seq_length
+        self.min_word_len = min_word_len
+        self.max_word_len = max_word_len
+        self.corr_type = corr_type
```
alexhernandezgarcia, line 68: To be revised

alexhernandezgarcia: Actually, this does not need to be included as an argument in any case, because it should be part of kwargs

PR #111, opened 2023-04-12
To review: Batch class

AlexandraVolokhova:
Changed batch form a list of lists to a class.

Key improvements:
* batch is storing the envs now and functions like state2policy are called from these envs, not the generic self.env in the gflownet agent (this allows for working with conditional envs)
* depending on the loss (tb or flowmatch), batch stores different things needed for the loss. In particular, there's no need for get_parents method in the envs, where flowmatch doesn't make sense (like ctorus). We can remove them now

Tested with grid (first screenshot, purple is main branch, green is batch_class branch) and ctorus (still running, but looks good so far. Second screenshot, purple is main)

Screenshot 2023-04-12 at 16 08 33

Screenshot 2023-04-12 at 16 18 35

* [x] Fix batch loss


gflownet/gflownet.py
```diff
@@ -604,43 +536,55 @@ def trajectorybalance_loss(self, it, batch, loginf=1000):
...
+        #     states,
+        #     actions,
+        #     parents,
+        #     parents_a,
+        #     done,
+        #     traj_id_parents,
+        #     state_id,
+        #     masks_sf,
+        #     masks_b,
+        # ) = zip(*batch)
```
alexhernandezgarcia, line 554: Could we remove this commented block?

gflownet/gflownet.py
```diff
@@ -604,43 +536,55 @@ def trajectorybalance_loss(self, it, batch, loginf=1000):
...
         # Keep only parents in trajectory
-        parents = [
-            p[torch.where(torch.all(torch.eq(a, p_a), axis=1))]
-            for a, p, p_a in zip(actions, parents, parents_a)
-        ]
-        traj_id = torch.cat([el[:1] for el in traj_id_parents])
+        # parents = [
+        #     p[torch.where(torch.all(torch.eq(a, p_a), axis=1))]
+        #     for a, p, p_a in zip(actions, parents, parents_a)
+        # ]
```
alexhernandezgarcia, line 559: Could we remove this commented block?

gflownet/gflownet.py
```diff
@@ -604,43 +536,55 @@ def trajectorybalance_loss(self, it, batch, loginf=1000):
...
+        #     [
+        #         states,
+        #         actions,
+        #         parents,
+        #         done,
+        #         state_id,
+        #         masks_sf,
+        #         masks_b,
+        #     ],
+        # )
```
alexhernandezgarcia, line 582: Could we remove this commented block?

gflownet/gflownet.py
AlexandraVolokhova: Of course! Sorry, I forgot about cleaning them up

AlexandraVolokhova:
Wait a bit, I actually found out some bug yesterday while rerunning ctorus with different hyperparameters. I'll figure that out and also clean up the comments

AlexandraVolokhova:
Fixed the bug and cleaned up comments, ready for review 🙂

gflownet/utils/batch.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def add_to_batch(self, envs, actions, valids, train=True):
+        for env in envs:
+            self.envs.update({env.id: env})
+
+        for env, action, valid in zip(envs, actions, valids):
+            if not valid:
+                continue
+            if train:
+                self.state.append(deepcopy(env.state))
```
alexhernandezgarcia, line 40: Why is deepcopy needed?

alexhernandezgarcia:
> there's no need for get_parents method in the envs, where flowmatch doesn't make sense (like ctorus). We can remove them now
I don't think we can just remove `get_parents()` right now. For example, it is used for back

gflownet/utils/batch.py
```diff
@@ -0,0 +1,175 @@
...
+        self.parents = []
+        self.parents_actions = []
+        self.step = []
+
+    def __len__(self):
+        return len(self.state)
+
+    def add_to_batch(self, envs, actions, valids, train=True):
+        for env in envs:
+            self.envs.update({env.id: env})
```
alexhernandezgarcia, line 34: I don't understand the need for this. @AlexandraVolokhova could you explain it please?

gflownet/utils/batch.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def _process_trajectory_indicies(self):
+        trajs = {env_id: [] for env_id in self.envs.keys()}
+        for idx, (env_id, step) in enumerate(zip(self.env_id, self.step)):
+            trajs[env_id.item()].append((idx, step))
+        trajs = {
+            env_id: list(map(lambda x: x[0], sorted(traj, key=lambda x: x[1])))
+            for env_id, traj in trajs.items()
+        }
+        self.trajectory_indicies = trajs
```
alexhernandezgarcia, line 150: Sasha, would mind writing the docstring with what this method is expected to do? It's got 3 for loops and it is a bit hard to understand. I am wondering if we could simplify this.

gflownet/utils/batch.py
AlexandraVolokhova: Yes, you are right, we can do it in the beginning of the loop behind, there's no need for two loops here. My bad, I didn't code it optimally here

gflownet/utils/batch.py
AlexandraVolokhova: Sure, I can do that. It basically creates a dict where keys are env_ids (which are also trajectory ids as we assume that one env corresponds to only one trajectory in the batch) and values are indexes of the states in self.states which correspond to this trajectory (and they go in the right order from s_1 to s_f). 
As for the loops, the fist one is to create an empty dict, we can probably eliminate it with the defaultdict. The second loop gathers all the indexes for each trajectory together with step numbers (for sorting them later). And the last one sorts the indexes according to steps and removes steps.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,179 @@
+from collections import defaultdict
```
michalkoziarski, line 1: Should this really be in `gflownet/utils`? It is an essential part of the library. I would personally just put in in `gflownet/batch.py` (and would do the same for the oracle, for that matter).

gflownet/utils/batch.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    Important note: one env should correspond to only one trajectory, all env_id should be unique
+    """
+
+    def __init__(self, loss, device, float):
+        self.loss = loss
+        self.device = device
+        self.float = float
+        self.envs = dict()
+        self.state = []
```
michalkoziarski, line 20: Shouldn't all of these iterable variable names be plural? So states, actions, are_done, etc.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,179 @@
...
+
+import numpy as np
+import torch
+
+from gflownet.utils.common import tbool, tfloat, tint, tlong
+
+
+class Batch:
+    """
+    Important note: one env should correspond to only one trajectory, all env_id should be unique
```
michalkoziarski, line 12: Some more detailed description would be great here. For instance, the things you mentioned in the PR description, that is that it stores envs (and why), that depending on the loss function it stores different things, etc. Also, what exactly does it store (because it's a bunch of things).

gflownet/utils/batch.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+
+    def __init__(self, loss, device, float):
+        self.loss = loss
```
michalkoziarski, line 16: Also for the docstring description of the losses, the allowed values (and perhaps a check in init to make sure that it's one of the allowed values).

gflownet/utils/batch.py
```diff
@@ -0,0 +1,179 @@
...
+
+from gflownet.utils.common import tbool, tfloat, tint, tlong
+
+
+class Batch:
+    """
+    Important note: one env should correspond to only one trajectory, all env_id should be unique
+    """
+
+    def __init__(self, loss, device, float):
```
michalkoziarski, line 15: Is it possible to use a default value for the float here? Or is this actually problem-dependent? Partially so that it gives an example of what should it be.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,179 @@
...
+        self.parents = []
+        self.parents_actions = []
+        self.step = []
+
+    def __len__(self):
+        return len(self.state)
+
+    def add_to_batch(self, envs, actions, valids, train=True):
+        for env, action, valid in zip(envs, actions, valids):
+            self.envs.update({env.id: env})
```
michalkoziarski, line 35: Not that this is wrong, but is there any benefit of using update with a single-element dict as an argument?

gflownet/utils/batch.py
```diff
@@ -0,0 +1,179 @@
...
+                        env.state, env.done, [action]
+                    )
+                    self.mask_invalid_actions_backward.append(mask_b)
+            else:
+                if env.done:
+                    self.state.append(env.state)
+                    self.env_id.append(env.id)
+                    self.step.append(env.n_actions)
+
+    def process_batch(self):
```
michalkoziarski, line 67: This (and all of the others, to be honest) could use some docstrings with a short description of what's being done here.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,179 @@
...
+                    ]
+                )
+            elif self.loss == "trajectorybalance":
+                self.mask_invalid_actions_backward = tbool(
+                    self.mask_invalid_actions_backward, device=self.device
+                )
+            self._process_parents()
+
+    def _process_states(self):
+        self.state_gfn = tfloat(self.state, device=self.device, float=self.float)
```
michalkoziarski, line 96: 1) `state_gfn` is a slightly confusing name (maybe `state_tensor` would be better?), 2) I would suggest initializing all of the class variables in init (this also applies to `self.trajectory_indices` etc).

gflownet/utils/batch.py
AlexandraVolokhova: I personally treat `gflownet/utils` as an essential part of the library, just maybe more technical :) We have buffer, logger, metrics there, it seems to me that batch goes under the same umbrella

gflownet/utils/batch.py
AlexandraVolokhova: yes, Alex fixed that in his branch with improvements to this PR

gflownet/utils/batch.py
AlexandraVolokhova: Sorry, I don't get what you mean, could you elaborate?

gflownet/utils/batch.py
AlexandraVolokhova: `state_gfn` means that they are in the gfn format, but these will be changed with the Alex's suggestion we discussed in the gflownet-repo channel. About initialising all the class variables in the init, could you elaborate why you think it should be this way?

gflownet/utils/batch.py
AlexandraVolokhova: I have no idea 😅 that's something I follow from the GFlowNetAgent class, it sets all the floats in tensors to self.float. And Batch class just takes this self.float from the agent and sets it to the tensors created inside its methods

AlexandraVolokhova:
work in progress: I've just merged Alex's changes and then the cluster got shut down :( will continue on Mon with adding more docs and methods like get_states_policy

AlexandraVolokhova:
@alexhernandezgarcia @michalkoziarski please take a look again, I've changed things according to your suggestions. Tested new version with grid and ctorus env, works smoothly

gflownet/utils/batch.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def _process_parents(self):
+        """
+        Prepares self.parents (gflownet format) and self.parents_policy (policy format) as torch tensors.
+        Different behaviour depending on self.loss:
+            - for flowmatch, parents contain all the possible parents for each state,
+            so this tensor is bigger than self.state (all the parents are alinged the zero dimension, )
+            - for trajectorybalance, parents contain only one parent for each state which was its parent in the trajectory
+        """
+        if self.loss == "flowmatch":
```
carriepl, line 184: Would it make sense for the batch object not to know the loss and instead always collect everything and the GFlowNet will decide what to use? 
For instance, instead of having self.parents mean something different depending on the loss, have something like self.possible_parents for the possible parents of each state (as in FM) and self.parents for the actual parents of each state (as in TB).

I think that it might help with making the code a bit simpler and easier to maintain : 
- The batch doesn't need to know the loss (less coupling and less if/else)
- Easier to extend to new losses in the future since any such change will only need to be implemented in gflownet.py
- No attributes (like self.parents) that completely change semantics depending on the loss so the content of each attribute becomes clearer

In the PR description, you mention that you went this way because env.get_parents() doesn't make sense for some envs like ctorus. Could this be decided in Batch by looking at the env instead of the loss? Like, if the env has a get_parents() implemented, then you use it and collect the possible parents of each state. If it doesn't, then you don't (and, therefore, that information is not made available to the GFlowNet so it will need to pick a loss other than FM for such envs).

gflownet/utils/batch.py
```diff
@@ -0,0 +1,288 @@
...
+        for state, env_id in zip(states, env_ids):
+            states_policy.append(self.envs[env_id].state2policy(state))
+        return tfloat(states_policy, device=self.device, float_type=self.float)
+
+    def _process_parents(self):
+        """
+        Prepares self.parents (gflownet format) and self.parents_policy (policy format) as torch tensors.
+        Different behaviour depending on self.loss:
+            - for flowmatch, parents contain all the possible parents for each state,
+            so this tensor is bigger than self.state (all the parents are alinged the zero dimension, )
```
carriepl, line 181: aligned
And I think that part in parenthesis is missing a word.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+        Converts self.statees from list to torch tensor, computes states in the policy format
+        (stored in self.states_policy)
+        """
+        self.states_policy = self.states2policy()
+        self.states = tfloat(self.states, device=self.device, float_type=self.float)
+
+    def states2policy(self, states=None, env_ids=None):
+        """
+        Converts states from a list of states in gflownet fromat to a tensor of states in policy format
```
carriepl, line 157: format

gflownet/utils/batch.py
carriepl: I think he means is there a benefit to doing this : 
`self.envs.update({env.id: env})`
rather than this : 
`self.envs[env.id] = env`

To my knowledge, most people consider the second one to be more readable. Aside from that, it is also faster (see : https://stackoverflow.com/questions/15456158/python-dict-update-vs-subscript-to-add-a-single-key-value-pair) but I don't think that speed is a big issue here.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,288 @@
...
+            for par, env_id in zip(self.parents, self.env_ids):
+                parents_policy.append(
+                    tfloat(
+                        self.envs[env_id.item()].statebatch2policy(par),
+                        device=self.device,
+                        float_type=self.float,
+                    )
+                )
+            self.parents_policy = torch.cat(parents_policy)
+            # import ipdb; ipdb.set_trace()
```
carriepl, line 195: This can be removed

gflownet/utils/batch.py
carriepl: Although, some people will argue that using the update function is more OO

gflownet/utils/batch.py
AlexandraVolokhova: I see the reason behind your point, but I would prefer to inform batch about the loss and collect only the necessary information. apart from the issue with `get_parrents`, it is also more efficient to collect only the things which are needed for computing the loss, especially for long trajectories 

As for changing the semantics of the attributes names, maybe we can resolve it by having two different attributes: `all_possible_parents` and `parents` ?

gflownet/utils/batch.py
AlexandraVolokhova: sorry, I didn't get what you mean, could you elaborate?

gflownet/utils/batch.py
AlexandraVolokhova: could you elaborate?

gflownet/utils/batch.py
AlexandraVolokhova: true :)

gflownet/utils/batch.py
carriepl-mila: My bad, I meant that "fromat" should be changed to "format".

gflownet/utils/batch.py
carriepl-mila: Sorry, I meant that : 
- "alinged" should be changed to "aligned" (at least, that what I think it's meant to say)
- that it seems like one or a few words are missing between "all the parents are aligned" and "the zero dimension". 
- finally, the content of the parenthesis ends with a comma. I wasn't sure if it was a typo or an indication that there should be some text following the comma and it was forgotten/mistakenly deleted.

gflownet/utils/batch.py
carriepl-mila: Would it make sense if the Batch class had arguments like "collect_X", "collect_Y", ... for the various optional things that could be collected but won't always be required? Like for the possible parents, for instance.
Then, instead of providing the loss to the Batch class, the GFlowNet class can instead simply tell the Batch class what should be collected.
I think that this achieves your goal of efficiency but it also makes Batch agnostic to the loss. For instance, if we ever implement a new loss or new performance metrics, we only need to make the change in GFlowNet and GFlowNet will tell Batch what needs to be collected to compute this new loss/metric.

For the semantics, yes, I think that the way you are proposing would much clearer.

gflownet/utils/batch.py
carriepl: My bad, I meant that "fromat" should be changed to "format".

gflownet/utils/batch.py
carriepl: Sorry, I meant that : 
- "alinged" should be changed to "aligned" (at least, that what I think it's meant to say)
- that it seems like one or a few words are missing between "all the parents are aligned" and "the zero dimension". 
- finally, the content of the parenthesis ends with a comma. I wasn't sure if it was a typo or an indication that there should be some text following the comma and it was forgotten/mistakenly deleted.

gflownet/utils/batch.py
carriepl: Would it make sense if the Batch class had arguments like "collect_X", "collect_Y", ... for the various optional things that could be collected but won't always be required? Like for the possible parents, for instance.
Then, instead of providing the loss to the Batch class, the GFlowNet class can instead simply tell the Batch class what should be collected.
I think that this achieves your goal of efficiency but it also makes Batch agnostic to the loss. For instance, if we ever implement a new loss or new performance metrics, we only need to make the change in GFlowNet and GFlowNet will tell Batch what needs to be collected to compute this new loss/metric.

For the semantics, yes, I think that the way you are proposing would much clearer.

gflownet/utils/batch.py
AlexandraVolokhova: fixed :) thanks for pointing out, I'm having a lot of typos in my English 😅

gflownet/utils/batch.py
carriepl: The new names look good.
For the additional arguments, that's fair. I mean, the current PR is still a good improvement over the current state and there's nothing preventing a revisit in the future.

AlexandraVolokhova:
Issue with the speed is fixed, PR is ready to review @alexhernandezgarcia, @michalkoziarski

gflownet/envs/base.py
```diff
@@ -36,8 +36,11 @@ def __init__(
         proxy_state_format: str = "oracle",
         fixed_distribution: Optional[dict] = None,
         random_distribution: Optional[dict] = None,
+        conditional=False,
```
alexhernandezgarcia, line 39: We should also add this variable to the config, so that it can be controlled via config or command line. I will do it.

alexhernandezgarcia, line 39:
```suggestion
        conditional: bool = False,
```

gflownet/utils/batch.py
alexhernandezgarcia: I do not have much experience in creating python packages so I do not know the conventions, but either way seems ok. Let's leave it here for now?

gflownet/utils/batch.py
alexhernandezgarcia: I was not aware of this "debate". I also tend to use `update` because you need it to update more than one key in one go, so it seems more consistent to me to do the same with just one key. Anyway, speed matters!

So, should we change it to `self.envs[env.id] = env`?

gflownet/utils/batch.py
```diff
@@ -0,0 +1,346 @@
...
+        self.states = []
+        self.actions = []
+        self.done = []
+        self.env_ids = []
+        self.masks_invalid_actions_forward = []
+        self.masks_invalid_actions_backward = []
+        self.parents = []
+        self.all_possible_parents = []
+        self.all_possible_parents_actions = []
+        self.steps = []
```
alexhernandezgarcia, line 59: Minor, but I would rename `steps` to `n_actions`, for consistency with the environments.

gflownet/utils/batch.py
alexhernandezgarcia: I think we should eventually make the Batch independent of the loss, as Pierre-Luc argued, but I would do this later. No needed for this PR IMO.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,346 @@
...
+            self._process_parents()
+        self.is_processed = True
+
+    def _process_states(self):
+        """
+        Converts self.statees from list to torch tensor, computes states in the policy format
+        (stored in self.states_policy)
+        """
+        self.states = tfloat(self.states, device=self.device, float_type=self.float)
+        self.states_policy = self.states2policy()
```
alexhernandezgarcia, line 161: In order to make the API more readable and transparent, I would make this method return `states` and `states_policy` (instead of updating/creating `self.xxx`). What do you think?

gflownet/utils/batch.py
```diff
@@ -0,0 +1,346 @@
...
+                (states.shape[0], env.policy_input_dim),
+                device=self.device,
+                dtype=self.float,
+            )
+            for env_id in torch.unique(env_ids):
+                states_policy[env_ids == env_id] = self.envs[
+                    env_id.item()
+                ].statetorch2policy(states[env_ids == env_id])
+        else:
+            states_policy = env.statetorch2policy(states)
```
alexhernandezgarcia, line 191: Nice! :)

gflownet/utils/batch.py
```diff
@@ -0,0 +1,346 @@
...
+                        [],
+                    ),
+                    device=self.device,
+                )
+                self.all_possible_parents_actions = torch.cat(
+                    [
+                        tfloat(x, device=self.device, float_type=self.float)
+                        for x in self.all_possible_parents_actions
+                    ]
+                )
```
alexhernandezgarcia, line 147: I think that, for efficiency reasons, we should concatenate first and then call `tfloat`, to avoid creating many unnecessary tensors (inside `tfloat`). Tagging @carriepl to confirm that this would be better.

alexhernandezgarcia: I have re-implemented it [here](https://github.com/alexhernandezgarcia/gflownet/blob/batch_class-ahg/gflownet/utils/batch.py#L166)

gflownet/utils/batch.py
```diff
@@ -0,0 +1,346 @@
...
+        # Initialize empty batch variables
+        self.envs = dict()
+        self.states = []
+        self.actions = []
+        self.done = []
+        self.env_ids = []
+        self.masks_invalid_actions_forward = []
+        self.masks_invalid_actions_backward = []
+        self.parents = []
+        self.all_possible_parents = []
```
alexhernandezgarcia, line 57: I would rename this to `parents_all`.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,346 @@
...
+        self.all_possible_parents_actions = []
```
alexhernandezgarcia, line 58: I would rename this to `parents_actions_all`.

gflownet/utils/batch.py
```diff
@@ -0,0 +1,346 @@
...
+                perm_index.append(index[env_ids == env_id])
+            perm_index = torch.cat(perm_index)
+            # reverse permutation to make it index the states_proxy array
+            index[perm_index] = index.clone()
+            states_proxy = concat_items(states_proxy, index)
+        else:
+            states_proxy = env.statetorch2proxy(states)
+        return states_proxy
+
+    def _process_parents(self):
```
alexhernandezgarcia, line 228: Similar as my comment [above](https://github.com/alexhernandezgarcia/gflownet/pull/111/files#r1240217145) - I would return the relevant variables.

gflownet/utils/batch.py
alexhernandezgarcia: I am having second thoughts though, since the the method is private (`_method`). Interested in your opinion too @carriepl

gflownet/utils/batch.py
```diff
@@ -0,0 +1,346 @@
...
+        # rewards will be)
+        if not self.is_processed:
+            self.process_batch()
+        traj_actions = []
+        terminal_states = []
+        for traj_idx in self.trajectory_indices.values():
+            traj_actions.append(self.actions[traj_idx].tolist())
+            terminal_states.append(tuple(self.states[traj_idx[-1]].tolist()))
+        traj_actions = [tuple([tuple(a) for a in t]) for t in traj_actions]
+        return terminal_states, traj_actions
```
alexhernandezgarcia, line 330: Just a note that this will need to be rethought and re-implemented to adapt to current environments and needs. It is currently used only for the buffer and it seems it doesn't crash, so let's leave it as is for now. I have added TODO notes [here](https://github.com/alexhernandezgarcia/gflownet/blob/batch_class-ahg/gflownet/utils/batch.py#L375). To clarify, you haven't done anything wrong, Sasha! ;) (it's the old `unpack` method that was outdated).

alexhernandezgarcia (commented):
This is fantastic work, Sasha, and it will be extremely helpful in the future, so thanks so much!

I have made multiple comments and suggestions, but nothing is critical and everything seems to work well for me too. I am not approving yet because I would like to discuss some of the comments and because I would like to first merge my changes in [PR 142](https://github.com/alexhernandezgarcia/gflownet/pull/142) (if you approve it), which are mostly about the docstring and adding typing. It also includes a few of the suggestions I have made here. I have referenced to those in my comments. I believe we can merge this very soon! :)

gflownet/utils/batch.py
carriepl: I wouldn't expect a big difference when running on CPU but it would definitely help when running on GPU (because of the latency associated with transfering each individual value VS the whole thing all at once).

gflownet/utils/batch.py
AlexandraVolokhova: I'm not sure how to implement it here because this method changes different attributes depending on the loss (`parents_all_policy` and `parents_all` in the flowmatch case and `parents` and `parents_policy` in the trajectory_balance case)

gflownet/utils/batch.py
AlexandraVolokhova: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/111/commits/ef368e5c3b9e080415616d5983f45e5d2fa967da)

gflownet/utils/batch.py
AlexandraVolokhova: Done (here)[https://github.com/alexhernandezgarcia/gflownet/pull/111/commits/d3631459ec9ae11edfb1253cc58fd0d6eb71aef5]

gflownet/utils/batch.py
AlexandraVolokhova: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/111/commits/d3631459ec9ae11edfb1253cc58fd0d6eb71aef5)

gflownet/utils/batch.py
AlexandraVolokhova: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/111/commits/a3cb86224d29adbb3bf00a009fb97b0498be2729)

gflownet/utils/batch.py
AlexandraVolokhova: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/111/commits/a3cb86224d29adbb3bf00a009fb97b0498be2729)

gflownet/utils/batch.py
alexhernandezgarcia: I would return the four variables and make the unused ones empty lists. What do you think?

gflownet/utils/batch.py
```diff
@@ -0,0 +1,411 @@
...
+        self.trajectory_indices.
+        """
+        trajs = defaultdict(list)
+        for idx, (env_id, step) in enumerate(zip(self.env_ids, self.n_actions)):
+            trajs[env_id.item()].append((idx, step))
+        trajs = {
+            env_id: list(map(lambda x: x[0], sorted(traj, key=lambda x: x[1])))
+            for env_id, traj in trajs.items()
+        }
+        self.trajectory_indices = trajs
```
alexhernandezgarcia, line 374: Since we are making the `process_bla` methods return the relevant variables instead of silently change the attributes, I would also do it for this method.

gflownet/utils/batch.py
AlexandraVolokhova: well, I'm not strongly against, but it looks heavy to me, because then with new losses we'll potentially need to return even more things...

alexhernandezgarcia: I don't think the list of returned things will grow (much). But if we do not return stuff here, then we should not return it anywhere. I leave it up to you :) I also do not feel strongly about either way.

gflownet/utils/batch.py
AlexandraVolokhova: [Done](https://github.com/alexhernandezgarcia/gflownet/pull/111/commits/d09d740a51868c8c7022dd9dc7ec4b712b472f20)

gflownet/utils/batch.py
AlexandraVolokhova: [Done](https://github.com/alexhernandezgarcia/gflownet/pull/111/commits/7eb0f53a03ef1a5734cff7a8bb77a67e79f2776a)

alexhernandezgarcia (approved):
Ready to merge!!! :rocket:

PR #112, opened 2023-04-13
update config according to the paper

AlexandraVolokhova:
Cleaned up configs for ctorus and changed parameters to ones we put in the paper. Now the experiments can be reproduced with a simple command:

`python main.py +experiments=icml23/ctorus user=YOUR_USER`


alexhernandezgarcia (approved):
Thanks! I had been thinking of doing this too. Looks good!

PR #114, opened 2023-04-17
Tetris enviroment

alexhernandezgarcia:
- [x] `black .`
- [x] `isort --profile black .`
- [x] python -m pytest tests/gflownet/envs/

This PR only has new files, therefore it should not negatively interact with important stuff. All the environment tests are passed.

No need for a in-depth review, but I would like to merge this soon because the tetris environment has some particular features (e.g. states are 2D tensors) that I would like to use to test new features, like the new Batch class.

### DIFF
--- a/gflownet/envs/tetris.py
+++ b/gflownet/envs/tetris.py
@@ -1,0 +1,489 @@
+"""
+An environment inspired by the game of Tetris.
+"""
+import itertools
+import re
+from typing import List, Optional, Tuple
+
+import numpy as np
+import numpy.typing as npt
+import torch
+from torchtyping import TensorType
+
+from gflownet.envs.base import GFlowNetEnv
+
+PIECES = {
+    "I": [1, [[1], [1], [1], [1]]],
+    "J": [2, [[0, 2], [0, 2], [2, 2]]],
+    "L": [3, [[3, 0], [3, 0], [3, 3]]],
+    "O": [4, [[4, 4], [4, 4]]],
+    "S": [5, [[0, 5, 5], [5, 5, 0]]],
+    "T": [6, [[6, 6, 6], [0, 6, 0]]],
+    "Z": [7, [[7, 7, 0], [0, 7, 7]]],
+}
+
+
+class Tetris(GFlowNetEnv):
+    """
+    Tetris environment: an environment inspired by the game of tetris. It's not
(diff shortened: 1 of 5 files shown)


gflownet/envs/tetris.py
```diff
@@ -0,0 +1,489 @@
...
+}
+
+
+class Tetris(GFlowNetEnv):
+    """
+    Tetris environment: an environment inspired by the game of tetris. It's not
+    supposed to be a game, but rather a toy environment with an intuitive state and
+    action space.
+
+    The state space is 2D board, with all the combinations of pieces on it. Pieces are
```
AlexandraVolokhova, line 32: I didn't get what you mean here, especially by "an index that starts from piece_idx * max_pieces_per_type". It seems to me that indexes of pieces just go form 1 to 7, or maybe I'm missing something?

AlexandraVolokhova: Ah, I got what you mean after reading the tests and seeng numbers like 600 and 601 to identify different pieces of type 6. That makes sense (and it's also a nice solution for figuring out invalid backward actions easily). I'd paraphrase the docstring this way to make it more clear: "Pieces on the board are identified by a number that starts from piece_idx * max_pieces_per_type, and is incremented by 1 with each new piece from the same type. This number fills in the cells of the board where the piece is located"

AlexandraVolokhova (approved):
Looks good to me! I related idea: it seems to me we can easily get a lot of gaming envs from something like gym via writing a simple wrapper env to incorporate a gym.env into our framework

gflownet/envs/tetris.py
alexhernandezgarcia: Yes, that's right. Sure, I'll do that, thanks!

AlexandraVolokhova:
> > Looks good to me! I related idea: it seems to me we can easily get a lot of gaming envs from something like gym via writing a simple wrapper env to incorporate a gym.env into our framework
> 
> I am not sure I quite get what you mean and I am somehow sceptical that something like that would be "easy", so I'd be happy to hear more about that!

Well, by "easily" I mean that we won't have to bother with coding game mechanics (what actions and states are and how action changes the state) and also gym.envs follow a particular interface so I assume it will be possible to implement just one wrapper env in our code which will work with many envs from gym

PR #116, opened 2023-04-19
[FIX BUG] Fix bug in trajectory balance loss

alexhernandezgarcia:
The [line](https://github.com/alexhernandezgarcia/gflownet/blob/main/gflownet/gflownet.py#L641)
```
state_id[traj_id == tid] -= state_id[traj_id == tid].min() + 1
```
was yielding negative `state_id` which in turn resulted in wrong masks, among other potential issues. Replaced for now with:
```
state_id[traj_id == tid] = state_id[traj_id == tid] - state_id[traj_id == tid].min() + 1
```

The new [Batch class](https://github.com/alexhernandezgarcia/gflownet/pull/111) should replace this. This is a temporary fix. 

Thanks @nikita-0209 for finding this!

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -638,7 +638,9 @@ class GFlowNetAgent:
         )
         # Shift state_id to [1, 2, ...]
         for tid in traj_id.unique():
-            state_id[traj_id == tid] -= state_id[traj_id == tid].min() + 1
+            state_id[traj_id == tid] = (
+                state_id[traj_id == tid] - state_id[traj_id == tid].min() + 1
+            )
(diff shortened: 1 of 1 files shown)


alexhernandezgarcia:
> Could you elaborate how batch class should replace this? I didn't change this line in the PR and right now it does the same as in the main branch (so it has the same bug)

Right, I did not mean that this _has already been fixed_ but that we will potentially have more flexibility to handle the state ids and trajectory ids in a neater, more elegant and less error-prone way. I am working on it in "my branch".

AlexandraVolokhova (approved):

PR #117, opened 2023-04-20
Composition proxy

AlexandraVolokhova:
Simple proxy for testing compositions env, returns negative frequency of the number of protons in the composition. Frequencies are computed for all the stable materials in MP


gflownet/envs/crystals/composition.py
```diff
@@ -367,3 +367,14 @@ def _can_produce_neutral_charge(self, state: Optional[List[int]] = None) -> bool
         ]
 
         return any(poss_charge_sum)
+
+    def get_diff_elem_number(self, state=None):
```
alexhernandezgarcia, line 371: Just a matter of taste, but I would prefer a method name like:
- `get_number_of_elements` or
- `get_n_elements`

gflownet/envs/crystals/composition.py
```diff
@@ -367,3 +367,14 @@ def _can_produce_neutral_charge(self, state: Optional[List[int]] = None) -> bool
...
+        if state is None:
+            state = self.state
+        return (np.array(state) != 0).sum()
+
+    def get_atoms_number(self, state=None):
```
alexhernandezgarcia, line 376: Just a matter of taste, but I would prefer a method name like:
- `get_number_of_atoms` or
- `get_n_atoms`

gflownet/proxy/crystals/composition.py
```diff
@@ -0,0 +1,73 @@
...
+        self.normalise = normalise
+
+    def _get_max_protons_state(self, env):
+        max_protons_state = env.source.copy()
+        for elem in env.required_elements:
+            max_protons_state[env.elem2idx[elem]] = max(1, env.min_atom_i)
+        for idx in range(env.min_diff_elem - len(env.required_elements)):
+            max_protons_state[-1 - idx] = max(1, env.min_atom_i)
+
+        for idx in range(len(max_protons_state)):
```
alexhernandezgarcia, line 36: I think it would be easier to read if instead of iterating over the indices of the list (`idx in range(len(max_protons_state))` and then indexing the list (e.g. `if max_protons_state[-1 - idx] == 0`, we iterate over the list directly (starting from the end, which is what seems to be needed here): `for el in max_protons_state[::-1]` and then simply `if el == 0`, etc.

alexhernandezgarcia (approved):
Thaks for this!

To be honest, I have not carefully reviewed everything because some things are a bit complex, but what I understand looks good to me. I have made some suggestions, but nothing is critical and I leave it to you to take them or not.

I have also pushed a commit after running `black .` and `isort --profile black .`. Let's try to adopt this as a general practice before opening PRs.

gflownet/envs/crystals/composition.py
AlexandraVolokhova: Ok, I've changed that

gflownet/proxy/crystals/composition.py
AlexandraVolokhova: The problem is that I need idx to increment integers in the list `max_protons_state`. If I iterate over it like `for el in max_protons_state[::-1]` and work with `el`, the element of the list won't be changed as integers are immutable. Leaving it as it is for now (looks like it works according to the tests), but feel free to change that if you see a better way to code it

gflownet/proxy/crystals/composition.py
```diff
@@ -0,0 +1,73 @@
...
+                and env.get_diff_elem_number(max_protons_state) < env.max_diff_elem
+            ) or max_protons_state[-1 - idx] != 0:
+                while (
+                    env.get_atoms_number(max_protons_state) < env.max_atoms
+                    and max_protons_state[-1 - idx] < env.max_atom_i
+                ):
+                    max_protons_state[-1 - idx] += 1
+        return max_protons_state
+
+    def _get_max_protons_number(self, env):
```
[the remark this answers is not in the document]
AlexandraVolokhova: oh, I missed that, sorry. Yes, it returns the maximum number of protons that a state in the state space can have. Ok, I'll change the name

AlexandraVolokhova:
Thanks for reviewing! I'm sorry, I keep forgetting about black and isort 😅. Will try to remember of them in the future PRs 🙂

gflownet/proxy/crystals/composition.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self, normalise: bool = True, **kwargs):
+        super().__init__(**kwargs)
+        self.counts_dict = _read_protons_number_counts()
+        self.normalise = normalise
+
+    def _get_max_protons_state(self, env):
+        max_protons_state = env.source.copy()
+        for elem in env.required_elements:
+            max_protons_state[env.elem2idx[elem]] = max(1, env.min_atom_i)
+        for idx in range(env.min_diff_elem - len(env.required_elements)):
```
michalkoziarski, line 33: What if given element was already set to something other than zero before (or in other words, last `env.min_diff_elem - len(env.required_elements)` elements included some of the `required_elements`? Wouldn't this try to set them twice, leading to not enough different elements being used?

gflownet/proxy/crystals/composition.py
alexhernandezgarcia: Well, if you need the index, the standard way is then `for idx, el in enumerate(max_protons_state[::-1])`

tests/gflownet/proxy/crystals/test_composition.py
```diff
@@ -0,0 +1,113 @@
...
+from pathlib import Path
+
+import numpy as np
+import pytest
+import torch
+
+from gflownet.envs.crystals.composition import Composition
+from gflownet.proxy.crystals.composition import CompositionMPFrequency
+
+DEVICE = "cuda"
```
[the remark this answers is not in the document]
alexhernandezgarcia: I agree with Michał

gflownet/proxy/crystals/composition.py
alexhernandezgarcia: @AlexandraVolokhova ?

AlexandraVolokhova: Good catch! I've fixed that and added tests where this occurs

vict0rsch (approved):
Just a few comments about vectorization ; it's not exhaustive not required (very marginal in those cases IMHO) but good practice in general where applicable

PR #119, opened 2023-04-26
Moved lattice_parameters.py to crystals directory

michalkoziarski:
As in title, simple clean-up.


alexhernandezgarcia (approved):
I was going to write that the config file should be updated too, but I've seen there is no config file yet. That should be created to be able to train with it. But that can be done on a separate PR. This can be merged in my opinion.

PR #121, opened 2023-04-27
Combined Crystal environment

michalkoziarski:
- Implements Crystal environment, combining Composition, SpaceGroup and LatticeParameters in a sequential order.
- Adds QoL change to LatticeParameters to accept arbitrary grid size.
- Changes Composition oracle state format to contain atom counts (instead of Li atom counts, total atom counts, and frequencies as before).
- Contains some minor fixes to Composition.
- (Temporarily) disables Composition's neutral charge check - which would cause issues with Crystal environment getting stuck occasionally, and would significantly reduce training speed of GFlowNet. This needs to be rewritten to be included in forward mask.
- Also has some changes from `dav-proxy` that were committed directly here.


vict0rsch:
I got kicked out by @michalkoziarski so I've moved contributions to `dav-proxy`

PR #122, opened 2023-05-10
Composition refactor

michalkoziarski:
Changes to Composition environment's `get_mask_invalid_actions_forward` that take into account 1) required elements, and 2) atoms and elements that have to be reserved to meet minimum required values. Without these changes, environment could get stuck in some edge cases.

NOT INCLUDED: changes to neutral charge check, which was temporarily disabled - looking ahead and predicting from which states reaching neutral charge won't be possible is trickier than I hoped for...


alexhernandezgarcia (approved):
Wow tricky stuff going on here! Thanks for the changes, which look all good to me!

PR #123, opened 2023-05-11
Combined Crystal environment

michalkoziarski:
Re-opened PR because I had to rename the branch to avoid confusion.

- Implements Crystal environment, combining Composition, SpaceGroup and LatticeParameters in a sequential order.
- Adds QoL change to LatticeParameters to accept arbitrary grid size.
- Changes Composition oracle state format to contain atom counts (instead of Li atom counts, total atom counts, and frequencies as before).
- Contains some minor fixes to Composition.
- (Temporarily) disables Composition's neutral charge check - which would cause issues with Crystal environment getting stuck occasionally, and would significantly reduce training speed of GFlowNet. This needs to be rewritten to be included in forward mask.


config/env/crystal.yaml
alexhernandezgarcia, line 1: Why not having this inside directory `crystals/`?

gflownet/envs/crystals/composition.py
```diff
@@ -333,12 +344,18 @@ def step(self, action: Tuple[int, int]) -> Tuple[List[int], Tuple[int, int], boo
...
-                    valid = False
+                # TODO: re-enable charge check
+                # Currently enabling it causes errors when training combined
+                # Crystal env, and very significantly increases training time.
+                # if self._can_produce_neutral_charge():
+                #     self.done = True
+                #     valid = True
+                #     self.n_actions += 1
+                # else:
+                #     valid = False
```
alexhernandezgarcia, line 355: How replacing the commented out piece of code by
```
if self.do_charge_check:
  # do charge check
```
and `do_charge_check` is control by config?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+
+class Stage(Enum):
+    """
+    In addition to encoding current stage, values of this enum are used for padding individual
+    component environment's actions (to ensure they have the same length for tensorization).
+    """
+
+    COMPOSITION = -2
+    SPACE_GROUP = -3
+    LATTICE_PARAMETERS = -4
```
alexhernandezgarcia, line 24: Sorry, but why these values? Could we add some more info to the docstring?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+        )
+
+        # 0-th element of state encodes current stage: 0 for composition,
+        # 1 for space group, 2 for lattice parameters
+        self.source = (
+            [0]
+            + self.composition.source
+            + self.space_group.source
+            + self.lattice_parameters.source
+        )
```
alexhernandezgarcia, line 62: states (and source) are lists in all other environments. Is it here a tuple for a particular reason?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+        self.space_group_mask_start = self.composition_mask_end
+        self.space_group_mask_end = self.space_group_mask_start + len(
+            self.space_group.action_space
+        )
+        self.lattice_parameters_mask_start = self.space_group_mask_end
+        self.lattice_parameters_mask_end = self.lattice_parameters_mask_start + len(
+            self.lattice_parameters.action_space
+        )
+
+        self.eos = self.lattice_parameters.eos
```
alexhernandezgarcia, line 92:
```suggestion
        # EOS is EOS of LatticeParameters because it is the last stage
        self.eos = self.lattice_parameters.eos
```

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+        self.stage = Stage.COMPOSITION
+        self.max_action_length = max(
+            max(len(a) for a in self.composition.action_space),
+            max(len(a) for a in self.space_group.action_space),
+            max(len(a) for a in self.lattice_parameters.action_space),
+        )
+        self.done = False
```
alexhernandezgarcia, line 99:
```suggestion
```
Unnecessary because set in `super().__init__(**kwargs)`

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+    def _set_lattice_parameters(self):
+        """
+        Sets LatticeParameters conditioned on the lattice system derived from the SpaceGroup.
+        """
+        crystal_system = self.space_group.get_crystal_system()
+
+        if crystal_system is None:
+            raise ValueError(
+                "Cannot set lattice parameters without crystal system determined in the space group."
+            )
```
alexhernandezgarcia, line 117: I think this is incorrect for a couple of reasons:
- 1. get_crystal_system is meant to return the name of the crystal system. Actually, it will never return None but instead it will return the string "None" if the crystal system has not been set. 
- 2. I think what we want here is anyway not the crystal system but the crystal lattice system, don't we? That would be:
```suggestion
        if self.space_group.state[0] == 0:
            raise ValueError(
                "Cannot set lattice parameters without crystal system determined in the space group."
            )
```

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+
+        self.lattice_parameters = LatticeParameters(
+            lattice_system=self.space_group.lattice_system,
+            **self.lattice_parameters_kwargs,
+        )
```
alexhernandezgarcia, line 122: Shouldn't the lattice parameters work with crystal-lattice system instead of crystal system?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+
+    def _pad_action(self, action: Tuple[int], stage: Stage) -> Tuple[int]:
+        """
+        Pads action such that all actions, regardless of the underlying environment, have
+        the same length. Required due to the fact that action space has to be convertable to
+        a tensor.
+        """
+        return action + (stage.value,) * (self.max_action_length - len(action))
```
alexhernandezgarcia, line 130: The multiplication thing is probably fine because it is a tuple but flagging this just in case it could have some unwanted side effects. @vict0rsch ?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+        """
+        Reverses padding operation, such that the resulting action can be passed to the
+        underlying environment.
+        """
+        if stage == Stage.COMPOSITION:
+            dim = max(len(a) for a in self.composition.action_space)
+        elif stage == Stage.SPACE_GROUP:
+            dim = max(len(a) for a in self.space_group.action_space)
+        elif stage == Stage.LATTICE_PARAMETERS:
+            dim = max(len(a) for a in self.lattice_parameters.action_space)
```
alexhernandezgarcia, line 147: These for loops can be avoided by storing the max values, right? This may not be negligible because the action spaces are large and this will be called many many times (each call to `step()` for example)

alexhernandezgarcia: Actually, now I think about it, the length of the actions should be the same for a given environment. At least this is the case for the space group environment. I started coding things this way since I decided that at some point I want to port this to all-torch-tensors.

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+            self.lattice_parameters.action_space, Stage.LATTICE_PARAMETERS
+        )
+
+        action_space = (
+            composition_action_space
+            + space_group_action_space
+            + lattice_parameters_action_space
+        )
+
+        if len(action_space) != len(set(action_space)):
```
alexhernandezgarcia, line 170: Nice check! :)

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+        self.composition_kwargs = composition_kwargs or {}
+        self.space_group_kwargs = space_group_kwargs or {}
+        self.lattice_parameters_kwargs = lattice_parameters_kwargs or {}
+
+        self.composition = Composition(**self.composition_kwargs)
+        self.space_group = SpaceGroup(**self.space_group_kwargs)
+        # We initialize lattice parameters here with triclinic lattice system to access
+        # all the methods of that environment, but it will have to be reinitialized using
+        # proper lattice system from space group once that is determined.
+        # Triclinic was used because it doesn't force any initial starting angles.
```
alexhernandezgarcia, line 50: This is a bit annoying no? How about we make `LatticeParameters` be unconstrained by default? Would that solve the mentioned stateless issues with the lattice parameters?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+        elif stage == Stage.SPACE_GROUP:
+            space_group_state = self._get_space_group_state(state)
+            space_group_mask = self.space_group.get_mask_invalid_actions_forward(
+                state=space_group_state, done=False
+            )
+            mask[
+                self.space_group_mask_start : self.space_group_mask_end
+            ] = space_group_mask
+        elif stage == Stage.LATTICE_PARAMETERS:
+            # TODO: to be stateless this needs to set lattice system based on state
```
alexhernandezgarcia, line 267: I am not sure I understand what you mean by this? Mind elaborate?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+            mask[
+                self.lattice_parameters_mask_start : self.lattice_parameters_mask_end
+            ] = lattice_parameters_mask
+        else:
+            raise ValueError(f"Unrecognized stage {stage}.")
+
+        return mask
+
+    @staticmethod
+    def _stage2value(stage: Stage) -> int:
```
alexhernandezgarcia, line 283: Should this not be a method of `Stage` class instead?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+        # -2 => 0, -3 => 1, -4 => 2
+        return -(stage.value + 2)
+
+    @staticmethod
+    def _value2stage(value: int) -> Stage:
```
alexhernandezgarcia, line 288: Should this not be a method of `Stage` class instead?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+
+        if stage == Stage.COMPOSITION or (
+            stage == Stage.SPACE_GROUP
+            and self.space_group.state == self.space_group.source
+        ):
+            parents, actions = self.composition.get_parents(
+                state=self._get_composition_state(state)
+            )
+            parents = [
+                [0] + p + self.space_group.source + self.lattice_parameters.source
```
alexhernandezgarcia, line 366: How about making creating a method `_build_state(stage, composition, space_group, lattice_parameters)`?

gflownet/envs/crystals/crystal.py
```diff
@@ -0,0 +1,488 @@
...
+            ]
+            actions = [self._pad_action(a, Stage.LATTICE_PARAMETERS) for a in actions]
+        else:
+            raise ValueError(f"Unrecognized stage {stage}.")
+
+        return parents, actions
+
+    def state2oracle(self, state: Optional[List[int]] = None) -> Tensor:
+        """
+        Prepares a list of states in "GFlowNet format" for the oracle.
```
alexhernandezgarcia, line 403:
```suggestion
        Prepares a list of states in "GFlowNet format" for the oracle. Simply
        a concatenation of all crystal components.
```

config/env/crystal.yaml
```diff
@@ -0,0 +1,25 @@
...
+# Composition parameters
+id: crystal
+composition_kwargs:
+  elements: 89
+lattice_parameters_kwargs:
+  min_length: 1.0
+  max_length: 5.0
+  min_angle: 30.0
+  max_angle: 150.0
+  grid_size: 100
```
alexhernandezgarcia, line 15:
```suggestion
  grid_size: 10
```

gflownet/envs/crystals/lattice_parameters.py
```diff
@@ -39,7 +41,7 @@ def __init__(
         max_length: float = 5.0,
         min_angle: float = 30.0,
         max_angle: float = 150.0,
-        grid_size: int = 61,
+        grid_size: int = 100,
```
alexhernandezgarcia, line 44:
```suggestion
        grid_size: int = 10,
```

alexhernandezgarcia (changes requested):
Really nice work! I generally looks good to me. I have made some suggestions that I would like to have or discuss before merging, but nothing is critical.

gflownet/envs/crystals/crystal.py
michalkoziarski: It is a list :) parenthesis are just to group the summation operation.

gflownet/envs/crystals/crystal.py
michalkoziarski: It's a tuple of ints that is never modified afterwards, I believe it should be find.

gflownet/envs/crystals/crystal.py
michalkoziarski: Good point about computational cost, I changed it to be computed only once.

Regarding same action length, I know, and I believe this is the case right now, but I think it's worth checking it once anyway to prevent issues with any possible future changes.

gflownet/envs/crystals/crystal.py
michalkoziarski: Stateless thing is a separate issue that wouldn't be solved by this. Yes, it is annoying, but I don't think making `LatticeParameters` unconstrained will help - the thing is that `LatticeParameters` just has to have constraints to work properly (to determine the source, masks etc), and if anything, removing the need for the constraint will make it work like using the triclinic system anyway, while making the `LatticeParameters` have additional edge case to consider internally.

gflownet/envs/crystals/crystal.py
michalkoziarski: I am not sure what you mean? `space_group.lattice_system` is being used currently in the code you cited, so that's not the crystal system, right?

michalkoziarski: (I think the confusion might have been, I was checking for `crystal_system` being set and raised error if it wasn't, but the `LatticeParameters` were, I think properly, conditioned on the `lattice_system`)

gflownet/envs/crystals/crystal.py
alexhernandezgarcia: 2. No, the lattice system and the crystal system are set simultaneously because they are both set via the crystal-lattice system, which is the first dimension of the space group env.

alexhernandezgarcia: Note that I have made an edit of the error message too

alexhernandezgarcia:
Thanks for the changes and answers! This can be merged for me - pending the related PR.

PR #129, opened 2023-05-31
To review: Dave proxy

vict0rsch:
## 💥 New requirements

Breaking change: new requirements for the proxy. 

**Recommended install:** make a new env and follow the instructions in `setup/crystals.sh`.

_Otherwise:_

1. Update `torch`
	```
	$ pip install torch --upgrade
	```
2. Install `torch_scatter` and `torch_geometric`
	```bash
	$ module load cuda/11.7
	$ pip install -r https://raw.githubusercontent.com/sh-divya/ActiveLearningMaterials/main/requirements_inference.txt
	```

Note: `torch_scatter` and `torch_geometric` are known to be difficult to install in general. This is due to their _precise_ dependencies on torch and CUDA versions. If you do not want to install with `torch>=2.0.1` and `CUDA 11.7` then you may have to install things [manually](https://pytorch-geometric.readthedocs.io/en/latest/notes/installation.html)

Note 2: `torch_scatter` is going to be slow to install (`Building wheel for torch_scatter (setup.py)` may take several minutes), even when all is successful.

### TODO:

* [ ] train proxy with conventional unit cell
* [x] remove requirements from `requirements.txt` and `pyproject.toml`


config/logger/base.yaml
```diff
@@ -17,6 +17,9 @@ test:
   kde:
     bandwidth: 0.1
     kernel: gaussian
+  n_top_k: 5000
+  top_k: 100
+  top_k_period: 2500
```
alexhernandezgarcia, line 22: I find this structure a bit confusing, since the top-k metrics variables are children of the `test` config but they are independent in the code. I would say that we either make it completely independent (and follow the same config structure as for `test`) or we merge them with the general `test.` Is there a strong reason to keep them separate? My feeling is that it would be nicer to just merge them (e.g. same period).

vict0rsch: I kept the structure that was in the code with the `do_*()` scheme. Now wrt the period, that procedure is typically much longer I believe than the regular test is it not? I would _not_ generate 5k samples every 100 steps whild we might want to do other tests.

alexhernandezgarcia: I think most things at test time are similarly expensive or cheap depending on the environment. For example, sampling 5k samples is be a big deal in small environments. A period of 100 is fine for small problems but for bigger problems I use longer periods, like 500.

vict0rsch: I guess the question, is: do we want the flexibility or not? I would say yes but if you don't it's fine, your call, I'll just increase the shared test period and we'll have fewer data points for those :)

gflownet/envs/crystals/composition.py
alexhernandezgarcia, line 1: These are *all* the changes from [PR 132](https://github.com/alexhernandezgarcia/gflownet/pull/132), right? Or is there anything new here?

gflownet/envs/crystals/crystal.py
```diff
@@ -535,3 +541,242 @@ def readable2state(self, readable: str) -> List[int]:
             + self.space_group.readable2state(readables[2])
             + self.lattice_parameters.readable2state(readables[3])
         )
+
+    @torch.no_grad()
+    def top_k_metrics_and_plots(
```
alexhernandezgarcia, line 546: This is great, and actually great enough to not be here for crystals only but actually for all environments and proxies in general. In other words, is there anything here that is specific to the crystal environment or the crystal proxy? It seems to me that the answer is no, or if anything, it could be changed easily. Thus, I think this should go in `env/base.py` or (better) in a new `utils/plotting.py` (or the like). What do you think?

gflownet/envs/crystals/crystal.py
```diff
@@ -535,3 +541,242 @@ def readable2state(self, readable: str) -> List[int]:
...
+                **kwargs,
+            )
+            metrics.update(proxy_metrics)
+            figs += proxy_figs
+            fig_names += proxy_fig_names
+
+        return metrics, figs, fig_names
+
+    @torch.no_grad()
+    def compute_train_energy_proxy_and_rewards(self):
```
alexhernandezgarcia, line 750: This looks more like a proxy method than a env method, right? Actually, this would potentially break if the proxy changes. Also, I could use a brief docstring here.

gflownet/gflownet.py
```diff
@@ -754,10 +775,10 @@ def train(self):
...
             proxy_vals = self.env.reward2proxy(rewards).tolist()
             rewards = rewards.tolist()
-            self.buffer.add(states_term, trajs_term, rewards, proxy_vals, it)
-            self.buffer.add(
-                states_term, trajs_term, rewards, proxy_vals, it, buffer="replay"
-            )
+            # self.buffer.add(states_term, trajs_term, rewards, proxy_vals, it)
+            # self.buffer.add(
+            #     states_term, trajs_term, rewards, proxy_vals, it, buffer="replay"
+            # )
```
alexhernandezgarcia, line 781: This should be uncommented, right?

gflownet/envs/crystals/composition.py
vict0rsch: Yes. Not sure why it's showing up

gflownet/envs/crystals/crystal.py
vict0rsch: I agree. I'll do that

gflownet/envs/crystals/crystal.py
vict0rsch: Agreed

gflownet/gflownet.py
vict0rsch: yeah that's an artifact from the memory leak debugging

config/README.md
```diff
@@ -0,0 +1,3 @@
+# Hydra configuration files
+
+This directory contains `yaml` files used to parameterize `glflownet` trainings using Hydra.
```
alexhernandezgarcia, line 3:
```suggestion
This directory contains `yaml` files used to parameterize `gflownet` trainings using Hydra.
```

alexhernandezgarcia (approved):
Thanks for the contributions about the proxy and also to the general logistics, documentation, etc. There are a bunch of things I think I want to do differently, like the way train energy and rewards are computed and other logging operations, but we can change this later.

I will do some tests and merge. 

Sorry for the late review. It is in part to the size and diversity of changes. I personally prefer if the PRs are narrower in scope, but I am guilty of the opposite sometimes too.

PR #130, opened 2023-06-01
Speedup tetris environment

carriepl:
This PR aims to accelerate training when using the Tetris environment. 

Listed below are the three settings in which I tested this PR on one of the nodes (always the same in all my timings) of the cluster. Depending on the config, the speed up goes from around 3.5x, for very small and very restricted Tetris boards, to around 9x for the default Tetris config.

All the unit tests are currently passing and I added two additional test cases for scenarios in which there is room for a piece at the bottom of the Tetris board but there are obstacles overhead so that space is actually unreachable.

Test case 1 : env.width=6 env.height=6 env.pieces=["O","J","L"] env.rotations=[0,180]
Before PR : ~1.25s/it -> ~0.80it/s
With PR : ~2.80it/s

Test case 2 : env.width=6 env.height=6 (all pieces and rotations available)
Before PR : ~4.10s/it -> ~0.24t/s
With PR : ~1.00it/s

Test case 3 : default config (height=20, width=10, all pieces and rotations available)
Before PR : ~88s/it
With PR : ~10.00s/it

Note : Test case 1 is the only one that I used to identify and fix bottlenecks. Test cases 2 and 3 were only measured just before submitting the PR, I didn't optimize for them specifically. That means that it is likely that there would still be some gains on the table for larger Tetris environments.


gflownet/envs/tetris.py
```diff
@@ -138,30 +152,73 @@ def _drop_piece_on_board(
...
-                    return board_aux, True
-        return board, False
+
+        # Find the highest row occupied by any other piece in the columns where we wish
+        # to add the new piece
+        occupied_rows = board[:, col : col + wp].sum(1).nonzero()
+        if len(occupied_rows) == 0:
+            # All rows are unoccupied. Set first occupied row as the row "below" the
+            # board.
+            first_occupied_row = self.height
```
alexhernandezgarcia, line 171: "first" seems a bit ambiguous, no?
```suggestion
            highest_occupied_row = self.height
```

gflownet/envs/tetris.py
```diff
@@ -138,30 +152,73 @@ def _drop_piece_on_board(
...
+        else:
+            first_occupied_row = occupied_rows[0, 0]
```
alexhernandezgarcia, line 173: "first" seems a bit ambiguous, no?
```suggestion
            highest_occupied_row = occupied_rows[0, 0]
```

alexhernandezgarcia (approved):
This is great! It will be super helpful and I agree there's more room for improvement. It all looks good to me. I have only suggested a change of name of a variable, but no need to change if you disagree.

gflownet/envs/tetris.py
carriepl: I agree, it's a much better variable name. I made the change.

alexhernandezgarcia:
@carriepl I leave it to you to merge this PR whenever you think it's ready. To me it's ready now :)

PR #132, opened 2023-06-06
Speedup crystal env and bugfix

carriepl:
This PR fixes an issue with the crystal/composition environment where the environment would sometimes accept actions with too few atoms. This leads to the environment sometimes getting "stuck" because `max_diff_elem` has been reached but some atoms are still needed to reach `min_atoms`. I've added a unit test to check for this and implemented a fix.

The PR also includes some modifications to speedup the code. These modifications mostly concern the `get_mask_invalid_actions_forward()` methods because most of the runtime is spent in those.

I tested the speedup on the Dave proxy branch (https://github.com/alexhernandezgarcia/gflownet/pull/129) with the command `python main.py user=$USER +experiments=neurips23/crystal-comp-sg-lp.yaml device=cpu gflownet.optimizer.batch_size=32` and saw a speedup of around 1.7x in terms of the number of iterations per second.

### DIFF
--- a/gflownet/envs/crystals/composition.py
+++ b/gflownet/envs/crystals/composition.py
@@ -123,7 +123,7 @@ class Composition(GFlowNetEnv):
         assert self.max_diff_elem >= self.min_diff_elem
         assert self.max_atom_i >= self.min_atom_i
         valid_word_len = np.arange(self.min_atom_i, self.max_atom_i + 1)
-        actions = [(element, n) for n in valid_word_len for element in self.elements]
+        actions = [(element, n) for element in self.elements for n in valid_word_len]
         actions.append(self.eos)
         return actions
 
--- a/gflownet/envs/crystals/composition.py
+++ b/gflownet/envs/crystals/composition.py
@@ -143,7 +143,7 @@ class Composition(GFlowNetEnv):
         if done:
             return [True for _ in range(self.action_space_dim)]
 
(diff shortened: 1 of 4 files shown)


gflownet/envs/crystals/composition.py
```diff
@@ -123,7 +123,7 @@ def get_action_space(self):
         assert self.max_diff_elem >= self.min_diff_elem
         assert self.max_atom_i >= self.min_atom_i
         valid_word_len = np.arange(self.min_atom_i, self.max_atom_i + 1)
-        actions = [(element, n) for n in valid_word_len for element in self.elements]
+        actions = [(element, n) for element in self.elements for n in valid_word_len]
```
alexhernandezgarcia, line 126: Is this the bug fix?

alexhernandezgarcia (approved):
This looks good to me!

I have made some `black` and `isort` cosmetic changes on this branch in [PR 136](https://github.com/alexhernandezgarcia/gflownet/pull/136)

gflownet/envs/crystals/composition.py
carriepl: No, but it was a required change for my new reimplementation of composition.get_mask_invalid_actions_forward() which does fix the bug (and it much faster than the original). Having the actions element-contiguous allows get_mask_invalid_actions_forward() to easily set the mask for all actions related to an element all at once and save time in the process.

PR #133, opened 2023-06-08
Compute EMAs outside of pytorch

carriepl:
This PR fixes a memory leak issue that occured in the code.

Basically, the GFlowNet computes a running average of loss for the terminal and the non-terminal states. The issue is that this computation was done with pytorch tensors and without any detach. The result is that the whole pytorch computational graph is maintained in memory (because pytorch wants us to be able to take gradients on those EMAs) and grows very fast with every iteration.

I've modified the code to perform this computation outside of pytorch and validated that this fixes the issue. The figure below shows the memory consumption over time for experiments with the Dave Proxy on its own branch : 

image

From the graph, we see that the fix in this PR almost entirely fixes the issue. There is still some memory leak due to the main buffer that grows over training but it is much smaller in scale.

I've also tested the fix on the main branch with the ctorus environment and found the same result : fast increase in memory consumption without the fix, very small increase in memory consumption (most likely the main buffer) with the fix.

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -794,10 +794,10 @@ class GFlowNetAgent:
             # Moving average of the loss for early stopping
             if loss_term_ema and loss_flow_ema:
                 loss_term_ema = (
-                    self.ema_alpha * losses[1] + (1.0 - self.ema_alpha) * loss_term_ema
+                    self.ema_alpha * losses[1].item() + (1.0 - self.ema_alpha) * loss_term_ema
                 )
                 loss_flow_ema = (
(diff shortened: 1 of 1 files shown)


vict0rsch (approved):
LGTM. Maybe to keep in mind when we go to GPUs: calling `.item()` requires a change of device, maybe `.detach()` would be enough?

alexhernandezgarcia (approved):
Great find! This looks good to me. And I learn that I should better monitor the memory use to avoid introducing leaks in the future.

PS: I agree with Victor that `detach()` might be enough here. Also, as I noted elsewhere, the EMA of the loss is currently unused.

PR #138, opened 2023-06-12
Improving efficiency of checking validity of actions with the mask

alexhernandezgarcia:
### Context
In certain environments, one of the most computationally demanding operations is the computation of the mask of invalid actions. However, this computation is repeated in several parts of the code. For example, it is computed when sampling the actions, and it is computed again in the `step()` method.

On a related but separate note, due to a low-level [issue in Pytorch](https://github.com/pytorch/pytorch/blob/57c63aad10bd0ae69d96c418551b6a894e2c4b2a/aten/src/ATen/native/Distributions.cpp#L632), the [Categorical](https://pytorch.org/docs/stable/_modules/torch/distributions/categorical.html#Categorical.sample) distribution rarely but sometimes samples elements with zero probability (about 40-80 times per million) - at the end of this description is a piece of code that reproduces the issue. This can result in wasted computational resources since some actions sampled will end up being invalid (in the best case) or in silent errors (in a worse case).

### Goals 
The goal of this PR is two-fold:
- Enable options to skipping the computation of the mask (by calling `get_masks_invalid_actions_forward()`) in `step()`, for example when the user is confident than it is unnecessary and/or expensive. 
- Overcome the aforementioned issue with PyTorch by checking that sampled actions are indeed valid.

### Changes
- Whether the mask of invalid actions will be (re-)computed and the action checked in `step() is now controlled by two boolean variables:
    - A class instance attribute `self.skip_mask_check` set via the environment configuration
    - An argument `skip_mask_check` of `step().
    - Both are False (mask is computed) by default and if both or one of the two is True, then the computation of the mask (and the check) will be skipped. The logic behind this is that in order to skip, the user has to manually and explicitly set so. 
    - The reason to have two variables is that one (via the config) allows to set the option for all calls to `step()` in a run; the other one (argument of `step()`) allows to skip the computation in some calls to `step()` but not all.
- A change that has been on the to-do list for a while: the generic checks at the beginning of `step()` have been moved to a method in `_pre_step()`in  `env/base.py` so as to simplify the code of each environment and avoid repetitions.
- In `sample_actions()` of `gflownet/base.env`, we check that the action indices correspond to valid actions, otherwise we repeat up to `max_sampling_attempts` times (new argument to the method, set to 10 by default). If the attempts are exhausted, the method throws an error, which might help catching bugs in the future.
- Environments that have intermediate states that cannot "be done" (e.g. space group, and crystal) should overwrite `set_state()` and check whether the state and `done` are compatible.
- I have implemented a couple of generic environment methods to sample random actions and take the step (`step_random()`) and to sampler random trajectories (`trajectory_random()`). I have used these methods in the tests that sample random actions.
- Adapt some parts of the tests to deal with the above. 
- I added a missing check of `done` in the step method of the space group environment.

### Other notes
The PyTorch issue might be fixed in upcoming versions, but we will the check for now.

It is possible that the new Batch class may change some of this or improve things further by allowing us to eliminate other unnecessary calls to `get_masks_invalid_actions_*`.

An example of running time improvement by skipping the check of masks.

Tetris environment, 10x10 *with* mask check in `step()` (`skip_mask_check=True`): ~48 iterations/second
```
python main.py user=$USER env=tetris proxy=tetris gflownet=trajectorybalance env.width=10 env.height=10 logger.do.online=False env.skip_mask_check=False
```

Tetris environment, 10x10 *without* mask check in `step()` (`skip_mask_check=False`): ~32 iterations/second
```
python main.py user=$USER env=tetris proxy=tetris gflownet=trajectorybalance env.width=10 env.height=10 logger.do.online=False env.skip_mask_check=True
```
---
Piece of code to reproduce the issue explained in the Context.

```python
import torch
from tqdm import tqdm
from torch.distributions import Categorical

device_str = "cuda"
if device_str.lower() == "cuda" and torch.cuda.is_available():
    device = torch.device("cuda")
    print("Using CUDA")
else:
    device = torch.device("cpu")
    print("Using CPU")

float_precision = 32
if float_precision == 32:
    float_torch = torch.float32
elif float_precision == 64:
    float_torch = torch.float64
else:
    raise ValueError("Invalid float precision")
loginf = 10000
repetitions = 1e6
batch_size = 1000
n_batches = int(repetitions / batch_size)
repetitions = n_batches * batch_size
n_dim = 1848
n_valid = 6

mask_invalid = torch.ones(n_dim, dtype=torch.bool, device=device)
mask_invalid[:n_valid] = False
logits = torch.ones(n_dim, dtype=float_torch, device=device)
logits[mask_invalid] = -loginf

n_masked = torch.zeros(n_batches)
for idx in tqdm(range(n_batches)):
    logits_batch = logits.repeat((int(batch_size), 1))
    idx_sampled = Categorical(logits=logits_batch).sample()
    is_loginf = logits_batch[torch.arange(int(batch_size), device=device), idx_sampled] == -loginf
    n_masked[idx] = torch.sum(is_loginf)

n_masked_total = int(torch.sum(n_masked).item())
pct_masked = float(n_masked_total / repetitions) * 100
print(f"Masked action was sampled: {pct_masked} % ({n_masked_total}/{repetitions})")
```


carriepl:
Given the message on slack about a fix being merged in pytorch main but not yet released, the part of this PR about resampling actions might become unneeded. 
We don't know for certain when the next pytorch release is so should we switch to using the nightly build until the next release?

tests/gflownet/envs/common.py
```diff
@@ -223,14 +226,19 @@ def test__step__returns_same_state_action_and_invalid_if_done(env):
...
-    if torch.is_tensor(env.state):
-        assert torch.equal(next_state, env.state)
-    else:
-        assert next_state == env.state
-    assert action_step == action
-    assert valid is False
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        env.set_state(env.state, done=True)
+    if env.done:
```
carriepl, line 232: So this test now doesn't do anything for environments where the initial state cannot be done (like crystal and composition)? Should this state be modified so that it will actually test something on all envs?
I think we could modify the test to sample actions randomly from the mask of valid actions until we reach a state where we can do `env.set_state(env.state, done=True)` and `env.done` becomes True. And then we can do the rest of the test. This way, I think it will work for all environments. What do you think?

tests/gflownet/envs/test_crystals.py
```diff
@@ -251,11 +251,13 @@ def test__step__does_not_change_state_if_element_already_set(
...
 
-    assert action == invalid_action
-    assert not valid
-    assert state_after_valid == state_after_invalid
+# TODO: uncomment when step can handle invalid actions
+#     state_after_invalid, action, valid = env.step(invalid_action)
+
+#     assert action == invalid_action
+#     assert not valid
+#     assert state_after_valid == state_after_invalid
```
carriepl, line 260: A bit of a general question on this PR... how damageable (or not) is it that the step() function doesn't ensure that the incoming action is valid anymore?
Was the CategoricalSampling the only possible source of invalid actions?
If all actions are sampled using the mask then I think it should be fine but I don't know the code base well enough to garantee that it is the case.
For instance, if we can have actions entered by a user or actions specified by a replay buffer stored on disk (and which could, therefore, contain actions that were valid under a previous version of the env but not the current one), then there could be an issue.

alexhernandezgarcia: As far as I can recall, the current typical pipeline to train GFlowNet should never sample and execute actions without a mask. However, the code allows using parts of it outside this pipeline and future code may call the step function without having computed a mask. This is why I am not fully satisfied with the removal of the check.

> Was the CategoricalSampling the only possible source of invalid actions?
Hard to tell.

tests/gflownet/envs/common.py
alexhernandezgarcia: I had thought of doing this but was lazy about it :sweat_smile: Since you said it too, I have done it. 

While doing that, I thought of implementing a couple of helper functions to sample random steps and trajectories, which I have used in other tests, so the PR has grown a little. Sorry for the mess.

tests/gflownet/envs/test_crystals.py
alexhernandezgarcia: What do you think about the tasks added in the PR to sort this out? Namely:
- Implement "safe mode", instead of completely removing the mask check in step()
- Add info about "safe mode" in the README

carriepl: I agree, these two things would do a lot to mitigate the risk of removing this layer of security. I think it's a good idea.

gflownet/envs/base.py
```diff
@@ -295,6 +295,53 @@ def get_logprobs(
...
+        actions, _ = self.sample_actions(
+            policy_outputs=random_policy, mask_invalid_actions=mask_invalid
+        )
+        action = actions[0]
+        return self.step(action)
+
+    def trajectory_random(self):
+        """
+        Samples a random trajectory, by sampling random actions until an EOS action is
+        sampled.
```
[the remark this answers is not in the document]
alexhernandezgarcia: Good point. Done.

gflownet/envs/base.py
```diff
@@ -295,6 +295,53 @@ def get_logprobs(
...
+        state : list
+            The final state.
+
+        action: list
+            The list of actions (tuples) in the trajectory.
+        """
+        actions = []
+        while self.done is not True:
+            _, action, valid = self.step_random()
+            if valid:
```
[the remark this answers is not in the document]
alexhernandezgarcia: I don't think so, because this depends on `step_random()`, which always obtains the mask and uses it to sample the random action. Do you agree?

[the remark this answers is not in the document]
carriepl: I agree with you which is why I raised that point. Therefore, I think that the `if` is unneeded and you can simply always append the action to the list.

gflownet/envs/base.py
```diff
@@ -207,19 +218,51 @@ def step(self, action: Tuple[int]) -> Tuple[List[int], Tuple[int], bool]:
...
+    ) -> Tuple[List[int], Tuple[int], bool]:
+        """
+        Executes step given an action.
+
+        Args
+        ----
+        action : tuple
+            Action from the action space.
+
+        skip_mask_check : bool
```
carriepl, line 248: The argument `skip_mask_check` has been added to the docstring here but many of the other environments do not have it in the docstring of their own `step()` function.

gflownet/envs/base.py
```diff
@@ -207,19 +218,51 @@ def step(self, action: Tuple[int]) -> Tuple[List[int], Tuple[int], bool]:
...
+            return False, self.state, action, False
         # If action is in invalid mask, exit immediately
-        if self.get_mask_invalid_actions_forward()[action_idx]:
-            return self.state, action, False
-        return None, None, None
+        if not (self.skip_mask_check or skip_mask_check):
+            action_idx = self.action_space.index(action)
+            if self.get_mask_invalid_actions_forward()[action_idx]:
+                return False, self.state, action, False
+        return True, self.state, action, True
```
carriepl, line 234: In all of this code, the outputs `do_step` and `valid` are always identical. Is there a reason to have both outputs?

alexhernandezgarcia: I thought of that too. The main reason is that in the future `_pre_step()` could be extended (or overwritten by certain environments) in ways that could yield a result with different `do_step` and `valid`. Another reason is that semantically they are different.

gflownet/envs/base.py
alexhernandezgarcia: Thanks for spotting this!

gflownet/envs/base.py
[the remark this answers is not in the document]
alexhernandezgarcia: Ah I get you. This is a subtle point. I think that it's good to keep the notion of `valid` inside the `step()` method as separate from the notion of valid as per the masks. Why? For example, we know that computing the mask can be very expensive, because we need to compute validity for each action of possibly many. Therefore, in some environments the mask might not ultimately determine whether an action is valid, but rather it is designed to determine which actions are _invalid_, to filter out actions (hence the word "invalid" in the variable name). It is in the `step()` action, that validity is ultimately determined after _attempting_ to execute the action.

Long story short: I believe we should keep checking `valid` in `step()` and in `trajectory_random()` before adding to the list of actions.

Does that make sense? Do you agree?

gflownet/envs/base.py
alexhernandezgarcia: Done

tests/gflownet/envs/test_crystals.py
alexhernandezgarcia: Done now

gflownet/envs/base.py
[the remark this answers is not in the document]
carriepl: So you're saying that there are cases where an action might be invalid even though the mask says it is valid.

Is that only for environments with continuous actions spaces or there are other cases?

If it can happen on environments with discrete action spaces I wonder whether something should be changed in the code to handle those.
I'm just thinking of a scenario in which the GFN assigns almost all probability mass to an action which is valid according to the mask but that the environment refuses. It could be stuck in this loop of "get mask, sample action, attempt action unsucceccfully" for a while.

At any rate, even if this is an issue, I think that it falls outside of the scope of your current PR. I think that my original comment has been addressed.

gflownet/envs/plane.py
```diff
@@ -380,6 +380,10 @@ def step(
             Action to be executed. An action is a tuple with two values:
             (dimension, increment).
 
+        skip_mask_check : bool
+            If True, skip computing forward mask of invalid actions to check if the
+            action is valid.
```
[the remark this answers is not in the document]
alexhernandezgarcia: Thanks, you're right. If replaced the docstring of this argument by "Ignored" in the continuous envs.

gflownet/envs/base.py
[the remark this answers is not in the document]
alexhernandezgarcia: > So you're saying that there are cases where an action might be invalid even though the mask says it is valid.

Almost. To be more precise, I say that an action might be invalid even though the mask says it is *not* invalid.

> Is that only for environments with continuous actions spaces or there are other cases?

No, I expect this to happen mostly in environments where computing the mask is expensive, for example the crystal.

To add a bit of an answer to the rest of your comments, which is relevant: I think of all this this way: a vanilla implementation of GFlowNet does not use any masks at all. This is not wrong, it's just slower to train (potentially too slow) because it's missing an important inductive bias and it has to learn that some actions are impossible (ideally masked). One can add _some_ masks of *invalid* actions to make things better. But, it is conceivable to think that we cannot catch all *invalid* masks easily, especially if there are too many actions. That's why the masks is about invalid actions, not about valid actions. One level more is to check, when we attempt to execute one step with one specific action, if the action is indeed valid (it may be cheap enough to check validity for one action, but too expensive to do it for all actions.

All this is mostly theory, I don't mean that all of it is currently implemented, but it's somehow behind some design choices I have made.

gflownet/envs/base.py
alexhernandezgarcia: Right, the only example I can think of is when the action is EOS. The action will be valid but _there is no need to proceed further_. Anyway, you have almost convinced me that it might be unnecessary to have both outputs. Should we keep only `do_step`?

alexhernandezgarcia: Ok, changed [here](https://github.com/alexhernandezgarcia/gflownet/pull/138/commits/c5011f35d20a601c7806b2f9693051a58e84c629)

carriepl (approved):
Looks pretty good to me. Nice work!

gflownet/envs/base.py
```diff
@@ -242,8 +282,27 @@ def sample_actions(
             logits = policy_outputs
             logits /= temperature_logits
         if mask_invalid_actions is not None:
+            assert not torch.all(
+                mask_invalid_actions
+            ), """
+            All actions in the mask are invalid.
```
vict0rsch, line 288: Maybe wrap this in `textwrap.dedent` ? otherwise this will print weird indents:

```python
In [1]: print("""
   ...:             All actions in the mask are invalid.
   ...:             """)

            All actions in the mask are invalid.
            

In [2]: from textwrap import dedent

In [3]: print(dedent("""
   ...:             All actions in the mask are invalid.
   ...:             """))

All actions in the mask are invalid.

In [4]: repr("""
   ...:             All actions in the mask are invalid.
   ...:             """)
Out[4]: "'\\n            All actions in the mask are invalid.\\n            '"

In [5]: repr(dedent("""
   ...:             All actions in the mask are invalid.
   ...:             """))
Out[5]: "'\\nAll actions in the mask are invalid.\\n'"
```

Importantly the `repr` shows that the current `print` will leave trailing whitespaces that will affect the next one

vict0rsch: Or just remove the `"""` and make it a single line str :)

gflownet/envs/base.py
```diff
@@ -242,8 +282,27 @@ def sample_actions(
...
+        else:
+            mask_invalid_actions = torch.zeros(
+                policy_outputs.shape, dtype=torch.bool, device=device
+            )
+        # Make sure that a valid action is sampled, otherwise throw an error.
+        for _ in range(max_sampling_attempts):
+            action_indices = Categorical(logits=logits).sample()
+            if not torch.any(mask_invalid_actions[ns_range, action_indices]):
+                break
+        else:
```
vict0rsch, line 300: Am I seeing this misindented on Github for some reason or is it a bug?

SCR-20230626-mqby

gflownet/envs/base.py
```diff
@@ -186,39 +189,75 @@ def get_parents(
...
             Action from the action space.
 
+        skip_mask_check : bool
+            If True, skip computing forward mask of invalid actions to check if the
+            action is valid.
+
         Returns
         -------
+        do_step : bool
+            If True, step() should continue further, False otherwise.
```
vict0rsch, line 210: Should `do_step` be returned last? It's purely cosmetic but it would yield an interface similar to `step()` that returns `(state, action, valid)`

gflownet/envs/crystals.py
```diff
@@ -332,17 +334,31 @@ def step(self, action: Tuple[int, int]) -> Tuple[List[int], Tuple[int, int], boo
...
             return self.state, self.eos, valid
 
+    def set_state(self, state: List, done: Optional[bool] = False):
+        """
+        Sets the state and done. If done is True but incompatible with state (space
+        group is missing), then force done False and print warning.
+        """
+        state_elem = [self.idx2elem[i] for i, e in enumerate(state) if e > 0]
+        n_state_atoms = sum(state)
+        if done == True and not self._can_eos(state, state_elem, n_state_atoms):
```
vict0rsch, line 352: Cosmetic again, but the proper `True` check in Python is `if var is True` not `if var == True`

gflownet/envs/crystals.py
```diff
@@ -332,17 +334,31 @@ def step(self, action: Tuple[int, int]) -> Tuple[List[int], Tuple[int, int], boo
...
+            done = False
+            warnings.warn(
+                f"""
+            Attempted to set state {self.state2readable(state)} with done = True, which
```
vict0rsch, line 356: idem, dedent multilines IMHO

gflownet/envs/crystals.py
```diff
@@ -369,3 +385,16 @@ def _can_produce_neutral_charge(self, state: Optional[List[int]] = None) -> bool
...
+    def _can_eos(self, state: List, state_elem: List, n_state_atoms: int):
+        """
+        Returns True if the end of sequence action (EOS) can be chosen from the
+        state passed as an argument. False otherwise.
+        """
+        if n_state_atoms < self.min_atoms:
+            return False
+        if len(state_elem) < self.min_diff_elem:
+            return False
+        if any(r not in state_elem for r in self.required_elements):
```
vict0rsch, line 398: Probably minor because `state_elem` is small but I'm always uncomfortable with `if el in list` vs `if el in set`. Feel free to just resolve without taking any action. I think the overhead of `set` creation is not worth it here.

```python
In [8]: def f(l1, l2):
   ...:     return any(r not in l1 for r in l2)
   ...: 

In [9]: def h(l1, l2):
   ...:     s1 = set(l1)
   ...:     return any(r not in s1 for r in l2)
   ...: 

In [10]: import numpy as np

In [11]: l1, l2 = np.random.randint(0, 100, (2, 30))

In [12]: %timeit f(l1, l2)
4.12 µs ± 90.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [13]: %timeit h(l1, l2)
3.04 µs ± 173 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [14]: l1, l2 = np.random.randint(0, 100, (2, 3000))

In [15]: %timeit f(l1, l2)
14.2 ms ± 3.48 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [16]: %timeit h(l1, l2)
521 µs ± 21.7 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

File: gflownet/envs/spacegroup.py [hunk 14]
```diff
@@ -370,6 +373,21 @@ def step(self, action: Tuple[int, int]) -> Tuple[List[int], Tuple[int, int], boo
     def get_max_traj_length(self):
         return 3
 
+    def set_state(self, state: List, done: Optional[bool] = False):
+        """
+        Sets the state and done. If done is True but incompatible with state (space
+        group is missing), then force done False and print warning.
+        """
+        if done == True and state[self.sg_idx] == 0:
```
vict0rsch, line 381: Idem `True`

vict0rsch (approved):
LGTM, minor things (except if my screenshot is a bug in the code not the rendering)

gflownet/envs/README.md
```diff
@@ -0,0 +1,11 @@
+# GFlowNet environments
```
AlexandraVolokhova, line 1: Could you add here a paragraph about the meaning of the mask_invalid_actions_forward and mask_invalid_actions_backward, which you told us about today (e.g. that non-invalid doesn't mean the action is valid)? It would be very useful to have this information documented here

gflownet/envs/base.py
vict0rsch:
```
        for _ in range(max_sampling_attempts):
            action_indices = Categorical(logits=logits).sample()
            if not torch.any(mask_invalid_actions[ns_range, action_indices]):
                logprobs = self.logsoftmax(logits)[ns_range, action_indices]
                # Build actions
                actions = [self.action_space[idx] for idx in action_indices]
                return actions, logprobs
        raise ValueError(
            f"No valid action could be sampled after {max_sampling_attempts} attempts."
        )
```

(or use `dedent` with `from textwrap import dedent` at the top of the file)

alexhernandezgarcia: Is your comment about the `else` part? If so, it is correct as you see it. It's a `for ... else` loop. If the `break` clause is not reached during the for loop, then the code gets into the `else` part. So this piece of code attempts to `max_sampling_attempts` times to sample actions that are not invalid, otherwise raises an error.

```python
import numpy as np
max_n = 10
threshold = 0.9
for _ in range(max_n):
    if np.random.uniform() > threshold:
        break
else:
    print(f"No random number in [0, 1] larger than {threshold} in {max_n} uniform samples")
```

gflownet/envs/base.py
alexhernandezgarcia: I put it first because `do_step` is the main output of `_pre_step` and to actually make it distinct to `valid`. The value of `do_step` is very much related to `valid`, but semantically it is different. So I would prefer to leave it as currently is.

gflownet/envs/base.py
alexhernandezgarcia: Good point, thanks! I have wrapped the string with `dedent` in [this commit](https://github.com/alexhernandezgarcia/gflownet/pull/138/commits/3b6ff1fbef5e168c8bcd6fc8c71dc11cd197b49a).

gflownet/envs/crystals.py
alexhernandezgarcia: Thanks, added `dedent` [here](https://github.com/alexhernandezgarcia/gflownet/pull/138/commits/9180ad18eff7ac81a8f7012fe271e139ea47a4ef).

gflownet/envs/crystals.py
alexhernandezgarcia: Sure, changed [here](https://github.com/alexhernandezgarcia/gflownet/pull/138/commits/1c85f2be20ac63e8903c571926f92bb1247b9ac2).

alexhernandezgarcia: (this is part of another PR merged here though)

gflownet/envs/crystals.py
alexhernandezgarcia: This is part of another PR merged here, but what is the problem with `if el in list`?

gflownet/envs/spacegroup.py
alexhernandezgarcia: done [here](https://github.com/alexhernandezgarcia/gflownet/pull/138/commits/a4943cba21790dc2bbaa09b77897369a4711fc84)

gflownet/envs/README.md
alexhernandezgarcia: Yes, addded [here](https://github.com/alexhernandezgarcia/gflownet/pull/138/commits/d974c50a29e1fcbda01859967835cfa9e44fab00).

gflownet/envs/base.py
vict0rsch: Amazing I didn't know about that!!

Researching this, I also learned about `try/except/*else*/finally` clauses

```python
try:
    a()
except Exception:
    # executed if a() throws an exception
    b()
else:
	# executed if a() does not throw an exception BUT c() won't be excepted
    c()
finally:
    # always executed
    d()
```

gflownet/envs/crystals.py
vict0rsch: `if el in list` is `O(len(list))` but `if el in set` is `O(1)` so if you want to check for presence in a list >1 times, it's worth turning it into a `set` first. Obviously the larger the list, and the more you test `in` the more it saves you time

gflownet/envs/base.py
alexhernandezgarcia: I never thought I would teach you any Python! :P 

I've also just learnt about `try/except/*else*/finally`

gflownet/envs/crystals.py
vict0rsch: https://wiki.python.org/moin/TimeComplexity

alexhernandezgarcia: Oh I see! I knew about this but I haven't incorporated it into my practices. There may be a few places in the repo where we check long lists, so something to review at some point. Thanks! And sorry I didn't understand initially, got confused by something.

PR #140, opened 2023-06-21
RF-GFN

michalkoziarski:


tests/gflownet/envs/common.py
```diff
@@ -204,7 +204,15 @@ def test__get_parents__returns_same_state_and_eos_if_done(env):
...
-        assert all([torch.equal(p, env.state) for p in parents])
+        assert all(
+            [torch.equal(torch.isnan(p), torch.isnan(env.state)) for p in parents]
+        )
+        assert all(
+            [
+                torch.equal(p[~torch.isnan(p)], env.state[~torch.isnan(env.state)])
+                for p in parents
+            ]
+        )
```
alexhernandezgarcia, line 215: Why was this needed?

alexhernandezgarcia: Ah it's because the states of the tree are initialised to `torch.nan`?

michalkoziarski: Yes, and (get ready to be amazed):
```
>>> torch.nan == torch.nan
False
```

alexhernandezgarcia: LOL

alexhernandezgarcia: I guess it makes sense (ish) if I think about it... At least it's consistent with numpy:
```
>>> numpy.inf * 0 == numpy.inf * 0
False
```

michalkoziarski: Well riddle me this:
```
>>> torch.inf == torch.inf
True
```
and
```
>>> torch.nan in [torch.nan]
True
```

alexhernandezgarcia: hahaha I can see the logic of either way...

gflownet/envs/tree.py
```diff
@@ -0,0 +1,614 @@
...
+    def step(
+        self, action: Tuple[int, int, float]
+    ) -> Tuple[List[int], Tuple[int, int, float], bool]:
+        # If action not found in action space raise an error.
+        if action[:2] not in self.action_space:
+            raise ValueError(
+                f"Tried to execute action {action} not present in action space."
+            )
+        else:
+            action_idx = self.action_space.index(action[:2])
```
alexhernandezgarcia, line 355: Why `action[:2]` instead of just `action`?

alexhernandezgarcia: Ok, got it by myself - actions are tuples of three things. I got confused by the dosctring in the action space method. I am updating it.

tests/gflownet/envs/common.py
alexhernandezgarcia: I have incorporated the check of NaNs in `env.equal(x, y)` so it's easier to check if two states are the same also with nans. See [here](https://github.com/alexhernandezgarcia/gflownet/commit/c347096a001a36858d99b13f6c3223b66ddff2b0)

alexhernandezgarcia:
Test passed after merging locally :+1: Will merge :)

alexhernandezgarcia (approved):
Looking goooood!

PR #143, opened 2023-06-26
Remove batch class get_mask_invalid_actions_forward() calls

carriepl:
This PR aims to reduce the number of calls to the `get_mask_invalid_actions_forward()` method of environment classes. 

Currently, whenever the GFN takes an action on an environment, it involves computing a mask of invalid actions, sampling one of the non-invalid actions, and calling `Batch.add_to_batch()` which will compute the mask of invalid actions a second time. With this PR, the mask of invalid actions is computed only once and provided as an optional argument to `add_to_batch()`.

I tested on the Tetris environments (which, at larger env sizes, has its runtime completely dominated by calls to `get_mask_invalid_actions_forward()` ) with the following command : 
`python main.py user=$USER env=tetris proxy=tetris gflownet=trajectorybalance env.width=10 env.height=10 logger.do.online=False`

Without this PR : ~6.0 seconds/iteration
With this PR : ~4.1 seconds/iteration

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -439,7 +439,9 @@ class GFlowNetAgent:
             envs, actions, valids = self.step(envs, actions, is_forward=True)
             # Add to batch
             t0_a_envs = time.time()
-            batch.add_to_batch(envs, actions, valids, train)
+            batch.add_to_batch(
+                envs, actions, valids, mask_invalid_actions_forward, train
+            )
             # Filter out finished trajectories
             envs = [env for env in envs if not env.done]
             t1_a_envs = time.time()
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -282,7 +282,7 @@ class GFlowNetAgent:
             mask_invalid_actions,
             temperature,
         )
-        return actions
+        return actions, mask_invalid_actions
 
     def step(
         self,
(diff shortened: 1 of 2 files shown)


alexhernandezgarcia (approved):
Awesome, thanks! It all looks good to me. [PR #111](https://github.com/alexhernandezgarcia/gflownet/pull/111) is already merged and there were a few changes after the creation of your branch, so there will be a couple of conflicts related to changes of variable names and typing in arguments.

carriepl:
With the Batch class merged, I rebased on top of the current main and redid my timings (the Batch branch I used previously didn't have a lot of the speedups that were recently merged into main).
@alexhernandezgarcia Since I had to change the code to solve conflicts and also added some comments, can you spare a look even though you already approved it?

alexhernandezgarcia (approved):
I have re-reviewed the changes and it looks good to me and ready to merge!

And wow, huge speed up with the tetris environment after all the changes! :heart_eyes:

AlexandraVolokhova (approved):
Looks good to me too!

carriepl:
> I have re-reviewed the changes and it looks good to me and ready to merge!
> 
> And wow, huge speed up with the tetris environment after all the changes! 😍

Yeah, the speedups from your mask PR and mine are fully additive so we end up with some experiments on Tetris that get about a 3x speedup! The crystal env is also pretty limited by the speed at which we compute the masks so we might see a 2-3x speedup there also.

With all those PRs merged, I'll re-profile the code to see the changes.

PR #145, opened 2023-06-27
Revert "Remove batch class get_mask_invalid_actions_forward() calls"

alexhernandezgarcia:
Reverts alexhernandezgarcia/gflownet#143

The PR removes the call to `get_mask_invalid_actions_forward()` inside the Batch class by using instead the masks computed when sampling the actions.

But that's not correct. The reason is that the masks used when sampling the actions for transitions between (say) t -> t+1. So those masks correspond to states at t. However, then steps are taken and we obtain the states at t+1, which is what needs to be added to the batch. And we need to store the masks of states at t+1 but we are now passing instead the masks of states at t.
That's why the masks were recomputed. 

This doesn't mean we cannot make the code more efficient, but it has to be the other way around: the batch computes the masks of states at t, then we re-use them to sample transitions t -> t+1. Let's do this in a new PR.

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -282,7 +282,7 @@ class GFlowNetAgent:
             mask_invalid_actions,
             temperature,
         )
-        return actions, mask_invalid_actions
+        return actions
 
     def step(
         self,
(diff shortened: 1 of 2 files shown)


carriepl (approved):
LGTM

PR #146, opened 2023-06-28
Remove batch class mask calls - Follow up

carriepl:
New, and hopefully improved, version of #143.

In this version, `GFlowNet.sample_actions()` is modified such that : 
- After actions are taken on the environments, new actions masks are computed
- These masks are given to Batch.add_to_batch to avoid recomputation in the batch class
- These masks are given to Env.sample_actions() the next time actions are sampled on these environments.

I tested on the ctorus environment with the following command : 
`python main.py user=$USER +experiments=icml23/ctorus device=cpu env.reward_beta=1`
and validated that it runs and that it produces the exact same graphs of metrics (loss, ...) over iterations in WandB with and without the PR

I also tested on the Tetris environments with the following command :
`python main.py user=$USER env=tetris proxy=tetris gflownet=trajectorybalance env.width=10 env.height=10 logger.do.online=False`
Without this PR : ~7.2 seconds/iteration
With this PR : ~4.8 seconds/iteration

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -209,6 +209,7 @@ class GFlowNetAgent:
         is_forward: bool = True,
         temperature=1.0,
         random_action_prob=0.0,
+        mask_invalid_actions=None,
     ):
         """
         Samples one action on each environment of a list.
@@ -233,6 +234,10 @@ class GFlowNetAgent:
 
         temperature : float
             Temperature to adjust the logits by logits /= temperature
+
+        mask_invalid_actions : list or None
+            List of invalid action masks for the environments in env. Optional, will be
+            computed if not provided.
         """
         # TODO: implement backward sampling from forward policy as in old
         # backward_sample.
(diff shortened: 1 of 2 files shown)


gflownet/gflownet.py
```diff
@@ -242,14 +247,22 @@ def sample_actions(
...
+            else:
+                mask_invalid_actions = tbool(
+                    [env.get_mask_invalid_actions_backward() for env in envs],
+                    device=self.device,
+                )
         else:
+            # Invalid action masks are provided, convert to pytorch tensors
             mask_invalid_actions = tbool(
-                [env.get_mask_invalid_actions_backward() for env in envs],
+                [m for m in mask_invalid_actions],
```
alexhernandezgarcia, line 265: Why do you need to iterate over `mask_invalid_actions` instead of converting calling `tbool` directly on it?
```suggestion
                mask_invalid_actions,
```
(the suggestion would need a `black` though.

gflownet/gflownet.py
```diff
@@ -424,6 +438,7 @@ def sample_batch(
                         is_forward=True,
                         temperature=1.0,
                         random_action_prob=self.random_action_prob,
+                        mask_invalid_actions=masks_invalid_actions_forward,
```
alexhernandezgarcia, line 441: I would simply pass `None` and remove the definition of `masks_invalid_actions_forward = None` in line 428 above.

gflownet/gflownet.py
```diff
@@ -434,14 +449,39 @@ def sample_batch(
                         is_forward=True,
                         temperature=self.temperature_logits,
                         random_action_prob=self.random_action_prob,
+                        mask_invalid_actions=masks_invalid_actions_forward,
```
alexhernandezgarcia, line 452: Idem. I would simply pass `None` and remove the definition of `masks_invalid_actions_forward = None` in line 428 above.

alexhernandezgarcia (approved):
See my (minor) comment, but it looks good to me.

As a note, I am currently making some changes in the Batch that will make it more flexible and among other things enable easily retrieving the masks from it. I think that will simplify the code a bit in the future, but for now this is a good solution!

gflownet/gflownet.py
carriepl-mila: Ha! I was definitely half asleep here. You're absolutely right, I'll change that.

carriepl-mila: Done.

carriepl-mila:
> See my (minor) comment, but it looks good to me.
> 
> As a note, I am currently making some changes in the Batch that will make it more flexible and among other things enable easily retrieving the masks from it. I think that will simplify the code a bit in the future, but for now this is a good solution!

I think I could have also modified step() to take the masks as optional inputs but I didn't do it because it felt redundant with your PR that was merged recently.  I guess the advantage would be that it would be a little safer than completely skipping the mask checking.

PR #147, opened 2023-06-30
Change loginf = -1000 by torch.inf and simplify part of flow matching loss

alexhernandezgarcia:
### Context

The current code uses the value -1000 in place of -infinity for, for example, the logits of invalid actions. However, there is no reason to not use -infinity instead and actually this may cause problems if for some reason the logits of valid actions become close or even lower than -1000. 

### Changes

This PR replaces the -1000 (via the `loginf` variable) by `torch.inf`. This actually simplifies the code a bit, since `loginf` was an argument of a few methods, without really being used (the default 1000 was used).

Furthermore, I have taken the opportunity to simplify the last part of the calculation of the flow matching loss, since simply replacing -1000 by -infinity would have made the code crash due to nan losses. Now, the flow matching loss is computed as the combination of the loss on the terminating states and the loss on the intermediate states. This is more readable, potentially slightly faster (although some quick tests show no noticeable difference), and it will be easy to add an option to weight differently the contribution of each loss (terminating vs. intermediate states).

### Sanity checks

- [10x10 grid FM wandb report](https://wandb.ai//alexhg/Sanity%20checks/reports/Sanity-check-loginf-to-torch-inf-10x10-grid---Vmlldzo0Nzc2MzU5)
- [continuous torus wandb report](https://wandb.ai//alexhg/Sanity%20checks/reports/Sanity-check-loginf-to-torch-inf-cont-torus---Vmlldzo0Nzc2MzUz)

---

@carriepl I believe this should resolve the issues you observed training with the tetris environment, but I have not been able to reproduce the problem.

### DIFF
--- a/gflownet/envs/base.py
+++ b/gflownet/envs/base.py
@@ -270,7 +270,6 @@ class GFlowNetEnv:
         sampling_method: str = "policy",
         mask_invalid_actions: TensorType["n_states", "policy_output_dim"] = None,
         temperature_logits: float = 1.0,
-        loginf: float = 1000,
         max_sampling_attempts: int = 10,
     ) -> Tuple[List[Tuple], TensorType["n_states"]]:
         """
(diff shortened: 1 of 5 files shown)


carriepl (approved):
Looks good to me. And making the code simpler is a definite plus.

PR #150, opened 2023-07-14
Batch v2 and backward trajectories

alexhernandezgarcia:
## Summary

The core of this PR is to adapt the `Batch` class to make it more flexible and extendable for future improvements. In particular, Batch instances do not depend any longer on the specific loss to store, retrieve or transform its elements in one way or another. Instead, I have implemented `get_*()` methods that retrieve the requested elements in the specified format (via arguments) from the batch. Another difference with the previous version is that there is no longer a need to "process the batch" (call `process_batch()` that results in a change of types of the elements in the batch. Now internally the types are consistent in the batch, and externally the getter methods are used to retrieve what is needed.

Another important focus of the work in this PR has been enabling backward sampling and training with backward trajectories. I have tried to (re-)implement it in a way that from the outside (from the GFlowNet agent) the interaction with the Batch is almost identical regardless of whether the trajectories are sampled backward or forward. This has simplified a lot the code of, for instance, `GFlowNetAgent.sample_batch()` and `GFlowNetAgent.sample_actions()` and will be simplified even further in the future.

Note that I have introduced a number of "collateral improvements" as I implemented the above. For example, new methods in the base Environment, a few fixes of non-crucial methods, changes in the Buffer, etc.

Finally, I have written extensive tests for the Batch that make me more confident that the core aspects are likely correct. The test include forming batches with single, multiple trajectories, forward, backward and mixed, and with multiple environments:

- Grid: discrete, states are lists
- Tetris: discrete, states are tensors
- Continuous torus: continuous

## Anticipating Batch v3

A lot of the changes I have made in this PR are introduced anticipating a future version ("v3") of the Batch class. The main idea will be to implement a cache-enabled Batch. Currently there is a lot of redundancy in what is stored in the batch and also the computations done to retrieve the requested elements to sample trajectories and train GFlowNet. I believe it will be relatively straightforward to extend the current version ("v2") in such a way that items are not stored multiple times if they are identical and instead references to them are kept.

On a different note but related to the batch, I think this future Batch will also be very close to having the functionality we need from the Buffer. Therefore, I am also thinking of implementing the functionality that is currently in the Buffer into the Batch. I think this will simplify the API a lot and enable computational advantages.

## Backward trajectories

I have made a lot of changes to (re-)enable training with backward trajectories, either from a training set or from a replay buffer. The type (forward, train set, replay buffer) and number of trajectories to sample at each training iteration can now be controlled with:

- `gflownet.optimizer.batch_size.forward` (default 10)
- `gflownet.optimizer.batch_size.train` (default 0)
- `gflownet.optimizer.batch_size.replay` (default 0)

## Sanity checks

I have run a bunch of GFlowNets on the grid, continuous torus and tetris environments to ensure the results are the same as before the implementation of the Batch class. The runs are on wandb [here](https://wandb.ai/alexhg/sanity_checks).

## Other notes

- I have left a lot of `TODO` comments in the code, pointing to things that will mostly be done in the upcoming Batch v3.
- Backward sampling does not work yet with the tetris environment because the states are tensors and I need to adapt the code.
- Backward sampling does not work yet with continuous environments that need to implement `step_backward()`.
- Merging batches is now more flexible and this might enable parallel sampling that could be easily merged.

## Tests and style

- [x] `python -m pytest -s tests/gflownet/utils/test_batch.py` :heavy_check_mark: 
- [x] `python -m pytest -s tests/gflownet/envs/` :heavy_check_mark: 
- [x] `black .`
- [x] `isort --profile black`


config/gflownet/gflownet.yaml
```diff
@@ -19,8 +19,11 @@ optimizer:
   adam_beta2: 0.999
   # Momentum for SGD
   sgd_momentum: 0.9
-  # Mini-batch size
-  batch_size: 10
+  # Number of trajectories of each kind
+  batch_size:
+    forward: 10
```
michalkoziarski, line 24: Could you clarify what are the specific batch sizes (forward, train, replay)?

gflownet/envs/base.py
```diff
@@ -223,14 +258,23 @@ def _pre_step(
             raise ValueError(
                 f"Tried to execute action {action} not present in action space."
             )
-        # If env is done, step should not proceed.
-        if self.done:
-            return False, self.state, action
+        # If backward and state is source, step should not proceed.
+        if backward is True:
```
michalkoziarski, line 262:
```suggestion
        if backward:
```
Or is there a specific reason to have `is`?

gflownet/envs/base.py
```diff
@@ -264,6 +308,50 @@ def step(
...
+        """
+        Executes a backward step given an action. This generic implementation should
+        work for all discrete environments, as it relies on get_parents(). Continuous
+        environments should re-implement a custom step_backwards(). Despite being valid
+        for any discrete environment, the call to get_parents() may be expensive. Thus,
+        it may be advantageous to re-implement step_backwards() in a more efficient
+        way as well for discrete environments. Especially, because this generic
+        implementation will make two calls to get_parents - once here and one in
+        _pre_step() through the call to get_mask_invalid_actions_backward() if
+        skip_mask_check is True.
```
michalkoziarski, line 323: Would it be possible to have some check for continuous environments having implemented this? For example, having separate `DiscreteEnvironment` and `ContinuousEnvironment` classes (from which all envs inherit), and forcing the continuous ones to implement this? In general, this could be a blueprint for what a user needs to implement itself for all of this to work. I am mostly talking from a point of view of someone new not being surprised that the things randomly don't work because they forgot to implement something.

gflownet/gflownet.py
```diff
@@ -202,66 +194,127 @@ def parameters(self):
...
+        backward: Optional[bool] = False,
+        temperature: Optional[float] = 1.0,
+        random_action_prob: Optional[float] = 0.0,
+        no_random: Optional[bool] = True,
+        times: Optional[dict] = None,
+    ) -> List[Tuple]:
         """
-        Samples one action on each environment of a list.
+        Samples one action on each environment of the list envs, according to the
+        sampling method specify by sampling_method.
```
michalkoziarski, line 208:
```suggestion
        sampling method specified by sampling_method.
```

gflownet/gflownet.py
```diff
@@ -202,66 +194,127 @@ def parameters(self):
...
 
         Args
         ----
         envs : list of GFlowNetEnv or derived
             A list of instances of the environment
 
-        times : dict
-            Dictionary to store times
+        batch_forward : Batch
+            A batch from which obtain required variables (e.g. masks) to avoid
```
michalkoziarski, line 224:
```suggestion
            A batch from which to obtain required variables (e.g. masks) to avoid
```

gflownet/gflownet.py
```diff
@@ -202,66 +194,127 @@ def parameters(self):
...
+            ), "sampling_method random and no_random True is ambiguous"
             random_action_prob = 1.0
+            temperature = 1.0
+        elif no_random is True:
+            temperature = 1.0
+            random_action_prob = 0.0
+        else:
+            if temperature is not None:
+                temperature = self.temperature_logits
+            if random_action_prob is not None:
```
michalkoziarski, line 270:
```suggestion
            if random_action_prob is None:
```
?

gflownet/gflownet.py
```diff
@@ -202,66 +194,127 @@ def parameters(self):
...
+            assert (
+                no_random is False
+            ), "sampling_method random and no_random True is ambiguous"
             random_action_prob = 1.0
+            temperature = 1.0
+        elif no_random is True:
+            temperature = 1.0
+            random_action_prob = 0.0
+        else:
+            if temperature is not None:
```
michalkoziarski, line 268:
```suggestion
            if temperature is None:
```
Shouldn't this be `if temperature is None`?

gflownet/gflownet.py
```diff
@@ -202,66 +194,127 @@ def parameters(self):
...
-        if mask_invalid_actions is None:
-            # Invalid actions masks are not provided, they need to be computed
-            if is_forward:
-                mask_invalid_actions = tbool(
-                    [env.get_mask_invalid_actions_forward() for env in envs],
-                    device=self.device,
-                )
+
+        # Retrieve masks from the batch (batch.get_item("mask_*") computes the mask if
+        # it is not available and stores it to the batch)
```
michalkoziarski, line 286:
```suggestion
        # it is not available and stores it in the batch)
```

gflownet/gflownet.py
```diff
@@ -311,178 +369,145 @@ def step(
...
-                    env.n_actions -= 1
-                    valids.append(True)
-                else:
-                    valids.append(False)
+            _, actions, valids = zip(
+                *[env.step(action) for env, action in zip(envs, actions)]
+            )
         return envs, actions, valids
 
     # @profile
```
michalkoziarski, line 388: Can this be removed?

gflownet/utils/batch.py
```diff
@@ -22,66 +23,137 @@ class Batch:
...
-
-    float_type: torch.dtype or int
-        One of float torch.dtype or an int indicating the float precision (16, 32 or
-        64).
-
     Important note: one env should correspond to only one trajectory, all env_id should
     be unique.
+
+    Note: self.state_indices start by index 1 to indicate that index 0 would correspond
+    to the source state but the latter is not stored in the batch for each trajectory.
```
michalkoziarski, line 30:
```suggestion
    Note: self.state_indices start from index 1 to indicate that index 0 would correspond
    to the source state, but the latter is not stored in the batch for each trajectory.
```

gflownet/utils/batch.py
```diff
@@ -22,66 +23,137 @@ class Batch:
...
+        if action_idx > len(self.trajectories[traj_idx]):
+            return None
+        return self.trajectories[traj_idx][action_idx - 1]
+
+    def idx2state_idx(self, idx: int):
+        return self.trajectories[self.traj_indices[idx]].index(idx)
+
+    def set_env(self, env: GFlowNetEnv):
+        """
+        Sets the generic environment passed as an argument an initializes the
```
michalkoziarski, line 132:
```suggestion
        Sets the generic environment passed as an argument and initializes the
```

gflownet/utils/batch.py
```diff
@@ -94,131 +166,164 @@ def add_to_batch(
         valids : list
             A list of boolean values indicated whether the actions were valid.
 
-        masks_invalid_actions_forward : list
-            A list of masks indicating, among all the actions that could have been
-            selected for all environments, which ones were invalid. Optional, will be
-            computed if not provided.
+        backward : bool
+            A boolean value indicating whether the action was sampled backward (False
+            by dfefault). If True, the behavior is slightly different so as to match
```
michalkoziarski, line 171:
```suggestion
            by default). If True, the behavior is slightly different so as to match
```

gflownet/utils/batch.py
```diff
@@ -228,226 +333,870 @@ def states2policy(
...
+        traj_idx: int,
+        states: Optional[
+            Union[TensorType["n_states", "..."], npt.NDArray[np.float32], List]
+        ] = None,
+        traj_indices: Optional[Union[List, TensorType["n_states"]]] = None,
+    ) -> Union[
+        TensorType["n_states", "state_proxy_dims"], npt.NDArray[np.float32], List
+    ]:
+        """
+        TODO: docstring
```
michalkoziarski, line 930: Blasphemy!

michalkoziarski (commented):
Looks (mostly) good to me, though full disclosure, due to the size I didn't get into it into very much detail.

gflownet/utils/batch.py
```diff
@@ -94,131 +166,164 @@ def add_to_batch(
...
+                continue
             if not valid:
                 continue
-            if train:
-                self.states.append(deepcopy(env.state))
-                self.actions.append(action)
-                self.env_ids.append(env.id)
+            # Add env to dictionary
+            if env.id not in self.envs:
+                self.envs.update({env.id: env})
```
josephdviviano, line 197: If the number of envs is potentially large (say in the 100+ range) I'd suggest using tuples here, which are much more memory efficient.

https://www.geeksforgeeks.org/python-memory-consumption-dictionary-vs-list-of-tuples/

josephdviviano: This comment goes for any place you index a dict with `env.id`

gflownet/utils/batch.py
josephdviviano, line 1: I will admit I only skimmed this as it's extremely long. Nothing looked objectionable but my gut tells me this could be simpler (to be explored in future work).

gflownet/utils/buffer.py
```diff
@@ -36,6 +36,11 @@ def __init__(
         self.replay.reward = pd.to_numeric(self.replay.reward)
         self.replay.energy = pd.to_numeric(self.replay.energy)
         self.replay.reward = [-1 for _ in range(self.replay_capacity)]
+        self.replay_states = {}
+        self.replay_trajs = {}
+        self.replay_pkl = "replay.pkl"
+        with open(self.replay_pkl, "wb") as f:
```
josephdviviano, line 42: Similar question -- why is the buffer being written to disk exactly?

tests/gflownet/utils/test_batch.py
josephdviviano, line 1: I skipped this. But I'm happy you have so many tests :)

josephdviviano (commented):
Left some comments, and thoughts around - in general I thought this was a really nice PR. Your code is very easy to read and clear in function.

config/gflownet/gflownet.yaml
alexhernandezgarcia: I explain that a bit in the PR description but I should add some readme info (where though?). In short:

- forward: number of forward trajectories to include in the training batch. These are on-policy trajectories possibly with random actions (if `random_action_prob > 0`) or with a tempered policy if `temperature < 1.0`
- train: number of backward trajectories to include in the training batch, sampled (backwards) from data points in a "training set"
- replay: number of backward trajectories to include in the training batch, sampled (backwards) from data points in the replay buffer.

The total number of trajectories in the training batch is the sum of the above.

gflownet/envs/base.py
alexhernandezgarcia: I have two reasons: 

- to avoid confusion with truthy or falsy values.
- to make it explicit what is being checked.

Is that bad or not pythonic?

gflownet/envs/base.py
alexhernandezgarcia: Yes, this is in the to-do list indeed.

gflownet/gflownet.py
alexhernandezgarcia: Not sure I understand. Do you mean whether `valids.append(False)` can be removed? (Yes, it can because `valids` is filled now by the `zip` loop. Or are you suggesting to remove this whole thing?

gflownet/utils/batch.py
alexhernandezgarcia: oops haha I wrote so much docs here to be fair!!!

gflownet/gflownet.py
```diff
@@ -191,7 +183,7 @@ def __init__(
         self.jsd = -1.0
 
     def parameters(self):
-        if self.backward_policy.is_model == False:
+        if self.backward_policy.is_model is False:
```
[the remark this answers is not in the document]
alexhernandezgarcia: Agree with `is` instead of `==` but what is not pythonic about `foo is False` vs `not foo`? `not foo` will be True for things other than `False`, so isn't [explicit better than implicit](https://peps.python.org/pep-0020/)? ^^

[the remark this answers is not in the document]
josephdviviano: Hm interesting I'm not longer sure. Since you named the variable `is_model`, which is explicitly boolean, I would personally default to `if is_model` and `if not is_model`, rather than `if is_model is True` and `if is_model is False`, but I take your point.

config/gflownet/gflownet.yaml
michalkoziarski: I see, how about renaming `train` to something like `backward_dataset`, and `replay` to something like `backward_replay`?

And yeah, I meant some readme/docs - I know that at this point there are more urgent things to do, so how about just creating an issue `Create training README`, and slowly start dumping ideas for what needs to be included there (I think they come up periodically during code review), and eventually at some point actually using this to create some doc?

michalkoziarski: Also, maybe just add some small comments (essentially paste what you wrote here) in the `config/gflownet/gflownet.yaml` itself?

gflownet/envs/base.py
michalkoziarski: I don't think it's a problem, it's a little non-pythonic I'd say (and to me it would normally indicate that there is a real possibility that values can be something other than boolean, and that has to be treated separately). I was wondering if that's actually one of these cases.

gflownet/gflownet.py
michalkoziarski: Oh, sorry, I was referring specifically to `# @profile` below - not what you've changed in this PR (I am not sure if that's an artifact someone forgot to remove earlier or not).

gflownet/gflownet.py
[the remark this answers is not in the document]
michalkoziarski: I commented on this [above](https://github.com/alexhernandezgarcia/gflownet/pull/150#discussion_r1277524340), at least that's my take.

config/gflownet/gflownet.yaml
alexhernandezgarcia: Sounds good! Actually I wanted to start writing docs as I coded but I have never done something like that in a big project, so I am not sure what the best way is. If you know about all this, I'd be very happy to discuss.

gflownet/envs/base.py
alexhernandezgarcia: Ok, happy to change to what is more pythonic or more common, but there are many many of such cases, so maybe let's change later as this is not urgent.

config/gflownet/gflownet.yaml
alexhernandezgarcia: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/150/commits/897ffe32ca84a131f5ffe5b69c72f9dcefc32f4e).

gflownet/envs/base.py
josephdviviano: I think this is the realm of subjective opinion. I would also go for `is backward` and enforce type checking on backward somewhere else to ensure it is a boolean always, but you could make an argument for being explicit everywhere as you have done.

gflownet/utils/batch.py
alexhernandezgarcia: [There you go](https://github.com/alexhernandezgarcia/gflownet/pull/150/commits/ac8a8921e8b46d6bde7151634c40f42806a454a2)!

gflownet/gflownet.py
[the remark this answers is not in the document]
alexhernandezgarcia: Yes, you are right about the name of this variable. There are many other cases like this, so I will make these changes another time since it is not urgent. Thanks for comment!

gflownet/gflownet.py
```diff
@@ -296,12 +352,14 @@ def sample_actions(
 
     def step(
         self,
-        envs: List,
+        envs: List[GFlowNetEnv],
```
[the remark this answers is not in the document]
alexhernandezgarcia: You are right, this is how it currently works. The environments should not be too heavy, but it is true that this may be improved. Actually, having this batch class should make this change easier. But it would be a big one anyway, so not now :sweat_smile:.

gflownet/gflownet.py
```diff
@@ -311,178 +369,145 @@ def step(
...
-        if train:
-            n_empirical = int(self.pct_offline * len(envs))
-        else:
-            n_empirical = 0
-        if n_empirical > 0 and self.buffer.train_pkl is not None:
+        batch = Batch(env=self.env, device=self.device, float_type=self.float)
+
+        # ON-POLICY FORWARD trajectories
+        t0_forward = time.time()
+        envs = [self.env.copy().reset(idx) for idx in range(n_forward)]
```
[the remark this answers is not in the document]
alexhernandezgarcia: That is what `self.env.copy()` does (see [here](vhttps://github.com/alexhernandezgarcia/gflownet/blob/batch-v2/gflownet/envs/base.py#L743)). The reason to have it this way is that at some point for something I cannot remember exactly, deepcopy wasn't enough so we needed a method to take care of it.

gflownet/gflownet.py
```diff
@@ -311,178 +369,145 @@ def step(
...
-        # Policy trajectories
-        masks_invalid_actions_forward = None
+            envs = [env for env in envs if not env.equal(env.state, env.source)]
+        times["train_actions"] = time.time() - t0_train
+
+        # REPLAY BACKWARD trajectories
+        t0_replay = time.time()
+        batch_replay = Batch(env=self.env, device=self.device, float_type=self.float)
+        if n_replay > 0 and self.buffer.replay_pkl is not None:
+            with open(self.buffer.replay_pkl, "rb") as f:
```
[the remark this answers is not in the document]
alexhernandezgarcia: This is for a live replay buffer, not a data set. Your comment about "faster" is about opening the pickled file, specifically?

gflownet/gflownet.py
```diff
@@ -311,178 +369,145 @@ def step(
...
+                dict_replay = pickle.load(f)
+                n_replay = min(n_replay, len(dict_replay["x"]))
+                envs = [self.env.copy().reset(idx) for idx in range(n_replay)]
+                # TODO: implement other sampling options besides permutation
+                if n_replay > 0:
+                    x_replay = [x for x in dict_replay["x"].values()]
+                    x_replay = [x_replay[idx] for idx in self.rng.permutation(n_replay)]
```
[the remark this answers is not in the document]
alexhernandezgarcia: I agree. This will be updated soon, so I will leave it for now.

gflownet/gflownet.py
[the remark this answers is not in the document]
josephdviviano: interesting that this isn't an equivalent operation! there must be a gap in my knowledge. anyway glad it's deep

gflownet/gflownet.py
```diff
@@ -311,178 +369,145 @@ def step(
...
+            # Update environments with sampled actions
+            envs, actions, valids = self.step(envs, actions, backward=True)
+            # Add to batch
+            batch_replay.add_to_batch(envs, actions, valids, backward=True, train=train)
+            assert all(valids)
+            # Filter out finished trajectories
+            envs = [env for env in envs if not env.equal(env.state, env.source)]
+        times["replay_actions"] = time.time() - t0_replay
+
+        # Merge forward and backward batches
```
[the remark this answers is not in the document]
alexhernandezgarcia: You mean "# Merge forward and backward batches"? What's  wrong?

gflownet/utils/batch.py
alexhernandezgarcia: uhm ok good to know, thanks! Adding an issue.

gflownet/utils/batch.py
alexhernandezgarcia: You mean the Batch class could be simpler?

gflownet/utils/buffer.py
alexhernandezgarcia: Because it's easy :P In the future the buffer will be of type Batch and won't be in disk.

gflownet/utils/buffer.py
```diff
@@ -143,22 +148,29 @@ def _add_greater(
...
-            if not self.replay["state"].isin([readable_state]).any():
-                self.replay.iloc[self.replay.reward.argmin()] = {
-                    "state": self.env.state2readable(states[idx_new_max]),
-                    "traj": self.env.traj2readable(trajs[idx_new_max]),
-                    "reward": rewards[idx_new_max],
-                    "energy": energies[idx_new_max],
+        for idx, (state, traj, reward, energy) in enumerate(
+            zip(states, trajs, rewards, energies)
+        ):
+            if allow_duplicate_states is False and state in self.replay_states.values():
```
michalkoziarski, line 156: This doesn't seem to be working for tensor states (in particular for trees): it results in `RuntimeError: Boolean value of Tensor with more than one value is ambiguous`.

alexhernandezgarcia: Yes, I am aware. It is mentioned in the description, in "Other notes":
> Backward sampling does not work yet with the tetris environment because the states are tensors and I need to adapt the code.

I am sorry but all these changes took me too long and I wanted to give priority to merging sooner than later. It's not that this PR breaks it. It didn't work before either any way. And I am planning a big refactor of the buffer class, so I thought I'll fix that later.

alexhernandezgarcia:
> Looks (mostly) good to me, though full disclosure, due to the size I didn't get into it into very much detail.

@michalkoziarski if there is nothing too bad, could you approve the PR or even merge it yourself?

michalkoziarski:
> > Looks (mostly) good to me, though full disclosure, due to the size I didn't get into it into very much detail.
> 
> @michalkoziarski if there is nothing too bad, could you approve the PR or even merge it yourself?

I think you missed a couple of comments that are in "hidden conversation" mode (I believe hidden due to the number of them), both from me and @josephdviviano; some of mine are regarding a possible bug.

gflownet/gflownet.py
alexhernandezgarcia: You are right, of course. Have tested these options yet :roll_eyes:

josephdviviano:
to be clear I don't think any of my comments should block a merge (I just didn't feel like I had the authority to approve :) )

gflownet/gflownet.py
```diff
@@ -311,178 +369,145 @@ def step(
...
+        times["train_actions"] = time.time() - t0_train
+
+        # REPLAY BACKWARD trajectories
+        t0_replay = time.time()
+        batch_replay = Batch(env=self.env, device=self.device, float_type=self.float)
+        if n_replay > 0 and self.buffer.replay_pkl is not None:
+            with open(self.buffer.replay_pkl, "rb") as f:
+                dict_replay = pickle.load(f)
+                n_replay = min(n_replay, len(dict_replay["x"]))
+                envs = [self.env.copy().reset(idx) for idx in range(n_replay)]
```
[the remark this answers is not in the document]
alexhernandezgarcia: I agree it's not ideal but that would only happen if the batch size of thousands trajectories too, which would imply a much larger effective batch size if the trajectories are long (easily dozens of steps). In that case, the overhead in memory from copying the envs would be nearly negligible compared to the memory needs to store all states, masks, actions, etc. In sum, I don't think this is a big problem. That said, it is in my to-do list to remove the need to copy the environments. Nonetheless, not in this PR :P. Thanks for the comments though, of course!

gflownet/gflownet.py
```diff
@@ -311,178 +369,145 @@ def step(
...
+                times=times,
             )
+            times["actions_envs"] += time.time() - t0_a_envs
+            # Update environments with sampled actions
+            envs, actions, valids = self.step(envs, actions, backward=True)
+            # Add to batch
+            batch_replay.add_to_batch(envs, actions, valids, backward=True, train=train)
+            assert all(valids)
+            # Filter out finished trajectories
+            envs = [env for env in envs if not env.equal(env.state, env.source)]
```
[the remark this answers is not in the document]
alexhernandezgarcia: This is to sample backward trajectories. The end of a backward trajectory occurs when the state is equal the source state (`self.source`). The end of a forward trajectory occurs when `env.done is True` (or, more pythonically, when `env.done` :stuck_out_tongue_closed_eyes:).

alexhernandezgarcia:
Oops too bad I missed those comments. I think I have addressed everything now.

Is there a quick way to see all unresolved comments, by the way?

michalkoziarski:
> Oops too bad I missed those comments. I think I have addressed everything now.
> 
> Is there a quick way to see all unresolved comments, by the way?

image

michalkoziarski (approved):

PR #151, opened 2023-07-20
RF-GFN - completion of environment and tests

michalkoziarski:
Completion of remaining components of the Tree environment. Most importantly, I have implemented the methods `sampled_actions()` and `get_logprobs()` but I have made other changes along the way to facilitate the implementation. I have also implemented new tests and re-enabled all the common tests, making sure that these are passed. GFlowNet runs without errors when trained with the uniform proxy, using the following command:

```
python main.py user=$USER env=tree proxy=uniform gflownet=trajectorybalance logger.do.online=False env.buffer.replay_capacity=0
```


alexhernandezgarcia (approved):
Looks good to me :roll_eyes:

gflownet/envs/tree.py
```diff
@@ -0,0 +1,1079 @@
...
+                for output in [0, 1]:
+                    parent = state.clone()
+                    attributes = parent[k]
+
+                    parent = self._set_stage(Stage.COMPLETE, parent)
+                    attributes[Attribute.TYPE] = NodeType.CLASSIFIER
+                    attributes[Attribute.FEATURE] = -1
+                    attributes[Attribute.THRESHOLD] = -1
+                    attributes[Attribute.CLASS] = output
+                    attributes[Attribute.ACTIVE] = Status.INACTIVE
```
michalkoziarski, line 941: Similar to the previous comment, I think this could use `_insert_classifier`.

gflownet/envs/tree.py
```diff
@@ -0,0 +1,1079 @@
...
+
+    def _to_pyg(self) -> pyg.data.Data:
+        """
+        Convert self.state into a PyG graph.
+        """
+        return from_networkx(self._get_graph())
+
+    @staticmethod
+    def _load_dataset(data_path):
+        from pathlib import Path
```
michalkoziarski, line 1014: Why not have this import at the top of the file?

gflownet/envs/tree.py
```diff
@@ -0,0 +1,1079 @@
...
+
+        data_path = Path(data_path)
+        if data_path.suffix == ".csv":
+            import pandas as pd
+
+            df = pd.read_csv(data_path)
+            X = df.iloc[:, 0:-1].values
+            y = df.iloc[:, -1].values
+        elif data_path.suffix == ".pkl":
+            import pickle
```
michalkoziarski, line 1024: Why not have this import at the top of the file?

gflownet/envs/tree.py
```diff
@@ -0,0 +1,1079 @@
...
+        """
+        return from_networkx(self._get_graph())
+
+    @staticmethod
+    def _load_dataset(data_path):
+        from pathlib import Path
+
+        data_path = Path(data_path)
+        if data_path.suffix == ".csv":
+            import pandas as pd
```
michalkoziarski, line 1018: Why not have this import at the top of the file?

gflownet/envs/tree.py
```diff
@@ -0,0 +1,1079 @@
...
+        mask_invalid_actions: TensorType["n_states", "action_space_dim"] = None,
+        temperature_logits: float = 1.0,
+    ) -> Tuple[List[Tuple], TensorType["n_states"]]:
+        """
+        Samples a batch of actions from a batch of policy outputs.
+        """
+        n_states = policy_outputs.shape[0]
+        logprobs = torch.zeros(n_states, device=self.device, dtype=self.float)
+        # Discrete actions
+        mask_discrete = mask_invalid_actions[:, self._action_index_pick_threshold]
```
michalkoziarski, line 519: I might be missing something, but aren't the actions associated with picking threshold continuous, not discrete? So shouldn't the mask associated with discrete actions be *everything but* `[:, self._action_index_pick_threshold]`?

gflownet/envs/tree.py
```diff
@@ -0,0 +1,1079 @@
...
+        Computes log probabilities of actions given policy outputs and actions.
+        """
+        n_states = policy_outputs.shape[0]
+        if states_target is None:
+            states_target = torch.empty(
+                (n_states, self.policy_input_dim), device=self.device
+            )
+        logprobs = torch.zeros(n_states, device=self.device, dtype=self.float)
+        # Discrete actions
+        mask_discrete = mask_invalid_actions[:, self._action_index_pick_threshold]
```
michalkoziarski, line 586: Same remark as above.

michalkoziarski:
Most of it looks good, thank you for that! I left some minor suggestions & one more important question regarding masking (that I likely just misunderstood).

gflownet/envs/base.py
```diff
@@ -113,6 +136,46 @@ def actions2indices(
...
+        state : list or tensor or None
+            None, or a state in GFlowNet format.
+
+        Returns
+        -------
+        state : list or tensor
+            The argument state, or self.state if state is None.
+        """
+        if state is None:
+            state = copy(self.state)
```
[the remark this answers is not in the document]
alexhernandezgarcia: It's a good question. I am sure it is not necessary to copy in all cases but I am also sure that it is in some cases and also that not copying is the source of a lot of silent bugs I have seen. Copying is safer and easier so I'd go with it for now, but I'm open to discuss an alternative, even though it may involve careful review of many parts of the code.

gflownet/envs/base.py
```diff
@@ -145,13 +208,31 @@ def get_mask_invalid_actions_backward(
...
 
+    def get_valid_actions(
+        self,
+        state: Optional[List] = None,
+        done: Optional[bool] = None,
+        backward: Optional[bool] = False,
+    ) -> List[Tuple]:
+        """
+        Returns the list of non-invalid (valid, for short) according to the mask of
+        invalid actions.
```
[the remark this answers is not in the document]
alexhernandezgarcia: It's already in the README of the environments. Do you think it is enough to point there?

gflownet/envs/tree.py
alexhernandezgarcia: Just because it's only needed there, but it's a lightweight import so I will move it up.

gflownet/envs/tree.py
alexhernandezgarcia: Just because it's only needed there, but it's a lightweight import so I will move it up.

gflownet/envs/tree.py
alexhernandezgarcia: Just because it's only needed there, but it's a lightweight import so I will move it up.

gflownet/envs/tree.py
alexhernandezgarcia: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/151/commits/896196df7007dc7ecea651b4ca2f3558efb46d55.

gflownet/envs/tree.py
alexhernandezgarcia: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/151/commits/896196df7007dc7ecea651b4ca2f3558efb46d55.

gflownet/envs/tree.py
alexhernandezgarcia: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/151/commits/896196df7007dc7ecea651b4ca2f3558efb46d55.

gflownet/envs/tree.py
alexhernandezgarcia:
```
mask_discrete = mask_invalid_actions[:, self._action_index_pick_threshold]
```
What this line does is creating a boolean mask of the elements (actions) in the batch that are discrete. It is done by taking the value (True or False) in the mask of invalid actions corresponding to the representative of the action "pick threshold", indicated by `self._action_index_pick_threshold`. If it is True, then the action "pick threshold" is invalid and therefore the action must be discrete and it will become `True` in `mask_discrete`. Does that make sense?

gflownet/envs/tree.py
alexhernandezgarcia: Just copying reply from above:

```
mask_discrete = mask_invalid_actions[:, self._action_index_pick_threshold]
```
What this line does is creating a boolean mask of the elements (actions) in the batch that are discrete. It is done by taking the value (True or False) in the mask of invalid actions corresponding to the representative of the action "pick threshold", indicated by `self._action_index_pick_threshold`. If it is True, then the action "pick threshold" is invalid and therefore the action must be discrete and it will become `True` in `mask_discrete`. Does that make sense?

gflownet/envs/base.py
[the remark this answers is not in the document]
michalkoziarski: Oh, I wasn't aware of that - I'd personally add a line referencing README for a more detailed explanations, but it's not a big deal.

gflownet/envs/tree.py
michalkoziarski: Oh, I see - makes sense!

michalkoziarski:
Looks good to me (apparently I can't approve a PR I created myself)!

gflownet/envs/base.py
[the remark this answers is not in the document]
alexhernandezgarcia: The idea is to, in an upcoming update, use `_get_state()` in all the environments, instead of `if state is None...`. Perhaps that will be the chance to rethink whether we should copy or not or optionally.

gflownet/envs/base.py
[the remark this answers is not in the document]
alexhernandezgarcia: Done [here](https://github.com/alexhernandezgarcia/gflownet/commit/d80f8b756ef2e3d7807ce30006124b0ff0047a29).

PR #152, opened 2023-07-20
Boring PR to make Tree states 2D tensors

alexhernandezgarcia:
This PR does what the title says plus other small things. Altogether:
- Make states a 2D tensor instead of 1D. Specifically:
  - Shape is n_nodes + 1 x n_attributes=5
  - k-th row has the attributes of k-th node
  - 0th element of the last row (`state[-1, 0])` contains the state's stage
- I have created getter and setter methods for the stage.
- I have created a class `Attribute` with constants to identify the index of each attribute. This helps me code and understand the code. I hope it's ok!
- I have made some methods receive the state as argument optionally (as many other methods in environments). Hope that's ok too!
- I have removed the methods that are not used any more (e.g. `_get_attributes()`)
  - Also, `N_ATTRIBUTES` -> `Attribute.N`
- I have updated the tests, which are easier to write and read with the 2D states.

### Tests
- [x] `python -m pytest tests/gflownet/envs/test_tree.py` :heavy_check_mark:

### DIFF
--- a/gflownet/envs/tree.py
+++ b/gflownet/envs/tree.py
@@ -195,12 +195,13 @@ class Tree(GFlowNetEnv):
         # source is initialized with a classifier.
         self.n_nodes = 2**max_depth - 1
         self.source = torch.full((self.n_nodes + 1, Attribute.N), torch.nan)
-        self.source[-1, 0] = Stage.COMPLETE
-        attributes_root = self.source[0, :]
-        attributes_root[0] = NodeType.CLASSIFIER
-        attributes_root[1:3] = -1
-        attributes_root[3] = Counter(self.y).most_common()[0][0]
-        attributes_root[4] = Status.INACTIVE
+        self.source._set_stage(Stage.COMPLETE, self.source)
+        attributes_root = self.source[0]
+        attributes_root[Attribute.TYPE] = NodeType.CLASSIFIER
+        attributes_root[Attribute.FEATURE] = -1
+        attributes_root[Attribute.THRESHOLD] = -1
+        attributes_root[Attribute.CLASS] = Counter(self.y).most_common()[0][0]
+        attributes_root[Attribute.ACTIVE] = Status.INACTIVE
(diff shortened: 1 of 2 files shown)


gflownet/envs/tree.py
```diff
@@ -104,14 +122,18 @@ class Tree(GFlowNetEnv):
...
     the maximum depth is required), with nodes indexed from k = 0 to 2**max_depth - 2,
-    and each node containing a 5-element attribute tensor (see _get_attributes for
+    and each node containing a 5-element attribute tensor (see class Attribute for
     details). The nodes are indexed from top left to bottom right, as follows:
 
                 0
         1               2
     3       4       5       6
+
+    States are represented by a tensor with shape [n_nodes + 1, 5], where each row k-th
```
michalkoziarski, line 134:
```suggestion
    States are represented by a tensor with shape [n_nodes + 1, 5], where each k-th row
```

gflownet/envs/tree.py
```diff
@@ -89,8 +89,26 @@ class ActionType:
     PICK_OPERATOR = 3
 
 
-# Number of attributes encoding each node; see Tree._get_attributes.
-N_ATTRIBUTES = 5
+class Attribute:
+    """
```
michalkoziarski, line 93:
```suggestion
    """
    Contains indices of individual attributes in a state tensor.
    
```

gflownet/envs/tree.py
```diff
@@ -177,17 +199,18 @@ def __init__(
         self.beta_params_min = beta_params_min
         self.beta_params_max = beta_params_max
         # Source will contain information about the current stage (on the 0-th position),
```
michalkoziarski, line 201:
```suggestion
        # Source will contain information about the current stage (on the last position),
```

michalkoziarski (approved):
Looks good to me! I added some minor suggestions for docstrings.

PR #153, opened 2023-07-24
Suggested improvements to composition-sg-constraints PR

carriepl:
Here are the changes to #148 that I am suggesting in this PR : 
- If the current state is deemed incompatible with the space group, then EOS is not marked as invalid to allow the GFN to end the trajectory ASAP since no actions it could take could bring it back into a valid state.
- Adding some code to mask actions that would lead to states incompatible with the space group
- Implement some utilitary fonctions to cache results from PyXtal's function because a lot of those are very very expensive to compute.

**Runtime**
I evaluated the same it takes to complete 10 trajectories, for the composition environment only and with its default hyperparameters, for every space group (2300 trajectories total). Since the cache mechanism means that the timings will change over time as the caches get larger, I repeated the process 40 times and produced here the total time, in seconds, for each attempt. 

1 - 6.4116950035095215
2 - 4.453245639801025
3 - 4.2290565967559814
4 - 4.128775358200073
5 - 4.059585094451904
6 - 3.9281723499298096
7 - 3.892940044403076
8 - 3.927881956100464
9 - 3.8759658336639404
10 - 3.8844809532165527
11 - 3.873915195465088
12 - 3.8338799476623535
13 - 3.86641788482666
14 - 3.880934953689575
15 - 3.833317279815674
16 - 3.8493034839630127
17 - 3.8233022689819336
18 - 3.824620246887207
19 - 3.823237895965576
20 - 3.7951202392578125
21 - 3.80896258354187
22 - 3.791884422302246
23 - 3.813082695007324
24 - 3.8194580078125
25 - 3.8226542472839355
26 - 3.8004648685455322
27 - 3.7659032344818115
28 - 3.81996488571167
29 - 3.812319040298462
30 - 3.824629068374634
31 - 3.821237802505493
32 - 3.8417959213256836
33 - 3.8055989742279053
34 - 3.800288677215576
35 - 3.8210251331329346
36 - 3.8367960453033447
37 - 3.8054256439208984
38 - 3.8340342044830322
39 - 3.808349609375
40 - 3.803480863571167

For reference, on the same compute node, running the same number of trajectories with the space group checks deactivated took 3.23 seconds. That's about 18% more runtime with the checks.

**Valid vs invalid samples**

On that 40th attempt, I looked at the number of valid samples that were generated for each space group to see if the increase was worth the additional runtime. The results are summarized in the bar graph below : 

- The figure shows results with the space group checks and without
- The vertical bar show how many of the 10 samples generated for each space group were considered valid by PyXtal's check_compatible(). The results without the space group checks are in blue. The results with the checks are in red. Since there is some alpha in this figure, the locations in purple show where the two methods overlap.
- The black dots indicate, for each individual space group, the ratio of time it takes to complete 1 trajectory with the checks vs 1 trajectory without the checks.

image

Out of 2300 possible valid samples, generating samples without the space group checks give us ~386 valid samples from very few space groups. With the space group checks, it increases to 2270 (5.88x increase). There are three space groups (220, 228, 230) for which neither method produces valid samples. This is because the default hyperparameters of the composition environment are incompatible with these space groups.

Note : the samples are classified as valid/invalid using PyXtal's check_compatible() method. As discussed in the Slack channel, that method is not 100% reliable but it generally gives the right answer.

### DIFF
--- a/gflownet/envs/crystals/composition.py
+++ b/gflownet/envs/crystals/composition.py
@@ -13,10 +13,8 @@ from torchtyping import TensorType
 from gflownet.envs.base import GFlowNetEnv
 from gflownet.utils.crystals.constants import ELEMENT_NAMES, OXIDATION_STATES
 from gflownet.utils.crystals.pyxtal_cache import (
-    get_space_group,
-    space_group_check_compatible,
-    space_group_lowest_free_wp_multiplicity,
-)
+    get_space_group, space_group_check_compatible,
+    space_group_lowest_free_wp_multiplicity, space_group_wyckoff_gcd)
 
 
 class Composition(GFlowNetEnv):
@@ -178,7 +176,9 @@ class Composition(GFlowNetEnv):
             # added.
             most_specific_wp = space_group.get_wyckoff_position(-1)
             min_atom_i = most_specific_wp.multiplicity
-            max_atom_i = (self.max_atom_i // min_atom_i) * min_atom_i
+
+            wyckoff_gcd = space_group_wyckoff_gcd(self.space_group)
+            max_atom_i = (self.max_atom_i // wyckoff_gcd) * wyckoff_gcd
 
             # Determine if the current composition is compatible with the
             # space group
@@ -293,8 +293,10 @@ class Composition(GFlowNetEnv):
             space_group = get_space_group(self.space_group)
             n_atoms = [s for s in state if s > 0]
 
-            # Get the multiplicity of the group's most specific fixed wyckoff position
-            fixed_multiplicity = space_group.get_wyckoff_position(-1).multiplicity
+            # Get the greated common divisor of the group's wyckoff position.
+            # It cannot be valid to add a number of atoms that is not a
+            # multiple of this value
+            wyckoff_gcd = space_group_wyckoff_gcd(self.space_group)
 
             # Get the multiplicity of the group's most specific wyckoff position with
             # at least one degree of freedom
@@ -312,9 +314,9 @@ class Composition(GFlowNetEnv):
                     or not mask_unrequired_element[action_idx]
                 ):
                     # If the number of atoms added by this action is not a
-                    # multiple of the most specific wyckoff position, mark
-                    # action as invalid
-                    if nb_atoms_action % fixed_multiplicity != 0:
+                    # multiple of the greatest common divisor of the wyckoff
+                    # positions' multiplicities, mark action as invalid
+                    if nb_atoms_action % wyckoff_gcd != 0:
                         mask_required_element[action_idx] = True
                         mask_unrequired_element[action_idx] = True
                         continue
--- a/gflownet/envs/crystals/composition.py
+++ b/gflownet/envs/crystals/composition.py
@@ -12,8 +12,11 @@ from torchtyping import TensorType
 
 from gflownet.envs.base import GFlowNetEnv
 from gflownet.utils.crystals.constants import ELEMENT_NAMES, OXIDATION_STATES
-from gflownet.utils.crystals.pyxtal_cache import (get_space_group,
-                                                  space_group_check_compatible)
+from gflownet.utils.crystals.pyxtal_cache import (
+    get_space_group,
+    space_group_check_compatible,
+    space_group_lowest_free_wp_multiplicity,
+)
 
 
 class Composition(GFlowNetEnv):
@@ -295,12 +298,9 @@ class Composition(GFlowNetEnv):
 
             # Get the multiplicity of the group's most specific wyckoff position with
             # at least one degree of freedom
-            free_multiplicity = None
-            for wyckoff_idx in range(1, len(space_group.wyckoffs) + 1):
-                wyckoff_position = space_group.get_wyckoff_position(-wyckoff_idx)
-                if wyckoff_position.get_dof() > 0:
-                    free_multiplicity = wyckoff_position.multiplicity
-                    break
+            free_multiplicity = space_group_lowest_free_wp_multiplicity(
+                self.space_group
+            )
 
             # Go through each action in the masks, validating them
             # individually
@@ -339,7 +339,9 @@ class Composition(GFlowNetEnv):
                     # incompatible with the space group, mark action as
                     # invalid
                     n_atoms_post_action = n_atoms + [nb_atoms_action]
-                    sg_compatible = space_group_check_compatible(self.space_group, n_atoms_post_action)
+                    sg_compatible = space_group_check_compatible(
+                        self.space_group, n_atoms_post_action
+                    )
                     if not sg_compatible:
                         mask_required_element[action_idx] = True
                         mask_unrequired_element[action_idx] = True
--- a/gflownet/envs/crystals/composition.py
+++ b/gflownet/envs/crystals/composition.py
@@ -165,32 +165,60 @@ class Composition(GFlowNetEnv):
         n_unused_required_elements = len(unused_required_elements)
         n_used_atoms = sum(state)
 
+        if self.do_spacegroup_check and isinstance(self.space_group, int):
+            space_group = Group(self.space_group)
+
+            # Determine, based on the space group's Wyckoff's positions, what
+            # is the min/max number of atoms of a given element that could be
+            # added.
+            most_specific_wp = space_group.get_wyckoff_position(-1)
+            min_atom_i = most_specific_wp.multiplicity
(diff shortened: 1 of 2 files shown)


gflownet/utils/crystals/pyxtal_cache.py
```diff
@@ -0,0 +1,122 @@
...
+    # space group's most specific free wyckoff position. This improves the
+    # cache hit rate without affecting the validity of the composition
+    free_multiplicity = space_group_lowest_free_wp_multiplicity(group_index)
+    composition = [c for c in composition if c % free_multiplicity != 0]
+
+    # Get a tuple version of composition to ensure immutability and,
+    # therefore, allow dictionary indexing by composition. Ensure the elements
+    # in the tuple are sorted to improve cache hit rate since it doesn't
+    # affect the validity of a composition
+    t_composition = tuple(sorted(composition))
```
alexhernandezgarcia, line 64: Nice! :)

gflownet/utils/crystals/pyxtal_cache.py
```diff
@@ -0,0 +1,122 @@
...
+    Returns
+    -------
+    is_compatible : bool
+        True if the composition and space group are compatible. False
+        otherwise.
+    """
+    # Remove from composition the number of atoms that are a multiple of the
+    # space group's most specific free wyckoff position. This improves the
+    # cache hit rate without affecting the validity of the composition
+    free_multiplicity = space_group_lowest_free_wp_multiplicity(group_index)
```
alexhernandezgarcia, line 57: This is pretty cool :ok_hand:

gflownet/envs/crystals/composition.py
```diff
@@ -165,47 +170,76 @@ def get_mask_invalid_actions_forward(self, state=None, done=None):
...
 
+        if self.do_spacegroup_check and isinstance(self.space_group, int):
+            space_group = get_space_group(self.space_group)
+
+            # Determine, based on the space group's Wyckoff's positions, what
+            # is the min/max number of atoms of a given element that could be
+            # added.
+            most_specific_wp = space_group.get_wyckoff_position(-1)
+            min_atom_i = most_specific_wp.multiplicity
+            max_atom_i = (self.max_atom_i // min_atom_i) * min_atom_i
```
alexhernandezgarcia, line 181: This may potentially increase the minimum number of atoms and decrease the maximum number of atoms with respect to the config values, right?

gflownet/envs/crystals/composition.py
```diff
@@ -258,6 +287,65 @@ def get_element_mask(min_atoms, max_atoms):
...
+                    # If the composition resulting from this action is
+                    # incompatible with the space group, mark action as
+                    # invalid
+                    n_atoms_post_action = n_atoms + [nb_atoms_action]
+                    sg_compatible = space_group_check_compatible(
+                        self.space_group, n_atoms_post_action
+                    )
+                    if not sg_compatible:
+                        mask_required_element[action_idx] = True
+                        mask_unrequired_element[action_idx] = True
```
alexhernandezgarcia, line 347: Nice stuff! Suggestion, but as you wish: perhaps pack this into a separate method?

alexhernandezgarcia (approved):
Very nice work! It all looks good to me. I made one suggestion but I'd be happy to merge regardless.

gflownet/envs/crystals/composition.py
carriepl: Yeah, I agree that this method is starting to be pretty long. I was wondering where I could put the method's boundries so that it is well contained but without having to provide a huge number of arguments. Any suggestion?

gflownet/envs/crystals/composition.py
carriepl: Btw, thanks for the comment. It made me revisit the code and find a bug.
I have originally written this making the assumption that, for a space group, all WPs are multiples of the most specific WP. 
I figured I would validate that assumption, and fix the code if needed, once the pipeline was working. Well, I forgot to do that before doing my PR.
Your comment made me revisit that bit of code and fix it because I checked and there are 4 space groups for which my assumption was invalid.

gflownet/envs/crystals/composition.py
```diff
@@ -258,6 +287,67 @@ def get_element_mask(min_atoms, max_atoms):
...
+            # Get the multiplicity of the group's most specific wyckoff position with
+            # at least one degree of freedom
+            free_multiplicity = space_group_lowest_free_wp_multiplicity(
+                self.space_group
+            )
+
+            # Go through each action in the masks, validating them
+            # individually
+            for action_idx, nb_atoms_action in enumerate(
+                range(self.min_atom_i, self.max_atom_i + 1)
```
alexhernandezgarcia, line 310: Is it correct that here we should use `self.min_atom_i` and `self.max_atom_i` instead of `min_atom_i` and `max_atom_i` that are updated above?

gflownet/envs/crystals/composition.py
alexhernandezgarcia: I have taken a stab at it [here](https://github.com/alexhernandezgarcia/gflownet/commit/48df853d64f8078430df2e05d17955c798772550). What do you think?

gflownet/envs/crystals/composition.py
alexhernandezgarcia: Well, glad to hear :) And thanks for the explanation!

gflownet/envs/crystals/composition.py
carriepl: No because, here, the goal is to go over all the actions in the masks (`mask_required_element `and `mask_unrequired_element`). And, currently, these masks have the same length whatever the space group.
We could change the implementation where the masks are only as long as necessary given the space group but that would involve changing the `get_element_mask()` function above, and the `for` loop below to copy those element masks into the global mask afterwards.

alexhernandezgarcia: I see, that makes sense. Let's leave it as is then. Then my changes [here](https://github.com/alexhernandezgarcia/gflownet/commit/48df853d64f8078430df2e05d17955c798772550) should be fine enough, I hope.

gflownet/envs/crystals/composition.py
carriepl: Seems absolutely fine. I don't know why I thought the result would be messier, this is pretty good.

There are only two things that I'm wondering about : 
- The method since to be both modifying the input masks inplace and returning the updated masks afterwards. Might be simpler to have only one of these mechanism. For speed reason, we might prefer the updating in place. If that's the case, we would need to update the docstring to make it obvious that the change happens inplace.
- In the docstring argument descriptions, it says `Input element-wise mask indicating whether the element is required` for one of the arguments and something similar for the other arguments/return values. That's not exactly right. Each input is an element-wise mask indicating whether each action is invalid for a type of elements. There is one mask for the elements that are required by the environment (they absolutely have to be in the composition) and one mask for the other elements which are not required but may be added if there is room for them.
I'd reword to something like 
```
Input element-wise mask indicating invalid actions for required elements.
This masks indicates whether each individual actions is invalid or not for elements that are required to be in the composition by the end of the trajectory.
```
and
```
Input element-wise mask indicating invalid actions for unrequired elements.
This masks indicates whether each individual actions is invalid or not for elements that are not required to be in the composition by the end of the trajectory.
```

michalkoziarski:
I just read the README (didn't get into the code), but the results look good to me, thank you for this! Regarding the three problematic space groups: so do I understand correctly, using this in combination with the defaults would lead to them never being able to be sampled? And if so, do you know what (reasonable from other points of view, e.g. not too large search space) defaults would work for all SGs?

gflownet/envs/crystals/composition.py
alexhernandezgarcia: Sorry for the errors in the docs and thanks for the suggestions! I have made the changes and opened a [PR](https://github.com/alexhernandezgarcia/gflownet/pull/165).

carriepl:
> using this in combination with the defaults would lead to them never being able to be sampled? 

Well, they could be sampled (at the space-group phase) but then no atom would be added at all during the composition phase because nothing valid could be added.

But it's just as bad without the checks; we could sample those space groups and then sample compositions. But we'd never sample valid compositions here either so all samples generated for these space groups would be meaningless.


> And if so, do you know what (reasonable from other points of view, e.g. not too large search space) defaults would work for all SGs?

The minimal change to the defaults which would make it possible to generate valid compositions for these space groups would be to set `max_atom_i = 16`.

PR #154, opened 2023-07-27
RF-GFN changes to evaluation

michalkoziarski:


config/experiments/tree.yaml
```diff
@@ -0,0 +1,56 @@
...
+  random_action_prob: 0.1
+  optimizer:
+    batch_size:
+      forward: 100
+      backward_dataset: 0
+      backward_replay: 0
+    lr: 0.00001
+    z_dim: 16
+    lr_z_mult: 1000
+    n_train_steps: 40000
```
alexhernandezgarcia, line 26: 40000 steps by default may be too much?

michalkoziarski: Yeah, I wasn't sure how quick convergence to expect / how appropriate lowering the number of steps would be without tuning the remaining parameters (these are taken from alanine dipeptide experiment), so I just started with a long training and figured I would lower it later. But feel free to propose any other value.

gflownet/envs/tree.py
```diff
@@ -1070,3 +1130,37 @@ def plot(self) -> None:
             node_size=800,
         )
         plt.show()
+
+    @staticmethod
+    def _compute_scores(
```
alexhernandezgarcia: Some docstring would be great.

alexhernandezgarcia (approved):
It looks good to me! It would be nice to add some missing docstring in some methods of the tree environment, but I am happy to merge asap regardless.

PR #155, opened 2023-07-27
Generic env method to sample random terminating states

alexhernandezgarcia:
A generic method in `gflownet/envs/base.py` to sample random trajectories by using the random policy of the environment, that is by using `self.trajectory_random()` which in turn uses `self.step_random()`.

I have also implemented alongside the option `random` for the buffers, which constructs a train or test buffer by calling `get_random_terminating_states()`. 

It can be used by setting `type: random` and `n: 100` (for example) in the buffers config. For instance, from the command line:

```
python main.py proxy=corners env=grid env.length=10 logger.do.online=False env.buffer.test.type=random +env.buffer.test.n=25
python main.py proxy=tetris env=tetris env.height=6 env.width=4 logger.do.online=False +env.buffer.test.type=random +env.buffer.test.n=10
``` 
Note that I had to merge `batch-v2` so as to be able to use method `equal()` from the environment, so the changes are hard to see.

### DIFF
--- a/gflownet/envs/base.py
+++ b/gflownet/envs/base.py
@@ -1,6 +1,7 @@
 """
 Base class of GFlowNet environments
 """
+import uuid
 from abc import abstractmethod
 from copy import deepcopy
 from textwrap import dedent
@@ -13,7 +14,7 @@ import torch
 from torch.distributions import Categorical
 from torchtyping import TensorType
 
-from gflownet.utils.common import set_device, set_float_precision
+from gflownet.utils.common import copy, set_device, set_float_precision, tbool, tfloat
 
 
 class GFlowNetEnv:
@@ -40,10 +41,13 @@ class GFlowNetEnv:
         fixed_distribution: Optional[dict] = None,
         random_distribution: Optional[dict] = None,
         conditional: bool = False,
+        continuous: bool = False,
         **kwargs,
     ):
         # Flag whether env is conditional
         self.conditional = conditional
+        # Flag whether env is continuous
+        self.continuous = continuous
         # Call reset() to set initial state, done, n_actions
         self.reset()
         # Device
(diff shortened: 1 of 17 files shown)


gflownet/envs/base.py
```diff
@@ -392,6 +491,56 @@ def trajectory_random(self):
...
+            self.reset()
+            state, _ = self.trajectory_random()
+            if unique is True:
+                if any([self.equal(state, s) for s in states]):
+                    add = False
+            if add is True:
+                states.append(state)
+            count += 1
+        return states
+
```
alexhernandezgarcia, line 543: @michalkoziarski this is new

gflownet/utils/buffer.py
```diff
@@ -189,6 +201,12 @@ def make_data_set(self, config):
             and hasattr(self.env, "get_uniform_terminating_states")
         ):
             samples = self.env.get_uniform_terminating_states(config.n, config.seed)
+        elif (
+            config.type == "random"
+            and "n" in config
+            and hasattr(self.env, "get_random_terminating_states")
+        ):
+            samples = self.env.get_random_terminating_states(config.n)
```
alexhernandezgarcia, line 209: @michalkoziarski this is new too

gflownet/utils/buffer.py
```diff
@@ -189,6 +201,12 @@ def make_data_set(self, config):
             and hasattr(self.env, "get_uniform_terminating_states")
         ):
             samples = self.env.get_uniform_terminating_states(config.n, config.seed)
+        elif (
+            config.type == "random"
+            and "n" in config
+            and hasattr(self.env, "get_random_terminating_states")
```
michalkoziarski, line 207: I wouldn't bother with changing this in this PR, but a comment for the evaluation refactor: I would add a better exception handling, i.e. only check if `config.type == ...`, and do the remaining checks within that if clause (and raise appropriate errors if, for example, `n` is not in config). The reason being, from a user perspective it might not always be clear why we are not getting any evaluation, while we are clearly requesting one.

gflownet/envs/base.py
```diff
@@ -392,6 +491,56 @@ def trajectory_random(self):
...
+            default, therefore if more than 100000 are requested, max_attempts should
+            be increased accordingly.
+
+        Returns
+        -------
+        states : list
+            A list of randomly sampled terminating states.
+        """
+        if unique is False:
+            max_attempts = n_states + 1
```
michalkoziarski, line 529: So this is more of a comment about `self.trajectory_random`, but just to confirm - current convention for handling states that reached an invalid, well, state, from which it is impossible to reach a valid state anymore (like in crystals in some cases) is that they would return `valid = False`, but `done = True`, right?

If yes, couldn't `trajectory_random` get stuck in such case (with a state that is `done`, but `not valid`)? If not, how is it handled exactly?

michalkoziarski (approved):
Looks good to me, left some comments not strictly related to this PR.

gflownet/utils/buffer.py
alexhernandezgarcia: you are right, adding an issue

gflownet/envs/base.py
alexhernandezgarcia: I had not thought of this, to be honest. I will investigate it later.

PR #165, opened 2023-07-28
Pack method to refine compatibility check

alexhernandezgarcia:


gflownet/envs/crystals/composition.py
```diff
@@ -290,64 +373,13 @@ def get_element_mask(min_atoms, max_atoms):
...
-
-            # Get the multiplicity of the group's most specific wyckoff position with
-            # at least one degree of freedom
-            free_multiplicity = space_group_lowest_free_wp_multiplicity(
-                self.space_group
+            (
+                mask_required_element,
+                mask_unrequired_element,
+            ) = self._refine_compatibility_check(
+                state, mask_required_element, mask_unrequired_element
```
carriepl, line 380: _refine_compatibility_check doesn't return anything anymore so I'd guess that this doesn't run in practice.

alexhernandezgarcia: Thanks for the catch! Fixed in [d8736db](https://github.com/alexhernandezgarcia/gflownet/pull/165/commits/d8736db677f78dc73e4e6c5361689008b3cbd71b).

PR #167, opened 2023-07-29
Log likelihood estimation

alexhernandezgarcia:
## Summary

Code to estimate the log likelihood of sampling data points with the current GFLowNet policy, which can be used to compute potentially insightful metrics. This PR includes the calculation of:

- The correlation between the estimated log-likelihood of test data and the rewards of the data.
- The mean negative log likelihood of the test data.

You can see one example on [wandb](https://wandb.ai/alexhg/gfn_sanity_checks/runs/o37obgos).

## Estimation of the log-likelihood

The log-likelihood is estimated by sampling backward trajectories according to the backward policy, then calculating the log probability of a sample with importance sampling, where the weights are the backward transition probabilities of the trajectories. In particular:

$\log p_T(x) = \int_{x \in \tau} P_F(\tau)d\tau$

$= \log \mathbb{E}_{P_B(\tau|x)} \frac{P_F(x)}{P_B(\tau|x)}$

$\approx \log \frac{1}{N} \sum_{i=1}^{N} \frac{P_F(x_i)}{P_B(\tau|x_i)}, x_i \sim P_B(\tau|x_i)$

## Other notes

- For convenience, I have implemented a new method `GFlowNetAgent.compute_logprobs_trajectories()`, which is used now too by the trajectory balance loss method.
- The code of `estimate_logprobs_data()`may be simplified and modularised a bit, but it's not a big deal anyway.

The core of this PR is ready to review but note that the code to calculate the metrics in `test()` has become even messier and there may be issues before/after merging. A future PR should organise the evaluation code.

## Things in this PR not directly related to the goal

- `set_state()` for the Tetris environment, to catch cases where the state and done are incompatible.
- I had to implement a couple of methods in the Batch to make the trajectory indices consecutive.


alexhernandezgarcia:
I have initially implemented the estimation of the log-likelihood such that the trajectories need to be unique, but I am now thinking that this way the samples are not iid. It should be samples purely from PB, right?

gflownet/utils/batch.py
```diff
@@ -1062,6 +1066,48 @@ def is_valid(self) -> bool:
...
         return True
 
+    def traj_indices_are_consecutive(self) -> bool:
+        """
+        Returns True if the trajectory indices start from 0 and are consecutive; False
+        otherwise.
+        """
+        return set(self.trajectories) == set(np.arange(self.get_n_trajectories()))
+
+    def make_indices_consecutive(self):
```
michalkoziarski, line 1076: This is something for which a small unit test (with multiple expected cases) would be great.

gflownet/utils/batch.py
```diff
@@ -1062,6 +1066,48 @@ def is_valid(self) -> bool:
...
+        traj_indices_unique = np.array(traj_indices_unique)[np.sort(indices)]
+        traj_indices_map_dict = OrderedDict(
+            zip(
+                traj_indices_unique,
+                np.arange(len(self.trajectories)),
+            )
+        )
+        self.trajectories = OrderedDict(
+            {
+                traj_idx_new: self.trajectories.pop(traj_idx_old)
```
michalkoziarski, line 1096: This should be minor, but why use `pop` here (and below for envs) instead of simple access? It should be marginally slower for no benefit I think?
```
>>> timeit.timeit('d={i: i for i in range(1000)}; [d.pop(i) for i in range(1000)]', number=10000)
0.9229674290399998
>>> timeit.timeit('d={i: i for i in range(1000)}; [d[i] for i in range(1000)]', number=10000)
0.7814702009782195
```

gflownet/gflownet.py
```diff
@@ -591,43 +643,143 @@ def trajectorybalance_loss(self, it, batch):
...
+            The maximum number of attempts to sample a distinct trajectory, to avoid
+            getting trapped in an infinite loop.
+
+        Returns
+        -------
+        logprobs_estimates: torch.tensor
+            The logarithm of the average ratio PF/PB over n trajectories sampled for
+            each data point.
+        """
+        max_data = 1e5
```
michalkoziarski, line 704: 1) that's pretty hacky,
2) if it is necessary, maybe better make it an argument?

michalkoziarski: (2. in huge part because that would force you to describe what it is and why it's necessary in a docstring)

gflownet/gflownet.py
```diff
@@ -911,7 +1111,7 @@ def log_iter(
         # TODO: integrate corr into test()
         if not self.logger.lightweight and self.buffer.test is not None:
             corr, data_logq, times = self.get_log_corr(times)
-            self.logger.log_sampler_test(corr, data_logq, it, self.use_context)
+            self.logger.log_sampler_test(rr, data_logq, it, self.use_context)
```
michalkoziarski, line 1114: Uh oh.

gflownet/gflownet.py
```diff
@@ -765,13 +931,31 @@ def test(self, **plot_kwargs):
...
+                {},
+            )
         with open(self.buffer.test_pkl, "rb") as f:
             dict_tt = pickle.load(f)
             x_tt = dict_tt["x"]
+
+        # Compute correlation between the rewards of the test data and the log
+        # likelihood of the data according the the GFlowNet policy; and NLL.
+        # TODO: organise code for better efficiency and readability
+        logprobs_x_tt = self.estimate_logprobs_data(x_tt, n_trajectories=10)
```
michalkoziarski, line 950: Is there some empirical data indicating that 10 trajectories are sufficient for good estimation? Also, it would be nice to have this as an argument that can be changed from config.

michalkoziarski (commented):
I reviewed it, it seems ok to me (minus some minor comments). I would love to get one more review from someone else (at least for the `estimate_logprobs_data` function, the rest should be fine), but I guess that will be unlikely to get at the moment in a reasonable time.

One question: did you confirm that the results with this are comparable to previous versions on your test environments (since technically you change the TB loss)?

gflownet/utils/batch.py
```diff
@@ -1062,6 +1066,48 @@ def is_valid(self) -> bool:
             return False
         return True
 
+    def traj_indices_are_consecutive(self) -> bool:
```
AlexandraVolokhova, line 1069: This function does not check whether indices are going in the sorted order, only that the consecutive indices are present in the `self.trajectories `dict because `set` object does not care about the order (e.g. `set([1,2,3]) == set([3,1,2])` is `True`). Is that what you expect here or the order should matter (from the docstring it seems to me like you want to check the order too)

gflownet/gflownet.py
```diff
@@ -591,43 +643,143 @@ def trajectorybalance_loss(self, it, batch):
...
         )
-        return loss, loss, loss
+        traj_indices_mat = torch.full(
+            (data_indices.max() + 1, traj_indices.max() + 1),
+            -1,
+            dtype=torch.long,
+            device=self.device,
+        )
+        traj_indices_mat[data_indices, traj_indices] = traj_indices
+        n_trajs_per_sample, _ = traj_indices_mat.max(dim=1)
```
AlexandraVolokhova, line 770: Probably could be simplified as we should have the same number of trajectories for each sample provided as an argument to this function

gflownet/gflownet.py
```diff
@@ -591,43 +643,143 @@ def trajectorybalance_loss(self, it, batch):
...
+        # Compute log probabilities of the trajectories
+        logprobs_f[data_indices, traj_indices] = self.compute_logprobs_trajectories(
```
AlexandraVolokhova, line 772: it seems to me that this line (and the next one) relies on the fact that environments were added to the batch in a specific order, i.e. [first env for the first data sample, second env for the first data sample, ...]. There's nothing wrong with it and this order seems to be satisfied in the loop above, but maybe it's worth mentioning in a comment that the order is very important (otherwise we will mess up probs for different data samples)

AlexandraVolokhova (approved):
Thank you for this great job! I left a couple of comments for simplifying / making more readable the code, but in overall, it looks good to me.

AlexandraVolokhova:
One more thing: I'd add computing and tracking two variances of the log probs: 
1. variance over samples of logprobs_estimates (to understand better the behaviour of the correlation coefficient over the training)
2. median over samples of the variances of the logprobs_estimates over trajectories for each sample (to get a sense of how noisy the estimation is). The math is a bit tricky here as we use log mean as an estimation, not just the mean. But there're some work around: https://stats.stackexchange.com/questions/418313/variance-of-x-and-variance-of-logx-how-to-relate-them 
But in any case, we will need to compute empirical var(P_F(tau) / P_B (tau)) / n_traj for each sample and then play around a bit with it to get variance for the log mean estimation.

alexhernandezgarcia:
> 

I did compare the results with previous versions, at least with the Grid, and I have later run more instances. See this [report on the sanity checks project on wandb](https://wandb.ai/alexhg/gfn_sanity_checks/reports/Logprobs-Sanity-Checks--Vmlldzo1MjgxMjQ4).

gflownet/gflownet.py
alexhernandezgarcia: I think it's important to have to avoid someone passing a massive data set of say 10 M data points for which you have to sample 10 trajectories for each point...

alexhernandezgarcia: That's why I didn't make it an argument, but I will.

gflownet/gflownet.py
alexhernandezgarcia: Yes, I will remove all this.

gflownet/gflownet.py
alexhernandezgarcia: To your question: No, that I am aware of.

Agree with the _request_, so done [here](https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/567533529dfe76a74519817df941f4cd65cf6680).

gflownet/gflownet.py
alexhernandezgarcia: The order is ensured by both [this line](https://github.com/alexhernandezgarcia/gflownet/pull/167/files#diff-621df8cda7957075a15e68cd043eec610a0e71cb65791028ad4092ac1b725bbaR538) and because the ids of the environments are set in a specific way [here](https://github.com/alexhernandezgarcia/gflownet/pull/167/files#diff-621df8cda7957075a15e68cd043eec610a0e71cb65791028ad4092ac1b725bbaR729).

alexhernandezgarcia: I have added a note in the docstring anyway [here](https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/af8043c85175ada437ea05c9dcc030372bf407c6).

gflownet/gflownet.py
alexhernandezgarcia: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/698ba2c48c439bb43190ca24f5c4e510d575ac19)

gflownet/utils/batch.py
alexhernandezgarcia: Not so small...

gflownet/utils/batch.py
alexhernandezgarcia: Oops you're right! Fixed [here](https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/01a605e693e863b7c636040710f0d020debd4633).

gflownet/utils/batch.py
alexhernandezgarcia: [There](https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/28fa35c2c789253edc43e674144c5ca5dfabfd1c) you go.

gflownet/utils/batch.py
alexhernandezgarcia: I have re-implemented the whole thing anyway, without using `.pop()`. See [here](https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/b28ded1da22ddec29db00fba8e116feecae01100).

gflownet/utils/batch.py
alexhernandezgarcia: And note that I have reimplemented things:
- https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/b28ded1da22ddec29db00fba8e116feecae01100
- https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/27a5893b631729b6aa5ef2deed3d7d2dab7d5d73

alexhernandezgarcia: And finally, note that `make_indices_consecutive()` is now not needed after all... (https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/627ee04aebc6eda09e5ae45dd9299a97c691036f)

gflownet/gflownet.py
alexhernandezgarcia: Made clearer in https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/3599dbf13a7cf0e0bead4b937a85a9818a215cd9.

gflownet/gflownet.py
alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/167/commits/abc94e9f85197157aaf450be7ca480f2d543c3cf

alexhernandezgarcia:
> One more thing: I'd add computing and tracking two variances of the log probs:
> 
>     1. variance over samples of logprobs_estimates (to understand better the behaviour of the correlation coefficient over the training)
> 
>     2. median over samples of the variances of the logprobs_estimates over trajectories for each sample (to get a sense of how noisy the estimation is). The math is a bit tricky here as we use log mean as an estimation, not just the mean. But there're some work around: https://stats.stackexchange.com/questions/418313/variance-of-x-and-variance-of-logx-how-to-relate-them
>        But in any case, we will need to compute empirical var(P_F(tau) / P_B (tau)) / n_traj for each sample and then play around a bit with it to get variance for the log mean estimation.

I agree. I will not do this in this PR so as to merge asap. I have opened an issue instead: https://github.com/alexhernandezgarcia/gflownet/issues/192

michalkoziarski (approved):
Thank you for the changes and the tests - looks good to me!

gflownet/gflownet.py
alexhernandezgarcia: Actually, it doesn't rely on that because of the indexing provided by `data_indices` and `traj_indices`. I hope it is clearer now.

PR #168, opened 2023-08-01
[FIX] Crystal Env: getter and setter for stage instead of `self.stage`

alexhernandezgarcia:
This addresses issues identified when attempting the merge into `main`. This branch already has `main` merged into it.

This is a safer way of handling the stage. For example, the stage was not set in `set_state()` and that was the source of errors. There was also a bug somewhere where the stage was obtained with `stage = Crystal(state[0])`. 

The same mechanism has been implemented in the Tree environment.


michalkoziarski (approved):

carriepl (approved):
LGTM

PR #169, opened 2023-08-01
Tree: cosmetic changes for readability

alexhernandezgarcia:
I got crazy changing things because I thought something was wrong and in the end things are back (I believe) to what we had before. However, at least, the code should be more readable.

The policy output and mask passed to the discrete `sample_actions()" does contain an element corresponding to the "continuous action", but that is ok because it will always be masked out as invalid if a discrete action is to be sampled.


tests/gflownet/envs/test_tree.py
```diff
@@ -573,6 +573,11 @@ def test__get_masks_backward__returns_expected(
     assert tree_d3.get_mask_invalid_actions_backward() == mask
 
 
+def test__all_env_common__standard__treedd3(tree_d3):
```
michalkoziarski, line 576:
```suggestion
def test__all_env_common__standard__tree_d3(tree_d3):
```
Was that intended or a typo?

michalkoziarski (approved):
Looks good to me, though I would merge it into the `rf_gfn` for now, since I had some issues running the code after pulling from `main`.

PR #170, opened 2023-08-01
Space group: fix compatibility constraints

alexhernandezgarcia:
Checking compatibility with composition was missing for the case where neither the crystal-lattice system nor the point symmetry were set (source), which could lead to dead ends. This PR aims at fixing that. However, some issues remain.


carriepl (approved):
Looks good to me. Nice work!

PR #173, opened 2023-08-08
Setup continuous integration

carriepl:
This PR puts in place some continuous integration for the repo. The way it is setup, every time there is a push to a branch containing this code, it triggers 3 things : 
1- Validating the code formatting using Black (here, Black is not configured to change the files, only check them)
2- Validating the import formatting using isort (here, isort is not configured to change the files, only check them)
3- Running the unit tests using pytest
As soon as one step fails, the process stops. Assuming no error, the whole process currently takes about 7-8minutes.

The result of this check is displayed near the PR merge button but, as currently configured, it doesn't prevent a PR from being merged.


alexhernandezgarcia (approved):
Thanks so much for setting this up! I have no experience with Github's CI, so I can't really provide a proper review but what I see looks good to me.

PR #175, opened 2023-08-17
RF-GFN: bug fixes and improvements

michalkoziarski:


gflownet/gflownet.py
```diff
@@ -475,12 +460,27 @@ def sample_batch(
                 dict_replay = pickle.load(f)
                 n_replay = min(n_replay, len(dict_replay["x"]))
                 envs = [self.env.copy().reset(idx) for idx in range(n_replay)]
-                # TODO: implement other sampling options besides permutation
                 if n_replay > 0:
                     x_replay = [x for x in dict_replay["x"].values()]
```
alexhernandezgarcia, line 464: Convert to list through `list()` instead of through iteration (a bit faster and more readable). This was my code, admittedly.
```suggestion
                    x_replay = list(dict_replay["x"].values())
```

gflownet/gflownet.py
```diff
@@ -475,12 +460,27 @@ def sample_batch(
...
-            actions = []
-            valids = []
+                    if self.replay_sampling == "permutation":
+                        x_replay = [
+                            x_replay[idx] for idx in self.rng.permutation(n_replay)
+                        ]
+                    elif self.replay_sampling == "weighted":
+                        x_rewards = np.array(
+                            [r for r in dict_replay["rewards"].values()]
+                        )
```
alexhernandezgarcia, line 472: Convert to numpy array through `np.fromiter` instead of through iteration (a bit faster and more readable).
```suggestion
                        x_rewards = np.fromiter(dict_replay["rewards"].values(), dtype=float)
```

gflownet/gflownet.py
```diff
@@ -443,8 +430,6 @@ def sample_batch(
                 # TODO: implement other sampling options besides permutation
                 # TODO: this converts to numpy
                 x_tr = self.rng.permutation(dict_tr["x"])
-            actions = []
-            valids = []
```
alexhernandezgarcia, line 447: Note to self (to be directly resolved): removed because unused.

gflownet/gflownet.py
```diff
@@ -152,9 +142,6 @@ def __init__(
                 print("Reloaded GFN backward policy model Checkpoint")
         else:
             self.logger.set_backward_policy_ckpt_path(None)
-        self.ckpt_period = policy.ckpt_period
-        if self.ckpt_period in [None, -1]:
-            self.ckpt_period = np.inf
```
alexhernandezgarcia, line 157: Note to self (to be directly resolved): removed because unused.

gflownet/gflownet.py
```diff
@@ -475,12 +460,27 @@ def sample_batch(
                 dict_replay = pickle.load(f)
                 n_replay = min(n_replay, len(dict_replay["x"]))
                 envs = [self.env.copy().reset(idx) for idx in range(n_replay)]
-                # TODO: implement other sampling options besides permutation
```
alexhernandezgarcia, line 478: Note to self (to be directly resolved): new code about replay options looks good to me.

gflownet/gflownet.py
```diff
@@ -789,6 +789,10 @@ def test(self, **plot_kwargs):
             density_pred = np.array([hist[tuple(x)] / z_pred for x in x_tt])
             log_density_true = np.log(density_true + 1e-8)
             log_density_pred = np.log(density_pred + 1e-8)
+        elif self.buffer.test_type == "random":
+            # TODO: refactor
+            env_metrics = self.env.test(x_sampled)
+            return self.l1, self.kl, self.jsd, (None,), env_metrics
```
alexhernandezgarcia, line 795: Note to self (to be directly resolved): simply `elif` was moved up. Looks fine.

gflownet/policy/simple_tree.py
```diff
@@ -0,0 +1,163 @@
+from typing import Optional
+
+import torch
+import torch_geometric
+from torch_geometric.data import Batch
+from torch_geometric.nn import global_add_pool
+
+from gflownet.envs.tree import Tree
```
alexhernandezgarcia, line 8: Rather minor. Since only the (static) method `Tree.to_pyg()` is used in this class, shall we instead import only this method instead of the whole class? 

And actually I would rename `Tree.to_pyg()` as `Tree.state_to_pyg()` or `Tree.state2pyg()` to make it more explicit. But just my personal preference.

gflownet/policy/multihead_tree.py
```diff
@@ -0,0 +1,502 @@
+from typing import Optional
+
+import torch
+import torch_geometric
+from torch_geometric.data import Batch
+from torch_geometric.nn import global_add_pool
+from torch_geometric.utils import unbatch
+
+from gflownet.envs.tree import Attribute, Stage, Tree
```
alexhernandezgarcia, line 9: Rather minor. Similar as for `simple_tree.py` but I will repeat (and adapt) here: Since only the (static) methods `Tree.to_pyg()` and `Tree._find_active()` are used in this class, shall we instead import only these methods instead of the whole class? 

And actually, shall we also remove the underscore of `_find_active` since it is used outside the class, or keep it private? I am not sure of the convention.

gflownet/policy/simple_tree.py
```diff
@@ -0,0 +1,163 @@
...
+from gflownet.policy.base import Policy
+
+
+class Backbone(torch.nn.Module):
```
alexhernandezgarcia, line 12: Also here some docstring would be great :)

gflownet/policy/simple_tree.py
```diff
@@ -0,0 +1,163 @@
...
+            data.x,
+            data.edge_index,
+            data.edge_attr,
+            data.batch,
+        )
+
+        return self.model(x, edge_index, edge_attr, batch)
+
+
+class Head(torch.nn.Module):
```
alexhernandezgarcia, line 58: Docstring. Here. Too. If possible.

gflownet/policy/simple_tree.py
```diff
@@ -0,0 +1,163 @@
...
+            data.batch,
+        )
+        x = self.backbone(data)
+        x = global_add_pool(x, batch)
+        x = self.head(x)
+
+        return x
+
+
+class SimpleTreeModel(torch.nn.Module):
```
alexhernandezgarcia, line 101: gnirtscod. Just brief would do ;)

gflownet/policy/simple_tree.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+    def forward(self, x):
+        batch = Batch.from_data_list(
+            [Tree.to_pyg(state, self.n_features) for state in x]
+        )
+        return self.model(batch)
+
+
+class SimpleTreePolicy(Policy):
```
alexhernandezgarcia, line 131: Yup, here too :P

gflownet/policy/multihead_tree.py
```diff
@@ -0,0 +1,502 @@
+from typing import Optional
```
alexhernandezgarcia, line 1: At risk of being very annoying: docstring. Seriously, just a one- or two-liner would do.

gflownet/policy/multihead_tree.py
```diff
@@ -0,0 +1,502 @@
...
+import torch_geometric
+from torch_geometric.data import Batch
+from torch_geometric.nn import global_add_pool
+from torch_geometric.utils import unbatch
+
+from gflownet.envs.tree import Attribute, Stage, Tree
+from gflownet.policy.base import Policy
+
+
+class Backbone(torch.nn.Module):
```
alexhernandezgarcia, line 13: Could the Backbone of the simple tree be reused here? Alternatively, could we perhaps have a `tree.py` with the common code for both? Not sure what would be better.

gflownet/policy/multihead_tree.py
```diff
@@ -0,0 +1,502 @@
+from typing import Optional
+
```
alexhernandezgarcia, line 2: Note: I have taken a look at the structure of the class and some details but I haven't done a detailed review of all, because I prefer to review the rest of the PR first and delve into this later (after merging, I think).

gflownet/policy/simple_tree.py
```diff
@@ -0,0 +1,163 @@
+from typing import Optional
```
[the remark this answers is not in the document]
michalkoziarski: I added short docstrings for all of the policy classes.

gflownet/policy/multihead_tree.py
michalkoziarski: So right now the backbones are actually slightly different - one uses `GINEConv` layers, and the other `GCNConv` (and there are some differences in how they have to be constructed, `GINEConv` requires specifying an MLP model used by the layer, and also uses edge features.

I will simplify that in the end (and keep a single backbone class), but I haven't figured out yet which is better.

gflownet/policy/multihead_tree.py
michalkoziarski: As above, I renamed the method, and unless you can import some static methods directly I would personally keep it as is.

main.py
```diff
@@ -39,11 +45,38 @@ def main(config):
...
+    # The policy is used to model the probability of a forward/backward action
+    forward_config = OmegaConf.create(config.policy)
+    forward_config["config"] = config.policy.forward
+    del forward_config.forward
+    del forward_config.backward
+
+    backward_config = OmegaConf.create(config.policy)
+    backward_config["config"] = config.policy.backward
+    del backward_config.forward
+    del backward_config.backward
```
alexhernandezgarcia, line 57: Just a note for now. I am not super happy about this solution, since it is a bit convoluted and makes the code less readable. I have started working on figuring out a simpler and functional alternative way, but I don't have it yet. In any case, in practical terms: ok to merge to `rf_gfn` but for `main` I would like to find a simpler way.

gflownet/policy/multihead_tree.py
alexhernandezgarcia: Ok fine. We can keep them separate for now, no problem. And possibly combine things later.

gflownet/policy/simple_tree.py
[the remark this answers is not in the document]
alexhernandezgarcia: Awesome, thanks a lot!

gflownet/policy/multihead_tree.py
alexhernandezgarcia: Alright!

gflownet/policy/simple_tree.py
alexhernandezgarcia: Thanks! Right right that was stupid of me...

main.py
```diff
@@ -16,6 +16,12 @@
...
     # Get current directory and set it as root log dir for Logger
     cwd = os.getcwd()
+
+    # TODO: fix race condition in a more elegant way
+    cwd += "/%08x" % random.getrandbits(32)
+    os.mkdir(cwd)
+    os.chdir(cwd)
+
     config.logger.logdir.root = cwd
     print(f"\nLogging directory of this run:  {cwd}\n")
```
alexhernandezgarcia, line 26: Since this is pretty low level, I think we should put it outside as a method in utils. What do you think?

michalkoziarski: Moved to `utils.common` (with docstring describing what it's trying to solve).

alexhernandezgarcia: Nice thanks!

gflownet/envs/tree.py
```diff
@@ -254,22 +274,27 @@ def __init__(
...
-        self.state2policy = self.state2policy_mlp
-        self.statetorch2policy = self.statetorch2policy_mlp
-        # self.statetorch2proxy = self.statetorch2policy
+        if policy_type == "mlp":
+            self.state2policy = self.state2policy_mlp
+            self.statetorch2policy = self.statetorch2policy_mlp
+        elif policy_type != "gnn":
+            raise ValueError(
+                f"Unrecognized policy_type = {policy_type}, expected either 'mlp' or 'gnn'."
+            )
```
alexhernandezgarcia, line 291: What happens if `policy_type` is `gnn`, is it in the other PR?

michalkoziarski: Then we use default `state2policy` (returning complete state), which is converted to a graph in the policy itself (to avoid issues with current GFN code expecting tensors to be passed around).

alexhernandezgarcia: OK I see, I forgot the conversion happens _in_ the policy.

alexhernandezgarcia (changes requested):
See my comments

gflownet/envs/tree.py
```diff
@@ -1174,6 +1492,51 @@ def _compute_scores(states: torch.Tensor, X: npt.NDArray, y: npt.NDArray) -> dic
 
         return scores
 
+    @staticmethod
+    def _plot_trees(
```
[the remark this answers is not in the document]
michalkoziarski: This was modified in the later PR (the name is the same, but the top-k tree selection is done outside this method).

gflownet/envs/tree.py
```diff
@@ -1174,6 +1492,51 @@ def _compute_scores(states: torch.Tensor, X: npt.NDArray, y: npt.NDArray) -> dic
...
+        states: torch.Tensor,
+        X: npt.NDArray,
+        y: npt.NDArray,
+        iteration: int,
+        n: int = -1,
```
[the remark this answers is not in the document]
michalkoziarski: Modified in the later PR.

gflownet/envs/tree.py
```diff
@@ -533,6 +611,49 @@ def step(
             self.done = True
             return self.state, action, True
 
+    def step_backwards(
```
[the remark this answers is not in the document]
michalkoziarski: Changed (in a way that fixes replay sampling issue).

gflownet/envs/tree.py
```diff
@@ -500,29 +567,40 @@ def step(
...
+            # Replace the continuous value of threshold by -1 to allow checking it.
+            action_to_check = self.action2representative(action)
+            do_step, self.state, action_to_check = self._pre_step(
+                action_to_check,
+                skip_mask_check=(skip_mask_check or self.skip_mask_check),
+            )
+        else:
+            do_step, self.state, action_to_check = self._pre_step(
+                action, skip_mask_check=(skip_mask_check or self.skip_mask_check)
+            )
```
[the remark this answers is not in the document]
michalkoziarski: I think you are correct, changed.

gflownet/envs/tree.py
```diff
@@ -869,7 +1055,8 @@ def get_mask_invalid_actions_forward(
             for k in leaves:
                 # Check if splitting the node wouldn't exceed max depth.
                 if Tree._get_right_child(k) < self.n_nodes:
-                    mask[self._action_index_pick_leaf + k] = False
+                    current_class = state[k, Attribute.CLASS].long()
+                    mask[self._action_index_pick_leaf + 2 * k + current_class] = False
```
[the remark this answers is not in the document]
michalkoziarski: This is due to `action_space` change: to ensure compatibility with backward policy, pick leaf action now also stores the information about the current class (which is not really needed in the forward, but alas). Also, the actions in the `action_space` are in the order of `(leaf_0, 0), (leaf_0, 1), (leaf_1, 0), (leaf_1, 1), ...`, where `leaf_i` is the index of the leaf, and the second element is the class (either 0 or 1). Thus, the above just maps from the current class to the appropriate action index (which points to the action, for which current class is picked).

gflownet/envs/tree.py
```diff
@@ -1001,53 +1217,127 @@ def get_parents(
...
                 parents.append(parent)
-                actions.append((ActionType.PICK_THRESHOLD, -1))
+                if self.continuous:
+                    actions.append((ActionType.PICK_THRESHOLD, -1, -1))
+                else:
+                    threshold_idx = np.where(
+                        np.isclose(
+                            self.thresholds, state[k][Attribute.THRESHOLD].item()
+                        )
+                    )[0].item()
```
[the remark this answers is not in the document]
michalkoziarski: 1. If anything, it would avoid the inner call (the outer is necessary anyway I think, since we need to go from the tensor returned by `torch.where` to an `int` that can be used to construct an action); but that would require converting `self.thresholds` to a tensor - and it's being used in a couple of different places with direct access, so I am not sure if that would even be more efficient.
2. It definitely should! And was in all of my experiments.

michalkoziarski:
@alexhernandezgarcia note that I changed the target branch to `main` (since I merged it into this, and didn't want the diff to be a mess).

alexhernandezgarcia:
> @alexhernandezgarcia note that I changed the target branch to `main` (since I merged it into this, and didn't want the diff to be a mess).

@michalkoziarski have run any of the sanity checks to make sure that things will run normally after merging into main? See here: https://github.com/alexhernandezgarcia/gflownet/blob/main/mila/dev/sanity_check_runs.md

gflownet/envs/tree.py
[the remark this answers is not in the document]
alexhernandezgarcia: Alright, thanks for explaining!

alexhernandezgarcia:
@michalkoziarski now that main has been merged into `rf_gfn-mk` we need to make sure that the continuous integration works. In particular, it is currently failing (at least) because dependencies are missing. Would you mind adding the necessary packages to https://github.com/alexhernandezgarcia/gflownet/blob/rf_gfn-mk/setup/continuous_integration.sh

michalkoziarski:
> @michalkoziarski now that main has been merged into `rf_gfn-mk` we need to make sure that the continuous integration works. In particular, it is currently failing (at least) because dependencies are missing. Would you mind adding the necessary packages to https://github.com/alexhernandezgarcia/gflownet/blob/rf_gfn-mk/setup/continuous_integration.sh

CI and tests are fixed in https://github.com/alexhernandezgarcia/gflownet/pull/187 (which should be merged into https://github.com/alexhernandezgarcia/gflownet/pull/177, which on the other hand should be merged into this, to allow this being merge into `main`).

michalkoziarski:
> > @alexhernandezgarcia note that I changed the target branch to `main` (since I merged it into this, and didn't want the diff to be a mess).
> 
> @michalkoziarski have run any of the sanity checks to make sure that things will run normally after merging into main? See here: https://github.com/alexhernandezgarcia/gflownet/blob/main/mila/dev/sanity_check_runs.md

Not really, no, the changes to the shared GFN code were small enough that I didn't even think about it; I can try scheduling some of them now.

alexhernandezgarcia:
I would consider the changes in the policy a significant change no?

On 30 August 2023 18:48:11 GMT-04:00, "Michał Koziarski" ***@***.***> wrote:
>> > @alexhernandezgarcia note that I changed the target branch to `main` (since I merged it into this, and didn't want the diff to be a mess).
>> 
>> @michalkoziarski have run any of the sanity checks to make sure that things will run normally after merging into main? See here: https://github.com/alexhernandezgarcia/gflownet/blob/main/mila/dev/sanity_check_runs.md
>
>Not really, no, the changes to the shared GFN code were small enough that I didn't even think about it; I can try scheduling some of them now.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/alexhernandezgarcia/gflownet/pull/175#issuecomment-1699953052
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-- Sent from /e/ Mail.

michalkoziarski:
@alexhernandezgarcia I did sanity checks for grid (both variants) and ctorus, after reverting backward policy to be null by default in [here](https://github.com/alexhernandezgarcia/gflownet/pull/175/commits/9b2b228824ce671850251d25bcb1cd951a3d9793), everything [looks the same to me](https://wandb.ai/michalkoziarski/gfn_sanity_checks).

alexhernandezgarcia:
> @alexhernandezgarcia I did sanity checks for grid (both variants) and ctorus, after reverting backward policy to be null by default in [here](https://github.com/alexhernandezgarcia/gflownet/pull/175/commits/9b2b228824ce671850251d25bcb1cd951a3d9793), everything [looks the same to me](https://wandb.ai/michalkoziarski/gfn_sanity_checks).

Thanks for running the sanity checks. It's hard though to compare without the reference runs in the `main` branch. 

I have anyway run my own sanity check experiments and everything looks fine except the Grid with TB. I will message you outside the PR.

michalkoziarski:
As discussed, the differences were caused by no backward policy being used by default with the new configs. The issue here is, I have no idea how to override a config from _higher level_, e.g. modify `policy.backward` in `gflownet/flowmatch.yaml` (previously this could have been done, because the `policy` was just one of the properties of `gflownet`; now it's a separate config).

What I ended up doing [in here](https://github.com/alexhernandezgarcia/gflownet/pull/175/commits/738e7ffaa20f103007c79215567ab367b6515b9d) was: I created a separate configs for `mlp_flowmatch` and `mlp_trajectorybalance`, and used [variable interpolation](https://hydra.cc/docs/patterns/specializing_config/) to dynamically select the proper one based on the loss. I also created an `mlp` config to ensure backwards compatibility (some configs try to use `mlp` as default). This seems to [perfectly reproduce main](https://wandb.ai/michalkoziarski/gfn_sanity_checks) on both grid variants; ctorus is running right now (with neither `gflownet=trajectorybalance` nor `gflownet=flowmatch` being specified).

alexhernandezgarcia (approved):
Thanks for the changes with the config and making sure the sanity checks pass! Feel free to merge at your convenience! ;) And I hope you become friends with hydra soon :P

PR #177, opened 2023-08-24
Second batch of RF-GFN changes

michalkoziarski:
A bunch of smaller changes, including:
- dataset cleanup,
- modified format of `state2policy_mlp`: activity information removed, but currently active node explicitly concatenated to the state,
- metric computation based on top-k trees,
- exponential prior.


gflownet/envs/tree.py
```diff
@@ -824,17 +824,21 @@ def state2policy_mlp(
...
     ) -> TensorType["batch_size", "policy_input_dim"]:
         """
         Prepares a batch of states in torch "GFlowNet format" for an MLP policy model.
-        It simply replaces the NaNs by -1s.
+        It replaces the NaNs by -2s, removes the activity attribute, and explicitly
+        appends the attribute vector of the active node (if present).
         """
+        m, n = torch.where(states[:, :-1, Attribute.ACTIVE] == Status.ACTIVE)
+        active_features = torch.full((states.shape[0], 1, 4), -2.0)
+        active_features[m] = states[m, n, : Attribute.ACTIVE].unsqueeze(1)
```
alexhernandezgarcia, line 839: `m` and `n` are rows and columns right?

Minor suggestion: if that is the case, I would use `rows` and `cols` instead just to make it a bit easier to understand.

gflownet/proxy/tree.py
```diff
@@ -11,21 +11,20 @@
 class TreeProxy(Proxy):
     """
     Simple decision tree proxy that uses empirical frequency of correct predictions for
-    computing likelihood, and the number of leafs in the tree for computing the prior.
+    computing likelihood, and the number of nodes in the tree for computing the prior.
     """
 
-    def __init__(self, use_prior: bool = True, **kwargs):
+    def __init__(self, use_prior: bool = True, beta: float = 1.0, **kwargs):
```
alexhernandezgarcia, line 17: Little docstring for `use_prior` and `beta`? :sweat_smile:

gflownet/envs/tree.py
michalkoziarski: Yes, modified as suggested.

gflownet/envs/tree.py
```diff
@@ -1539,24 +1535,53 @@ def test(
...
+        if self.test_args["top_k_trees"] != 0:
             if not hasattr(self, "test_iteration"):
                 self.test_iteration = 0
+
+            # Select top-k trees.
+            accuracies = np.array(
+                [accuracy_score(self.y_train, y_pred) for y_pred in train_predictions]
+            )
+            order = np.argsort(accuracies)[::-1]
+            top_k_indices = order[: self.test_args["top_k_trees"]]
```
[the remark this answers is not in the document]
michalkoziarski: 1. I am not sure this actually works (but maybe I mixed something up)?
```
>>> top_k = 3
>>> order = np.arange(10)
>>> order[: top_k : -1]
array([9, 8, 7, 6, 5, 4])
```
3. It gave my a pause to parse the syntax, I think my version is simpler to understand - hence I will keep it as is.

gflownet/proxy/tree.py
michalkoziarski: I added something, though all of this should be re-implemented anyway.

gflownet/envs/tree.py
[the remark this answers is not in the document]
alexhernandezgarcia: Oops right, it should be `order[: top_k : -1]`, but ok to leave as is.

PR #178, opened 2023-08-25
Make main work again, other fixes, new defaults, readmes and sanity check runs 

alexhernandezgarcia:
The main goal of this PR is to make the `main` branch work again: due to the multi-way merges, the part of the code that computes topK evaluation metrics had not been adapted to the new way of sampling batches. I have _partially_ fixed this (in `gflownet.py`) but still the code in `top_k_metrics_and_plots()` (`envs/base.py`) crashes (see [Line 1069](https://github.com/alexhernandezgarcia/gflownet/compare/main...fixes-defaults-readmes#diff-700050cd4007666a2c357edc47c03336bdbaddab3e135d60a3f382718580780bR1069)). I leave fixing this for later, In any case, the code runs without issues because I have also disabled the topK evaluation _by default_.

Other changes:
- I have changed a few default configuration variables:
  - No wandb logging by default because this happens to be the most used mode during development, I think. It can be enabled with `logger.do.online=True`.
  - No sampling of train data set by default in the grid.
  - No replay buffer by default. It can be enabled with `env.buffer.replay_capacity` > 0.
- I have fixed `mila/launch.py` so that it works with `virtualenv` environments. In particular, `--template` was missing in the arguments parser (perhaps this was deliberate sabotage by @vict0rsch :stuck_out_tongue_closed_eyes:). More seriously, I have used `mila/launch.py` now for the first time and it's great, so many thanks!
- I have extended a bit the READMEs.
- I have created a document `mila/dev/sanity_check_runs.md` with information about sanity check runs for future reference.
- Other very minor fixes and changes.
- The PR shows a mind boggling **450 Files changed** because I have also removed the directory `potts` which contained an old data set that does not need to be in the repo and actually took significant disk space.

I have run the usual sanity check experiments, available on [wandb](https://wandb.ai/alexhg/gfn_sanity_checks) and everything seems alright.


mila/dev/sanity_check_runs.md
```diff
@@ -0,0 +1,101 @@
+# Dev Sanity Check Runs
+
+Here is a list of commands to train GFlowNets on various on environments with different configurations that can be used as sanity checks during development of the repository. A number of such runs can be found in [alexhg's `gfn_sanity_checks` wandb project](https://wandb.ai/alexhg/gfn_sanity_checks).
```
carriepl, line 3: Good idea to add this. 
Would it be pertinent to add some info on validating those sanity checks? Like "you expect [some metric] to be around [some value]"

gflownet/envs/base.py
```diff
@@ -1066,6 +1066,7 @@ def top_k_metrics_and_plots(
             name = "train ground truth"
             reward = energy_reward
         elif energy is None and reward is None:
+            # TODO: fix this
```
carriepl, line 1069: Is it worth documenting what is broken here? At glance, I can't really tell.

gflownet/envs/README.md
```diff
@@ -11,3 +11,35 @@ By default, this redundancy is in place. However, the user may choose to manuall
...
+
+## Buffer, train data and test data
+
+A train and a test set can be created at the beginning of training. The train set may be used to sample offline (backward) trajectories. The test set may be used to compute metrics during and after training. These sets may be created in different ways, specificied by the configuration variables `env.buffer.train.type` and `env.buffer.test.type`. Options for the data set `type` are
+
+- `all`: all terminating states in the output space $\mathcal{X}$ will be added - Convenient but only feasible for small, synthetic environments like the hyper-grid.
+- `grid`: a grid of points in the output space $\mathcal{X}$ - Only available in certain environments where obtaining a grid of points is meaningful. This mode also requires specifying the number of points via `env.buffer.<train/test>.n`.
+- `uniform`: points sampled uniformly output space $\mathcal{X}$ - This mode also requires specifying the number of points via `env.buffer.<train/test>.n` and the random seed via `env.buffer.<train/test>.seed`.
+
```
michalkoziarski, line 22: There is also a `random` option, why is this not mentioned? It's also the most general one, so might actually be most relevant.

gflownet/envs/README.md
```diff
@@ -11,3 +11,35 @@ By default, this redundancy is in place. However, the user may choose to manuall
...
+    type: all
+    output_csv: grid_test.csv
+    output_pkl: grid_test.pkl
+```
+
+The generation of a train and/or test set can be disabled by sampling their configuration to None/null (`env.buffer.train=null` via the command line).
+
+### Replay buffer
+
+It is possible to store the best trajectories, according to the reward, in a buffer and potentially use them for backward sampling or simply for evaluation during or at the end of training. In order to enable the replay buffer, its capacity must be larger than zero (default), which can be set with the configuration variable `env.buffer.replay_capacity`.
```
michalkoziarski, line 45: I guess this is not yet in this branch, but I wonder if it would be worth it to preemptively describe different variants of replay buffer (`permutation` and `weighted`), and how to use them in the config (`gflownet.replay_sampling: {permutation, weighted}` - since we will probably forget about adding this description later.

michalkoziarski (approved):
Other than README comments, looks good to me!

gflownet/gflownet.py
```diff
@@ -898,15 +899,17 @@ def test_top_k(self, it, progress=False, gfn_states=None, random_states=None):
...
-                gfn_states += self.sample_batch(
-                    self.env, len(b), train=False, progress=progress
-                )[0]
+            for b in batch_with_rest(
+                0, self.logger.test.n_top_k, self.batch_size_total
+            ):
+                sub_batch, _ = self.sample_batch(n_forward=len(b), train=False)
+                batch.merge(sub_batch)
             duration = time.time() - t
+            gfn_states = batch.get_terminating_states()
```
carriepl, line 912: For my own understanding : 
Is this change (and the following one a few lines down) merely for code consistency (using the new Batch class) or does it also solve a bug at the same time?

carriepl (commented):
Looks pretty good to me.
I had only two very small comments as well as a question for my own edification.
Thanks!

gflownet/envs/README.md
alexhernandezgarcia: Right, it's because I wrote this a while ago and things have changed, so thanks for pointing this out. I will add that.

gflownet/envs/README.md
alexhernandezgarcia: For sure, wanna take a stab at that? :smile:

mila/dev/sanity_check_runs.md
alexhernandezgarcia: Actually I expect all the training curves to be relative the same as in previous runs. Shall I write that?

gflownet/envs/base.py
alexhernandezgarcia: I don't know, to be honest. I haven't looked deeper into it, which is why I haven't fixed it. I will just fix it soon once I will actually need it for the crystals experiments.

gflownet/envs/README.md
alexhernandezgarcia: Done [here](https://github.com/alexhernandezgarcia/gflownet/pull/178/commits/87ba5e50102f35164e92fea13c63071d9c53fb3b).

mila/dev/sanity_check_runs.md
alexhernandezgarcia: How about [this](https://github.com/alexhernandezgarcia/gflownet/pull/178/commits/d4878c1ef5ca250c90c0caa5d9dafa318feb9c11)?

gflownet/gflownet.py
alexhernandezgarcia: The previous way simply didn't work (it did work when it was implemented by Victor, so not strictly a bug) because the interface of `gflownet.sample_batch()` changed with the Batch v2.

mila/launch.py
```diff
@@ -395,6 +395,11 @@ def print_md_help(parser, defaults):
         help="path to venv (without bin/activate)."
         + f" Defaults to {defaults['venv']}",
     )
+    parser.add_argument(
```
vict0rsch, line 398: Woups, my bad 😅

vict0rsch (changes requested):
Thanks for fixing things!

There's just a typo in the launch commands with `conda`, for sobriety I only flagged the first one but they all need a tiny adaptation

vict0rsch:
I've actually taken the liberty to fix things myself.

@alexhernandezgarcia if you need to update `launch.py` more in the future, and especially if you change the `args` (add/remove arg or update its `help` message), you can run 

```
python mila/launch.py --help-md > LAUNCH.md
```

in order to automatically update the launch instructions in `LAUNCH.md`.

Note that you should *not* update that file directly, rather change the `HELP` variable at the top of `launch.py`.

These (⬆️) instructions have been added to `LAUNCH.md` (well, in `launch.py:HELP` actually 😏)

mila/dev/sanity_check_runs.md
carriepl: Yup, that looks great.

gflownet/envs/README.md
michalkoziarski, line 46:
```suggestion
It is possible to store the best trajectories, according to the reward, in a buffer and potentially use them for backward sampling or simply for evaluation during or at the end of training. In order to enable the replay buffer, its capacity must be larger than zero (default), which can be set with the configuration variable `env.buffer.replay_capacity`.

To use the replay buffer (once enabled) for backward sampling, one can specify `gflownet.optimizer.batch_size.backward_replay`. Currently, backward sampling is supported in one of two modes (specified using `gflownet.replay_sampling`):
- `permutation` (default): terminating states from the replay buffer are randomly permuted at each iteration, and first `gflownet.optimizer.batch_size.backward_replay` of them are used to sample trajectories,
- `weighted`: terminating states are sampled with probability equal to their reward.
```

alexhernandezgarcia: Perfect, thanks!

vict0rsch:
@alexhernandezgarcia I've also pushed a couple commits that will hopefully allow to scale those sanity checks / integration tests:

1. `--jobs` can now point to any specific file (before the jobs `yaml` file had to be in `external/jobs`)
2. I've synthesized checks in `mila/dev/sanity_check_runs.yaml` such that you can simply add future sanity checks and still have a single command to run them:
	```
	python mila/launch.py --venv=<path-to-venv> --template=mila/sbatch/template-venv.sh --jobs=mila/dev/sanity_check_runs.yaml
    ```
	* This ^ is also documented in `sanity_check_runs.md`

vict0rsch (approved):

alexhernandezgarcia:
@vict0rsch thanks for the improvements on the launch functionality, but I have a suggestion: could the last 2 commits ([43c9e7d](https://github.com/alexhernandezgarcia/gflownet/pull/178/commits/43c9e7dccd44511ce0a8e7890a39446bed512734) and [a8d25ef](https://github.com/alexhernandezgarcia/gflownet/pull/178/commits/a8d25efd29fb9edeb34ee2e59e84a89c678bead4) go into a separate PR? Just in the spirit of keeping PRs small (when possible) and because the main goal of this PR was a fix in the main branch and I would like to merge it as soon as possible - and I would like to give a proper review to your latest changes :)

vict0rsch:
I can, but i'm done so we can also just stick to this and I'll do a PR if I need more changes. What do you think?

vict0rsch:
Also, in the tetris sanity check:

```
Error executing job with overrides: ['user=schmidtv', 'device=cpu', 'logger.project_name=gfn_sanity_checks', 'logger.do.online=True', 'env=tetris', 'env.width=5', 'env.height=10', 'gflownet=
flowmatch']
Error in call to target 'gflownet.envs.tetris.Tetris':
AttributeError("'Tetris' object has no attribute 'n_dim'")
full_key: env
```

Is this expected?

alexhernandezgarcia:
> I can, but i'm done so we can also just stick to this and I'll do a PR if I need more changes. What do you think?

If you don't mind if I don't do proper review (at least now), then fine. If review is important, then I'd prefer separate PR. (I'm fixing/testing an important bug at the moment).

vict0rsch:
> > I can, but i'm done so we can also just stick to this and I'll do a PR if I need more changes. What do you think?
> 
> If you don't mind if I don't do proper review (at least now), then fine. If review is important, then I'd prefer separate PR. (I'm fixing/testing an important bug at the moment).

I think we're good, it's just small stuff ; the only thing that may need an eye would be the jobs file (`mila/dev/sanity_check_runs.yaml`) but that's not for review, rather FYI for future sanity checks.

PR #180, opened 2023-08-25
Change max number of atoms per elements in crystals

carriepl:
Tiny PR to change the max number of atoms per elements from 10 to 16 in the crystal/composition environment.

This is because of the following spacegroups for which the default value prevent generating a valid composition : 
- Spacegroup 220 : requires at least 12 atoms per element
- Spacegroup 228 : requires at least 16 atoms per element
- Spacegroup 230 : requires at least 16 atoms per element


alexhernandezgarcia (approved):
I love tiny PRs! :smile: Makes sense and thanks!

PR #183, opened 2023-08-28
[IMPORTANT BUG FIX] Fix that random action prob. > 0 was doing nothing

alexhernandezgarcia:
Small but important PR, I guess the title is self-explanatory. But for context: I changed the method `gflownet.sample_actions()` (as well as `gflownet.sample_batch()` quite a bit in the Batch v2 PR and admittedly I didn't test everything in depth. As you can see, the default of `random_action_prob` should have been `None` and I kept it in 0.0, which made that `self.random_action_prob` (which contains the random action probability passed as argument/config) was never picked up. 

I am sorry for the impact of this bug!

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -202,7 +202,7 @@ class GFlowNetAgent:
         sampling_method: Optional[str] = "policy",
         backward: Optional[bool] = False,
         temperature: Optional[float] = 1.0,
-        random_action_prob: Optional[float] = 0.0,
+        random_action_prob: Optional[float] = None,
         no_random: Optional[bool] = True,
         times: Optional[dict] = None,
     ) -> List[Tuple]:
@@ -241,8 +241,9 @@ class GFlowNetAgent:
             self.temperature_logits is used.
 
         random_action_prob : float
-            Probability of sampling random actions. If None, self.random_action_prob is used.
-
+            Probability of sampling random actions. If None (default),
+            self.random_action_prob is used, unless its value is forced to either 0.0
+            or 1.0 by other arguments (sampling_method or no_random).
         no_random : bool
             If True, the samples will strictly be on-policy, that is
                 - temperature = 1.0


gflownet/gflownet.py
```diff
@@ -202,7 +202,7 @@ def sample_actions(
         sampling_method: Optional[str] = "policy",
         backward: Optional[bool] = False,
         temperature: Optional[float] = 1.0,
-        random_action_prob: Optional[float] = 0.0,
+        random_action_prob: Optional[float] = None,
```
michalkoziarski, line 205: Why is this even necessary? Is manually passing `random_action_prob` ever used? (I tried to quickly search for it on my branch and I couldn't find any such cases, but maybe I missed something) And if not, maybe just remove this argument?

alexhernandezgarcia: Yeah true it's not currently used I think. I thought it could be useful in certain situations. The one I am thinking is that we may want to have a schedule on the random_action_prob.

Also if we remove that argument, we should remove the temperature too (for consistency) and things would change significantly.

gflownet/gflownet.py
```diff
@@ -202,7 +202,7 @@ def sample_actions(
...
         no_random: Optional[bool] = True,
         times: Optional[dict] = None,
```
michalkoziarski, line 207: By the way, `times` argument is never used in the method.

alexhernandezgarcia: Yeah, I am aware. At some points I discontinued the support of calculating and logging times, but I would like to pick it up again at some point.

gflownet/gflownet.py
michalkoziarski: I am not sure if removing the temperature would be a significant change, but I see your point about having a scheduler. I am borderline on whether this should be kept for now or not, it's not a big difference either way; approving.

michalkoziarski (approved):

gflownet/gflownet.py
alexhernandezgarcia: Thanks for the review!

PR #184, opened 2023-08-28
Fix action selection in space group

carriepl:
This PR fixes a problem that was occuring when we provided a composition to the space group. 

Essentially, whenever the agent would like to pick a crystal-lattice system or a point symmetry, we would only allow option where at least one of the associated space groups was compatible with the composition.

The problem is that we would allow the agent to choose crystal-lattice systems and point symmetries that were individually compatible, but not jointly compatible.

We only ensured that each of them had 1+ space group compatible with the composition. But we didn't ensure that at least one of these space groups was shared by both the crystal-lattice system and the point symmetry.

### DIFF
--- a/gflownet/envs/crystals/spacegroup.py
+++ b/gflownet/envs/crystals/spacegroup.py
@@ -160,12 +160,12 @@ class SpaceGroup(GFlowNetEnv):
             crystal_lattice_systems = [
                 (self.cls_idx, idx + 1, ref)
                 for idx in range(self.n_crystal_lattice_systems)
-                if self._is_cls_compatible(idx + 1)
+                if self._is_compatible(cls_idx=idx + 1)
             ]
             point_symmetries = [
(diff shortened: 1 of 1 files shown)


alexhernandezgarcia (approved):
Thanks for the fix! It looks good to me. 

The only thing, if I were to be picky, is that since the code in the `else` is identical for both `_is_cls_compatible` and `_is_ps_compatible`, we may want to code it differently, but it really does not matter.

carriepl:
> Thanks for the fix! It looks good to me.
> 
> The only thing, if I were to be picky, is that since the code in the `else` is identical for both `_is_cls_compatible` and `_is_ps_compatible`, we may want to code it differently, but it really does not matter.

Good point. I refactored the two methods into a single one.

alexhernandezgarcia (approved):
That's great, thanks for the refactoring into a single method!

PR #185, opened 2023-08-28
Refactor spacegroup's ref-related stuff: Rename variables and add docstring

alexhernandezgarcia:
This PR barely changes if at all the functionality of the code. I have simply 1) renamed some variables and 2) added/updated the docstring and comments regarding the formerly (poorly) called `ref` in the space group. I have renamed it to "state type" and it refers to the four types of state from which an action can be executed in the space group environment:
- 0: both crystal-lattice system and point symmetry are unset (== 0)
- 1: crystal-lattice system is set (!= 0); point symmetry is unset
- 2: crystal-lattice system is unset; point symmetry is set
- 3: both crystal-lattice system and point symmetry are set

This is needed and is actually part of the actions and the action space so as to account for the possible multiple parents of a given state and the (important) detail that GFlowNet must learn a distribution over states and not over actions.

### DIFF
--- a/gflownet/envs/crystals/spacegroup.py
+++ b/gflownet/envs/crystals/spacegroup.py
@@ -61,16 +61,16 @@ class SpaceGroup(GFlowNetEnv):
         (230 options + none)
 
     The action space is the choice of property to update, the index within the property
-    and the combination of properties already set in the reference state (e.g.
-    crystal-lattice system 2 from source, point symmetry 4 from crystal-lattice system,
-    space group 69 from point symmetry, etc.). The reference state is included in the
-    action to differentiate actions that lead to same state from different states, as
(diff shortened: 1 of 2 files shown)


alexhernandezgarcia:
Funnily enough both @carriepl and @InfluenceFunctional have asked me about this within 2 hours today :smile: 

@michalkoziarski: I have added you as a reviewer because I think we coded this together or at least discussed the need for it. Feel free to ignore.

InfluenceFunctional (approved):
look good to me

carriepl (approved):
Looks good to me too. Much clearer docstrings. Thanks a lot!

carriepl:
I added a small commit to fix the CI and, with that, every test works for you branch so I see no reason not to merge your work.

PR #186, opened 2023-08-28
Minor changes on PR 175

alexhernandezgarcia:
Take these changes as suggestions only. Except adding the missing arguments to the config, which would be good to have there.

policy_type -> policy_format is because I have used policy_format elsewhere.


michalkoziarski (approved):
Looks good to me, thanks for suggestions!

PR #187, opened 2023-08-29
Updated tests for RF-GFN

michalkoziarski:
As in title, simply updating all of the tests to the changes made in two previous PRs.


alexhernandezgarcia (approved):
Without having made an in-depth review of the tests, things seem good to me! Unfortunately we'll probably revert some of these changes soon, but it's ok. Thanks for the extended tests!

PR #188, opened 2023-08-30
[WIP] Update dave proxy release

vict0rsch:
## 💥 Breaking changes: requires `pyxtal` package

Proxy trained on matbench formation energy with conventional cell

Proxy is an MLP with Physical embeddings for atomic species

CSV Data can be found in:

```
/network/scratch/s/schmidtv/crystals-proxys/data/materials_dataset_v0/data/matbench_mp_e_form/{train/val}_data.csv
```

Release details: https://github.com/sh-divya/ActiveLearningMaterials/releases/tag/v3-physmlp-mbform

**There currently is a bug here with mismatched dims**


alexhernandezgarcia:
Just to make sure, the only actual change is the [line specifying the release in the config](https://github.com/alexhernandezgarcia/gflownet/pull/188/files#diff-7e9dda8f0a8645db749b599683eabc74d9c9734a7671d8fb9f3d9e427b124457R3), right?

alexhernandezgarcia:
> There currently is a bug here with mismatched dims

Would you be able say more about this bug?

alexhernandezgarcia:
@vict0rsch hope you don't mind to follow up on this. I would like to merge this soon if possible. I would appreciate it if you could check my questions in comments above:
- https://github.com/alexhernandezgarcia/gflownet/pull/188#issuecomment-1701141459
- https://github.com/alexhernandezgarcia/gflownet/pull/188#issuecomment-1701142460

vict0rsch:
> @vict0rsch hope you don't mind to follow up on this. I would like to merge this soon if possible. I would appreciate it if you could check my questions in comments above:
> - https://github.com/alexhernandezgarcia/gflownet/pull/188#issuecomment-1701141459
> - https://github.com/alexhernandezgarcia/gflownet/pull/188#issuecomment-1701142460

I will. But I won't so late

vict0rsch:
> Just to make sure, the only actual change is the [line specifying the release in the config](https://github.com/alexhernandezgarcia/gflownet/pull/188/files#diff-7e9dda8f0a8645db749b599683eabc74d9c9734a7671d8fb9f3d9e427b124457R3), right?

Sorry I missed that ; yes this is currently the only change ; I'll close this PR and open a new one with a branch from the current state of `crystals` to make things clearer.

vict0rsch:
> > There currently is a bug here with mismatched dims
> 
> Would you be able say more about this bug?

Yes: this is what we talked about in last week's meeting about how to index elements. The gfn state contains atomic numbers but the proxy was coded to expect an index for its alphabet.

Example: the proxy's data set (matbench) contains 83 different atom types but they are not the first 83 atoms in the periodic table. It could be for instance `H, C, O` in which case it expects `0` to refer to `H`, `1` to refer to `C` etc. To do that it uses an "alphabet" `self.aphabet = [1, 6, 8]` in this example to map things correctly. I think we can do better and internalize that logic such that the proxy takes care of this under the hood in its `forward` and take the actual atomic numbers as inputs.

PR #193, opened 2023-09-06
Backward sampling for continuous torus (ctorus)

alexhernandezgarcia:
The main goal of this PR is to enable backward sampling in the Continuous Torus (`ctorus`) environment. This is needed to sample from the replay buffer and to compute evaluation metrics related to the likelihood of test data.#

Enabling backward sampling in the CTorus has required changing, among other things, the arguments of (formerly called) `sample_actions()` method of the environments. Specifically, it needs to know
- The originating states
- Whether the actions are forward or backward

Therefore, these has required changes wherever this method is used.

I have taken the chance to make other changes in this method:
- Add a good amount of docstrings
- Rename the method to `sample_actions_batch()`
- Other refactoring changes

More things:
- Important fix: `set_state(state)` now copies the state before setting it to `self.state`. This was the source of painful hidden errors.
- Extended tests for the `ctorus`.
- Documentation here and there, especially in `ctorus.py`
- `statebatch2policy()` of the tori calls `statetorch2proxy` instead of doing numpy-based transformations.
- Remove `get_uniform_terminating_states()` of Tetris since it is outdated and it can use the base env's `get_random_terminating_states()`.
- Enable test of backward sampling for all environments through the use of `get_uniform_terminating_states()`. (See important note about this below)

:warning: After having enabled tests of backward sampling for some environments that did not include these tests, they fail. This is the case of:
- Crystal: it gets stuck in no valid actions available, so we need to look into it.
- Tree: `Tried to execute action (2, 0.5769401788711548) not present in action space.` I think this may be fixed in another PR.

### Sanity checks:

[wandb sanity check runs](https://wandb.ai/alexhg/gfn_sanity_checks/reports/CTorus-with-backward-sampling--Vmlldzo1MzI1MjI3)


gflownet/envs/base.py
```diff
@@ -411,18 +411,60 @@ def step_backwards(
...
+    def sample_actions_batch(
         self,
         policy_outputs: TensorType["n_states", "policy_output_dim"],
-        sampling_method: str = "policy",
-        mask_invalid_actions: TensorType["n_states", "policy_output_dim"] = None,
-        temperature_logits: float = 1.0,
-        max_sampling_attempts: int = 10,
+        mask: Optional[TensorType["n_states", "policy_output_dim"]] = None,
+        states_from: Optional[List] = None,
+        is_backward: Optional[bool] = False,
```
michalkoziarski, line 421: I assume this is a known issue since you asked on Slack, but the mix of `is_backward`/`is_forward` that's in the code right now is a little confusing.

gflownet/envs/base.py
```diff
@@ -868,7 +911,7 @@ def set_state(self, state: List, done: Optional[bool] = False):
         at all states (intermediate states are not fully constructed objects) should
         overwrite this method and check for validity.
         """
-        self.state = state
+        self.state = copy(state)
```
michalkoziarski, line 914: 1. Does this actually need to be copied? 2. If so, this won't work, since some envs (Tree) have tensor states.

gflownet/envs/base.py
```diff
@@ -891,18 +934,18 @@ def equal(state_x, state_y):
             return state_x == state_y
 
     @staticmethod
-    def isclose(state_x, state_y):
+    def isclose(state_x, state_y, atol=1e-8):
```
michalkoziarski, line 937: Is there a reason you specify `atol` right now? I tried searching for some usage, but couldn't find any.

gflownet/envs/ctorus.py
```diff
@@ -72,34 +80,56 @@ def get_policy_output(self, params: dict):
 
     def get_mask_invalid_actions_forward(self, state=None, done=None):
         """
-        Returns [True] if the only possible action is eos, [False] otherwise.
+        The action is space is continuous, thus the mask is not of invalid actions as
```
michalkoziarski, line 83:
```suggestion
        The action space is continuous, thus the mask is not of invalid actions as
```
?

gflownet/envs/ctorus.py
```diff
@@ -72,34 +80,56 @@ def get_policy_output(self, params: dict):
...
+        states from which only certain actions are possible.
+
+        The "mask" has 2 elements - to match the mask of backward actions - but only
+        one is needed for forward actions, thus both elements take the same value,
+        according to the following:
+
+        - If done is True, then the mask is True.
+        - If the number of actions (state[-1]) is equal to the (fixed) trajectory
+          length, then only EOS is valid and the mask is True.
+        - Otherwise, any continuous action is valid (except EOS) and the mask is False.
```
michalkoziarski, line 94: I get that this mask is in a way fake, but for consistency wouldn't it make more sense to set the first entry of the mask (corresponding to the generic action) to True, and the second (corresponding to EOS) to False if the number of actions is equal to the trajectory length, and the other way around if any continuous action is valid (except EOS) is valid?

gflownet/envs/ctorus.py
```diff
@@ -143,54 +173,106 @@ def get_parents(
...
+        if is_backward:
+            do_return_to_source = mask[:, 0]
+            if torch.any(do_return_to_source):
+                source_angles = tfloat(
+                    self.source[: self.n_dim], float_type=self.float, device=self.device
+                )
+                states_from_angles = tfloat(
+                    states_from, float_type=self.float, device=self.device
+                )[do_return_to_source, : self.n_dim]
+                actions_return_to_source = states_from_angles - source_angles
```
michalkoziarski, line 273: Just to clarify, because there are multiple possible actions to reach the source (due to `% (2 * np.pi)`), the fact that we specify a particular one here shouldn't be a problem, right? (I don't think it should, but I recognize that's sort of an edge case, so just wondered if you thought about it.)

gflownet/envs/ctorus.py
```diff
@@ -206,36 +288,95 @@ def get_logprobs(
...
-                actions[mask_states_sample]
-            )
+            logprobs[do_sample] = distr_angles.log_prob(actions[do_sample])
         logprobs = torch.sum(logprobs, axis=1)
         return logprobs
 
+    def _step(
+        self,
+        action: Tuple[float],
+        backward: bool,
```
michalkoziarski, line 317: As before, would be awesome to have a single name and stick to it.

gflownet/envs/ctorus.py
```diff
@@ -260,20 +402,74 @@ def step(
...
             self.n_actions += 1
             return self.state, self.eos, True
-        # If action is eos, then it is invalid
+        # If anyway action is EOS, then it is invalid
         elif action == self.eos:
+            # [DEBUG] this point should never be reached. Consider removing this elif.
+            print(
+                "An EOS action has been sampled forward even though it is not allowed"
+            )
+            breakpoint()
```
michalkoziarski, line 417: I would definitely remove the breakpoint, this might cause the code to get stuck (and in particular on the cluster, it might go unnoticed), if anything I'd raise an exception.

gflownet/envs/ctorus.py
```diff
@@ -260,20 +402,74 @@ def step(
...
-                self.state[-1] = self.n_actions
             return self.state, action, True
+        # If done is not True but action is EOS, then it is invalid
+        elif action == self.eos:
+            # [DEBUG] this point should never be reached. Consider removing this elif.
+            print(
+                "An EOS action has been sampled backward even though it is not allowed"
+            )
+            breakpoint()
+            return self.state, action, False
```
michalkoziarski, line 468: As above.

gflownet/envs/ctorus.py
```diff
@@ -260,20 +402,74 @@ def step(
...
+        # Otherwise perform action
+        else:
+            self._step(action, backward=True)
+            return self.state, action, True
+
+    def get_max_traj_length(self):
+        return int(self.length_traj)
```
michalkoziarski, line 475: Minor, but I am never sure, should it be `self.length_traj` or `self.length_traj + 1` (since there is also EOS)? Also, why does this need to be converted to `int`?

gflownet/envs/tree.py
```diff
@@ -548,40 +548,44 @@ def set_state(self, state: List, done: Optional[bool] = False):
             )
         return super().set_state(state, done)
 
-    def sample_actions(
+    def sample_actions_batch(
```
michalkoziarski, line 551: This is wrong, but I fixed it in https://github.com/alexhernandezgarcia/gflownet/pull/194 (tree had a separate `sample_actions_continuous` after latest PR).

gflownet/envs/tree.py
```diff
@@ -599,11 +603,11 @@ def sample_actions(
             distr_threshold = MixtureSameFamily(mix, beta_distr)
         thresholds = distr_threshold.sample()
         # Log probs
-        logprobs[mask_cont] = distr_threshold.log_prob(thresholds)
+        logprobs[is_continuous] = distr_threshold.log_prob(thresholds)
         # Build actions
         actions_cont = [(ActionType.PICK_THRESHOLD, th.item()) for th in thresholds]
         actions = []
-        for is_discrete in mask_discrete:
+        for is_discrete in is_discrete:
```
michalkoziarski, line 610: That's not halal. Renamed in https://github.com/alexhernandezgarcia/gflownet/pull/194 (but feel free to name it something else).

michalkoziarski (changes requested):

gflownet/envs/base.py
alexhernandezgarcia: ABSOLUTELY. This cost me about 4 hours yesterday from 8 pm to midnight :sob:

alexhernandezgarcia: It does work with tensors because this `copy()` is from [here](https://github.com/alexhernandezgarcia/gflownet/blob/backward-sampling-continuous/gflownet/utils/common.py#L235), which catches tensor cases.

michalkoziarski: > It does work with tensors because this copy() is from [here](https://github.com/alexhernandezgarcia/gflownet/blob/backward-sampling-continuous/gflownet/utils/common.py#L235), which catches tensor cases.

Oh, you are right, my bad!

> ABSOLUTELY. This cost me about 4 hours yesterday from 8 pm to midnight

Oof! That's an issue with environments being messed up due to shared states is they happen to be initialized with the same state from a replay buffer/dataset, right?

gflownet/envs/base.py
alexhernandezgarcia: I'm aware and I agree, that's why I asked, but I would call this out of the scope of this rather urgent PR.

gflownet/envs/base.py
alexhernandezgarcia: Because at some point during development (bug investigation) I used it. Then I found an alternative way but thought this could be used at some point, so I decided to leave it. Would you remove it? For example, some envs might need more or less tolerance and could benefit from this.

gflownet/envs/base.py
alexhernandezgarcia: Exactly!

alexhernandezgarcia: And this only became an issue with the estimation of log probabilities because there you do initalise multiple envs (for multiple trajectories) with the very same state. And believe me it failed in the most weird ways...

gflownet/envs/ctorus.py
alexhernandezgarcia: Yes that would make sense actually, but may need changes in `sample_actions_batch` and `get_logprobs`

gflownet/envs/ctorus.py
alexhernandezgarcia: It was removed [here](https://github.com/alexhernandezgarcia/gflownet/pull/193/commits/ffdc043cfad3bbf7ea7ace12e167356a7256bcde). Sorry, I had forgotten to push those changes.

gflownet/envs/ctorus.py
alexhernandezgarcia: As [above](https://github.com/alexhernandezgarcia/gflownet/pull/193/commits/ffdc043cfad3bbf7ea7ace12e167356a7256bcde)

gflownet/envs/ctorus.py
alexhernandezgarcia: Nice catch! I had actually fixed this [here](https://github.com/alexhernandezgarcia/gflownet/pull/193/commits/c31ea39466f1d4bea6780c158c9b50df91a2a48a) but had forgotten to push, sorry.

gflownet/envs/tree.py
alexhernandezgarcia: hahahaha oops :sweat_smile:

gflownet/envs/base.py
michalkoziarski: Oh, absolutely; maybe worth creating an issue.

gflownet/envs/base.py
michalkoziarski: > Would you remove it?

Oh, no, I was just curious how why did you change that (and worried that maybe I missed something when I couldn't find any usage).

gflownet/envs/ctorus.py
michalkoziarski: And why the conversion to `int`?

alexhernandezgarcia: I don't remember tbh.

gflownet/envs/ctorus.py
alexhernandezgarcia: I am making this an issue instead of changing now because things work fine anyway and that implies changing + testing + re-reviewing, etc.

alexhernandezgarcia: Hope you don't mind.

gflownet/envs/ctorus.py
alexhernandezgarcia: Ok, false alarm.

alexhernandezgarcia: It's true that "there are multiple possible actions to reach the source". However, they all are the same action in practice because the VonMises distribution is defined in (-π, π).

Let's consider 1D for simplicity. Say we are in the the last state before the source going backwards, with angle θ, so state is [θ, 1]. There are infinite actions that would take us back to the source, but all have the same likelihood given a VonMises distribution defined by the parameters of the model, so it doesn't matter which one we choose. Let's consider the two actions with |Δθ| <= 2π:

- (-θ, )
- (2π - θ, )

Both have the same likelihood given a distribution. To check this empirically:

```
# VonMises distribution with mean at 90 degrees (0.5π) and concentration 1.
distr = VonMises(torch.tensor([0.5 * torch.pi]), torch.tensor([1.0]))
# θ = 0.5π
theta = torch.tensor(0.5 * torch.pi)
theta_inv = torch.tensor(2 * torch.pi) - theta
torch.isclose(distr.log_prob(-theta), distr.log_prob(theta_inv)) ->  tensor([True])
# θ = 0.12345
theta = torch.tensor(0.12345)
theta_inv = torch.tensor(2 * torch.pi) - theta
torch.isclose(distr.log_prob(-theta), distr.log_prob(theta_inv)) ->  tensor([True])
```

Does that make sense? Any remaining concerns?

alexhernandezgarcia: And actually, as discussed in the continuous GFlowNet paper, what we need to model is distribution over state transitions, from s_t to s_t+1 or vice versa, rather than over actions.

michalkoziarski (approved):

gflownet/envs/ctorus.py
```diff
@@ -143,54 +173,106 @@ def get_parents(
             parents = [state]
             return parents, [action]
 
-    def sample_actions(
+    def action2representative(self, action: Tuple) -> Tuple:
```
AlexandraVolokhova, line 176: if this method returns  `self.representative_action` regardless the `action`, why is it `action2representative` with an argument `action`, not just `get_representative_action` (w/o any arguments)?

gflownet/envs/ctorus.py
```diff
@@ -258,22 +393,64 @@ def step(
...
+
+        valid : bool
+            False, if the action is not allowed for the current state, e.g. stop at the
+            root state
+        """
+        # If done is True, set done to False, increment n_actions and return same state
+        if self.done:
+            assert action == self.eos
+            self.done = False
+            self.n_actions += 1
```
AlexandraVolokhova, line 446: Why `self.n_actions` is incremented in the backward step, not decremented? I thought it should be the same as the last dimension of the state

AlexandraVolokhova (approved):
I didn't get all the details but in overall it looks good to me

gflownet/envs/ctorus.py
alexhernandezgarcia: No, it is not the same. `self.n_actions` counts the number of (valid) actions in a trajectory, regardless of whether it is forward of backward.

PR #199, opened 2023-09-07
Add hyperparam to change crystal envs ordering

carriepl:
This PR adds a new option `do_sg_before_composition` to the Crystal environment to change the order in which the stages are traveled. With `do_sg_before_composition=False`, the default value, the order is : composition -> space group -> lattice parameters. With `do_sg_before_composition=True`, the order is : space group -> composition -> lattice parameters.

The reasoning behind this new option is that some space groups (like 220, 227 and 230) are very highly symmetric, to the point where it would be difficult to randomly generate a composition that would be valid with them. Thus, having an option for picking the space group first (and then enforce spacegroup-related constraints on the composition) might make it easier for the GFN to generate samples from those space groups.

To do : 
- [x] Adapt crystal env as it is right now in the main
- [x] Rebase on top of the backwards sampling branch to get Michal's fix to Crystal's set_state() and adapt that method to work with the new hyperparameter
- [x] Add tests for the new hyperparameter
- [x] Add tests for Crystal.set_state()
- [x] Make sure that when the space group stage ends, it activates the stochiometry checks in the composition stage.

Comments for the reviewers  : 
- Given the potential for this PR to break something, I would tend to set the bar to at least 2 reviewers approving
- Please look at the various tests in test_crystal.py and tell me if you think there are additional tests that you think I should have added/expanded to evaluate the new hyperparameter. 
- I would also like feedback on the way that I handle attempts to set the state of an environment (which uses composition/space-group checks) with a state that shouldn't be reachable with these checks active. I want to know if the approach makes sense to you.

### DIFF
--- a/gflownet/envs/crystals/crystal.py
+++ b/gflownet/envs/crystals/crystal.py
@@ -57,7 +57,9 @@ class Crystal(GFlowNetEnv):
         do_sg_before_composition: bool = False,
         **kwargs,
     ):
-        self.composition_kwargs = composition_kwargs or {}
+        self.composition_kwargs = dict(
+            composition_kwargs or {}, do_spacegroup_check=do_stoichiometry_sg_check
+        )
         self.space_group_kwargs = space_group_kwargs or {}
         self.lattice_parameters_kwargs = lattice_parameters_kwargs or {}
         self.do_stoichiometry_sg_check = do_stoichiometry_sg_check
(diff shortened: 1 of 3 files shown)


carriepl:
Something I just thought of...
The way I handle attempts to set the state of an environment (which uses composition/space-group checks) with a state that shouldn't be reachable with these checks active doesn't always work.

It will work if I try to set a state that is in a stage where the checks apply. But it won't work correctly if, for instance, I try to set a state with incompatible composition and space group if the current stage of that state is Stage.LATTICE_PARAMETERS. To handle those cases, I would need to do a full search backwards to ensure that there is at least one valid path from Crystal.source to the state that I want to set.

carriepl:
Closing since this has been implemented in the continuous environment instead.

PR #204, opened 2023-09-11
Cube working decently fi.na.lly

alexhernandezgarcia:
Finally, a working implementation of the continuous Cube. Alongside the Cube environment, this PR includes multiple new tests common to all environments that should ease the development and testing of new environments in the future. 

- [wandb run with corners proxy](https://wandb.ai/alexhg/cube/runs/2qbhxywf)
- [wandb run with uniform proxy](https://wandb.ai/alexhg/cube/runs/janqh3tv)

I still need to do some cleaning up and tests, but the job's a good'n, it seems. Feedback welcome already.

---

I have merged PR https://github.com/alexhernandezgarcia/gflownet/pull/214 into this branch. The description of https://github.com/alexhernandezgarcia/gflownet/pull/214 is below for reference.

## Cube design

- The source state is _not_ part of the object space any longer. Instead of being a regular state with value 0.0 for all dimensions, it has value -1 for all dimensions.
- I have removed the variable `max_val` and simply hard code it to 1.0. There is no clear reason to have this additional complexity.
- The actions now include an extra dimension to indicate whether the action is from or to source. This removes the need to comparison to [0.0, ...] with `.isclose`.
- Instead of asserting that the state is within bounds, the action is marked as invalid and sampling can continue.

## Code improvement 

- Since actions from source are simpler (Jacobian is one, conversion between relative/absolute is the identity, etc.), I have (hopefully) simplified the code of some methods.
- I include one attribute `epsilon` in the environment to control the boundaries for clamping the inputs to `log_prob`.
- The config distribution parameters are more intuitive:
  - Probabilities instead of logits for the Bernoulli distributions.
  - The actual alpha and beta that will go into the distribution (instead of the inputs to a sigmoid, then multiplied and shifted).

## Test data set

- I have also added an `epsilon` in the generation of the grid and uniform data sets. The tests I have run show that including states with dimensions equal or too close to the boundaries has an impact on the logprobs-based metrics. See [wandb report](https://wandb.ai/alexhg/cube/reports/Impact-of-epsilon-in-the-creation-of-the-test-set--Vmlldzo1NDU1MzE3).

## Fixes

- The variable name `fixed_distribution` in some config files was outdated and has been changed to the current `fixed_distr_params`

## Other

- Before @michalkoziarski complains, I acknowledge that I implemented the method `_get_beta_params_from_mean_variance()` which is eventually not used. But I have chosen to leave it for potential future use. Please excuse me the sacrilege.
- I have conducted a partial analysis of the influence of the hyperparameters and added some preliminary conclusions in this [wandb report](https://wandb.ai/alexhg/cube/reports/Cube-HP-analysis-batches-1-3--Vmlldzo1NDYwMDM5).
- New common test: forward and backward trajectories must be reversible (https://github.com/alexhernandezgarcia/gflownet/pull/214/commits/5b7ce4eab44e4a004e441a479b599fdddb42e803)
  - Note that this test surfaces issues in the Crystal, LatticeParameters and Tree environments and is thus skipped for these environments until they get fixed.


gflownet/envs/cube.py
```diff
@@ -0,0 +1,1317 @@
...
+                min_increments,
+                self.max_val,
+                is_backward=False,
+            )
+            # Get logprobs
+            distr_increments = self._make_increments_distribution(
+                policy_outputs[do_increments]
+            )
+            # Clamp because increments of 0.0 or 1.0 would yield nan
+            logprobs_increments_rel[do_increments] = distr_increments.log_prob(
```
michalkoziarski, line 919: As discussed on Slack, `get_logprobs` (both backward and forward) doesn't really work properly, at the very least with correlation estimation from dataset: in particular, in the highlighted line, `clamp` will mask the fact that some of the `increments_rel` are in fact `inf` (which I assume was not the intention).

gflownet/envs/base.py
```diff
@@ -528,6 +553,20 @@ def get_logprobs(
         logprobs = self.logsoftmax(logits)[ns_range, action_indices]
         return logprobs
 
+    def get_jacobian_diag(
```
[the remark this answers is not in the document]
alexhernandezgarcia: It is not needed actually. I think I added it (back in spring) thinking of a future `ContinuousEnv` (or the like) to explicitly indicate that it may be necessary to compute the Jacobian.

gflownet/envs/htorus.py
```diff
@@ -390,13 +394,14 @@ def sample_actions_batch(
         ]
         return actions, logprobs
 
+    # TODO: deprecated
```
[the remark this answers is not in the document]
alexhernandezgarcia: Is there a standard way to do this?

gflownet/envs/cube.py
```diff
@@ -0,0 +1,1317 @@
...
+from torch.distributions import Bernoulli, Beta, Categorical, MixtureSameFamily, Uniform
+from torchtyping import TensorType
+
+from gflownet.envs.base import GFlowNetEnv
+from gflownet.utils.common import copy, tbool, tfloat
+
+
+class Cube(GFlowNetEnv, ABC):
+    """
+    Continuous (hybrid: discrete and continuous) hyper-cube environment (continuous
```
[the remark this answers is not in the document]
alexhernandezgarcia: Actually, it should just say continuous for simplicity.

gflownet/envs/cube.py
```diff
@@ -0,0 +1,1317 @@
...
+        the special case.
+
+        The forward mask has the following structure:
+
+        - 0 : whether a continuous action is invalid. True if the value at any
+          dimension is larger than 1 - min_incr, or if done is True. False otherwise.
+        - 1 : special case when the state is the source state. False when the state is
+          the source state, True otherwise.
+        - 2 : whether EOS action is invalid. EOS is valid from any state, except the
+          source state or if done is True.
```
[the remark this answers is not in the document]
alexhernandezgarcia: I think it is important to describe the very important case where "any dimension is larger than 1 - min_incr". I am not against an alternative, but to be honest I have spent too much time thinking of and describing these masks and I am not super excited to do it and test it again :sweat_smile: If you wanna take a stab at it please be my guest!

gflownet/envs/cube.py
```diff
@@ -0,0 +1,1317 @@
...
+            return mask
+        # Otherwise, continuous actions are valid
+        mask[0] = False
+        return mask
+
+    def get_parents(
+        self, state: List = None, done: bool = None, action: Tuple[int, float] = None
+    ) -> Tuple[List[List], List[Tuple[int, float]]]:
+        """
+        Defined only because it is required. A ContinuousEnv should be created to avoid
```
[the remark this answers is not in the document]
alexhernandezgarcia: I agree, but is it a priority right now?

gflownet/envs/cube.py
```diff
@@ -0,0 +1,1317 @@
...
+        rng = np.random.default_rng(seed)
+        states = rng.uniform(low=0.0, high=self.max_val, size=(n_states, self.n_dim))
+        return states.tolist()
+
+    #     # TODO: make generic for all environments
+    def sample_from_reward(
+        self, n_samples: int, epsilon=1e-4
+    ) -> TensorType["n_samples", "state_dim"]:
+        """
+        Rejection sampling  with proposal the uniform distribution in
```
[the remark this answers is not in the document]
alexhernandezgarcia: Uhm, I don't think so. Being a bit more verbose it would say: "Rejection sampling where the proposal distribution is the uniform distribution in..."

This is being changed in a separate (WIP) [branch](https://github.com/alexhernandezgarcia/gflownet/tree/sample-from-reward). I will make the docstrings clearer there.

gflownet/envs/cube.py
alexhernandezgarcia: There may be something to fix regarding the source state, but regarding infinite increments, they should never make it to this line since only finite increments will be included in the filter `do_increments`. In fact, if we passed infinte increments to the Beta distribution it would complain badly.

config/env/ccube.yaml
```diff
@@ -0,0 +1,38 @@
...
+beta_params_min: 0.01
+beta_params_max: 1000.0
+min_incr: 0.1
+n_comp: 1
+fixed_distribution:
+  beta_weights: 1.0
+  beta_alpha: 0.01
+  beta_beta: 0.01
+  bernoulli_source_logit: 0.0
+  bernoulli_eos_logit: 0.0
```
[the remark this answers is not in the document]
alexhernandezgarcia: Oh, I realise now what you mean! You are thinking that this means probability zero, right? This is actually a logit, not a probability!

config/env/ccube.yaml
```diff
@@ -0,0 +1,38 @@
...
+random_distribution:
+  beta_weights: 1.0
+  # IMPORTANT: adjust because of sigmoid!
+  beta_alpha: 0.01
+  beta_beta: $beta_params_max
+  bernoulli_source_logit: 0.0
+  bernoulli_eos_logit: 0.0
```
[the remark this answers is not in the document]
alexhernandezgarcia: See https://github.com/alexhernandezgarcia/gflownet/pull/204/files#r1329322567

gflownet/envs/cube.py
michalkoziarski: Infs are being replaced by numbers due to clamping. But try inserting
```
if torch.any(torch.isinf(increments_rel)):
    import pdb; pdb.set_trace()
```
in the above line (just before clamping) and try running `+experiments=ccube/ccube_owl proxy=corners` (the same should be happening with backward).

alexhernandezgarcia: Oh oops, OK, I'll look into this.

gflownet/envs/htorus.py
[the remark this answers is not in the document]
michalkoziarski: Technically there is `DeprecationWarning`, but it seems to be ignored by default - I'd just do standard warning.

gflownet/envs/base.py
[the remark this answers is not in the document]
michalkoziarski: So perhaps just remove it for now?

config/env/ccube.yaml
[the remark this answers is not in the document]
michalkoziarski: Derp, my bad.

gflownet/envs/cube.py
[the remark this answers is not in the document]
michalkoziarski: Hell no, it's just a way for me to acknowledge that I also think it would be a good idea to have one. One day. When we retire, start living on a farm, and won't have any deadlines.

gflownet/envs/cube.py
alexhernandezgarcia: Ok, this is related to the source states. The environment does not allow EOS from the source, but the test set contains a (terminating) state which is the source... So weird things happen. I guess I will finally have to face this issue seriously.

gflownet/envs/base.py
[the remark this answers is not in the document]
alexhernandezgarcia: Ok I'll remove it, at the risk of forgetting to mention it later.

gflownet/envs/cube.py
```diff
@@ -0,0 +1,1317 @@
...
+import torch
+from sklearn.neighbors import KernelDensity
+from torch.distributions import Bernoulli, Beta, Categorical, MixtureSameFamily, Uniform
+from torchtyping import TensorType
+
+from gflownet.envs.base import GFlowNetEnv
+from gflownet.utils.common import copy, tbool, tfloat
+
+
+class Cube(GFlowNetEnv, ABC):
```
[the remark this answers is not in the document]
alexhernandezgarcia: Initially there was a version that only increments one dimension at a time. It would be good to have eventually. I have removed it for now. I'm good with the suggestion of `AbstractCube`.

gflownet/envs/cube.py
```diff
@@ -0,0 +1,1317 @@
...
+    """
+
+    def __init__(
+        self,
+        n_dim: int = 2,
+        max_val: float = 1.0,
+        min_incr: float = 0.1,
+        n_comp: int = 1,
+        beta_params_min: float = 0.1,
+        beta_params_max: float = 1000.0,
```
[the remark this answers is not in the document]
alexhernandezgarcia: I have tested it and it _is_ important but I haven't got to test it systematically yet. It's actually one of the reasons why the PR is still "Draft".

gflownet/envs/cube.py
```diff
@@ -0,0 +1,1317 @@
...
+        random policy. The environment consists of both continuous and discrete
+        actions.
+
+        Continuous actions
+
+        For each dimension d of the hyper-cube and component c of the mixture, the
+        output of the policy should return
+          1) the weight of the component in the mixture
+          2) the logit(alpha) parameter of the Beta distribution to sample the increment
+          3) the logit(beta) parameter of the Beta distribution to sample the increment
```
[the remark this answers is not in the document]
alexhernandezgarcia: Wow that's a picky one! I thought I was the only one making such suggestions haha

tests/gflownet/envs/common.py
```diff
@@ -359,6 +359,8 @@ def test__forward_actions_have_nonzero_backward_prob(env):
 
 def test__backward_actions_have_nonzero_forward_prob(env, n=1000):
     states = _get_terminating_states(env, n)
+    if states is None:
+        return
```
alexhernandezgarcia, line 363: I am surprised states can be None since `get_random_terminating_states()` is now implemented in the base env and hence is available to all states :confused:.

carriepl: I can't really comment on that. I just saw that it was None for the crystal env and that every other test had a mechanism for dealing with Nones except this one.
I guess this means that the true underlying issue isn't fixed and that other tests (which already had the "`if states is None`") might also have issues.

alexhernandezgarcia: Ok, thanks, I will take a closer look at this.

alexhernandezgarcia: I have kept the check-and-return-if-None but added warnings so that this doesn't go silent at least: https://github.com/alexhernandezgarcia/gflownet/pull/204/commits/993f4568fd8c466244d873de2b7133e9b21a6347

gflownet/envs/htorus.py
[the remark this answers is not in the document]
alexhernandezgarcia: I will not do this yet because actually the continuous torus inherits from the hybrid one, so things could become messier. We'll cleanup after deadlines.

alexhernandezgarcia (commented):
I cannot approve my own PR but I have done a re-review of the changes after merging https://github.com/alexhernandezgarcia/gflownet/pull/214 and things look alright to me.

carriepl:
I'll do my best to review but I'll be honest; there's no way for me to be anywhere near a decent job when reviewing a PR of this size. I know that some of that has previously been reviewed in other sub-PRs but I can't really know what since that's not indicated in the code.

config/experiments/ccube/hyperparams_search_20230920_batch1.yaml
```diff
@@ -0,0 +1,145 @@
+# Shared config
```
michalkoziarski, line 1: I don't mind much either way, but I wonder if it's actually important for all of these hyperparam configs to be merged.

gflownet/envs/cube.py
michalkoziarski: Did you end up addressing this?

gflownet/envs/cube.py
```diff
@@ -0,0 +1,1433 @@
...
+import torch
+from sklearn.neighbors import KernelDensity
+from torch.distributions import Bernoulli, Beta, Categorical, MixtureSameFamily
+from torchtyping import TensorType
+
+from gflownet.envs.base import GFlowNetEnv
+from gflownet.utils.common import copy, tbool, tfloat
+
+
+class CubeBase(GFlowNetEnv, ABC):
```
michalkoziarski, line 20: As a general rule I would advise against creating classes like this before they are actually required (subclassed by more than one class) - though I am not suggesting removing it at this point.

michalkoziarski (commented):
I didn't review it in much detail again (I did it before), but from the high level looks good to me.

config/experiments/ccube/hyperparams_search_20230920_batch1.yaml
alexhernandezgarcia: This could even be removed from the repo, so not important.

gflownet/envs/cube.py
alexhernandezgarcia: I agree. In my defense, when I created it there where 2 subclasses. Then I ended up removing one for now.

alexhernandezgarcia:
> I'll do my best to review but I'll be honest; there's no way for me to be anywhere near a decent job when reviewing a PR of this size. I know that some of that has previously been reviewed in other sub-PRs but I can't really know what since that's not indicated in the code.

I totally understand. You may also choose to review nothing and it's fine.

gflownet/proxy/corners.py
```diff
@@ -30,6 +30,15 @@ def setup(self, env=None):
...
             self.mulnormal_norm = 1.0 / ((2 * torch.pi) ** 2 * cov_det) ** 0.5
 
+    @property
+    def min(self):
+        if not hasattr(self, "_min"):
+            mode = self.mu * torch.ones(
+                self.n_dim, device=self.device, dtype=self.float
+            )
+            self._min = self(torch.unsqueeze(mode, 0))[0]
+        return self._min
```
carriepl, line 40: So if the caller of that property (accidently) does anything inplace on the returned tensor, that change will affect everyone else calling min in the future?
If so, it feels like it might lead to bugs that would be really difficult to track.
Same comment applies to the other proxies that implement the same strategy : torus (for min and norm) and uniform (for min).

tests/gflownet/envs/common.py
```diff
@@ -76,34 +120,94 @@ def test__get_parents_step_get_mask__are_compatible(env):
             assert mask[env.action_space.index(p_a)] is False
 
 
+@pytest.mark.repeat(500)
```
carriepl, line 123: This doesn't really do anything in this context.

Using this decorator will repeat the test 500 times but only when executed directly by pytest. When running these common test through `test__all_env_common()`, for instance, that decorator is ignored so the tests only run once each.

And since, to my knowledge, they are only ever invoked through `test__all_env_common()` or the continuous equivalent, that means that these decorators don't do anything.

alexhernandezgarcia: I wasn't aware of this. I think it is important to repeat some of these tests. Do you have any suggestions about how to better organise these common tests? Also, not a big deal, but because they are all part of `test__all_env_common()` they are only counted once.

gflownet/proxy/corners.py
alexhernandezgarcia: Uhm I see. Do you have a suggestion of a better way of doing this?

gflownet/envs/base.py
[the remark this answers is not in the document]
alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/204/commits/9cd47a6e186f79002d66f00463e231c26b89c65a

gflownet/envs/cube.py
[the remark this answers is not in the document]
alexhernandezgarcia: https://wandb.ai/alexhg/cube/reports/Cube-HP-analysis-batches-1-3--Vmlldzo1NDYwMDM5

gflownet/envs/cube.py
alexhernandezgarcia: Yes, it was a bunch of changes that I made after this so hard to summarise exactly what it was. But I've just tried again with the breakpoint and it does not stop.

config/experiments/ccube/hyperparams_search_20230920_batch1.yaml
alexhernandezgarcia: I will remove them and keep them elsewhere ;)

tests/gflownet/envs/common.py
alexhernandezgarcia: I will make this an issue so that it does not lock the merge, since it is not related to the PR but more general.

gflownet/proxy/corners.py
alexhernandezgarcia: Let's keep the conversation on this issue, please: https://github.com/alexhernandezgarcia/gflownet/issues/243

tests/gflownet/envs/common.py
alexhernandezgarcia: Let's keep the conversation on the issue thread please: https://github.com/alexhernandezgarcia/gflownet/issues/242

alexhernandezgarcia:
Thanks for the reviews! So as to move on, I have moved two hanging issues to the issues section, since they are important but not very specific to the PR. I will merge!

michalkoziarski (approved):

PR #205, opened 2023-09-12
Shared policy config

michalkoziarski:
Added option to specify `policy.shared` config (in addition to `policy.forward` and `policy.backward`), which will be used by both types of policies. Required for running sweeps. Tested on grid environment in https://wandb.ai/michalkoziarski/gfn_sanity_checks.


alexhernandezgarcia (approved):
Oh yeah this looks much nicer now thanks! :)

PR #207, opened 2023-09-13
Environment to sample Miller indices

alexhernandezgarcia:
It's extremely simple given the already existing code since it can be modelled as a hypercube.

### DIFF
--- a/gflownet/envs/crystals/miller.py
+++ b/gflownet/envs/crystals/miller.py
@@ -1,16 +1,6 @@
 """
 Class to represent an environment to sample Miller indices (hkl).
 """
-import itertools
-from typing import List, Optional, Tuple
-
-import matplotlib.pyplot as plt
-import numpy as np
-import numpy.typing as npt
-import torch
-from mpl_toolkits.axes_grid1 import make_axes_locatable
-from torchtyping import TensorType
-
 from gflownet.envs.grid import Grid
 
 
--- a/gflownet/envs/crystals/miller.py
+++ b/gflownet/envs/crystals/miller.py
@@ -8,7 +8,7 @@ class MillerIndices(Grid):
     """
     The Miller indices are represented by either 3 parameters (h, k, l) if the
     structure is cubic, or by 4 parameters (h, k, i, l) if the structure is hexagonal
-    or rhobohedral. Each parameter can take values in the set {-2, -1, 0, 1, 2}.
+    or rhombohedral. Each parameter can take values in the set {-2, -1, 0, 1, 2}.
     Therefore, we can represent the Miller indices environment by a hyper cube of
     length 5, with dimensionality 3 or 4 depending on the structure.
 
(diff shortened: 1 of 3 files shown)


AlexDuvalinho (approved):
Looks good

gflownet/envs/crystals/miller.py
```diff
@@ -0,0 +1,40 @@
...
+    structure is cubic, or by 4 parameters (h, k, i, l) if the structure is hexagonal
+    or rhombohedral. Each parameter can take values in the set {-2, -1, 0, 1, 2}.
+    Therefore, we can represent the Miller indices environment by a hyper cube of
+    length 5, with dimensionality 3 or 4 depending on the structure.
+
+    Attributes
+    ----------
+    is_cubic : bool
+        True if the structure is cubic, False if the structure is hexagonal or
+        rhombohedral.
```
carriepl, line 19: What if the structure is neither cubic, hexagonal or rhombohedral?
As far as I can tell, those remaining cases would also miller indices specified by 3 parameters like the cubic

gflownet/envs/crystals/miller.py
```diff
@@ -0,0 +1,40 @@
...
+        self,
+        is_cubic: bool,
+        max_increment: int = 1,
+        max_dim_per_action: int = 1,
+        **kwargs,
+    ):
+        if is_cubic:
+            n_dim = 3
+        else:
+            n_dim = 4
```
carriepl, line 39: I would suggest to actually always use `n_dim=3`. My reasoning : 

In the hexagonal/rhombohedral scenario, there are actually a constraint on the values of the Miller indices since there is some redundancy between them. The constraint is `h + k + i = 0.` 
Source : https://dictionary.iucr.org/Miller_indices#Bravais-Miller_indices_.28hexagonal_axes.29

That means that there are only ever 3 dimensions that we need to pick : 
- `h, k, i` for non-{hexagonal, rhombohedral}
- `h, k, l` for {hexagonal, rhombohedral}

In the hexagonal/rhombohedral scenario, we can simply infer the value of `i` from the values of `h` and `k`. The only thing that we need to do is ensure that `h` and `k` don't take values that makes it impossible to find a valid value of `i`. For that, I think we only need to ensure that `-2 <= h+k <=2`.

carriepl: If we do this change, we might need to add tests to make sure that we don't allow the picking of, say, `h=2 and k=2` for a rhombohedral structure.

alexhernandezgarcia: You are right, I had overlooked all this :( Then we need to make several changes here.

gflownet/envs/crystals/miller.py
alexhernandezgarcia: Yes, right. I guess by cubic I meant non-hexagonal/non-rhombohedral but it is not a good choice for the variable name.

carriepl (approved):
Looks good to me. If the CI is happy, so am I.

PR #209, opened 2023-09-16
[WIP] Continuous lattice parameters environment

alexhernandezgarcia:
Implementation of the a lattice parameters environment with continuous parameters as a (continuous) hypercube.

The implementation relies entirely on the hypercube and does not require any additional obscure modifications to deal with the constraints imposed by the lattice system. Simply the number of "effective dimensions" is obtained depending on the lattice system and is used as the (effective) dimensionality of the hypercube.

[Test runs](https://wandb.ai/alexhg/latticeparams)

---

To do list:

- [x] Adapt both the Cube and the CLatticeParameters environment so that it works within the Crystal: modify `sample_actions` and `get_logprobs` so that only the "effective dimensions" are taken into account.
- [x] Decide whether it is necessary to set "not equality" constraints of the kind `angle != 90` (in order to prevent one lattice system becoming another lattice system) or whether it is fine as is because continuous actions will never in theory and very rarely in practice yield the equality. *We have decided that this is not necessary because it will be extremely rare if possible at all and it should not cause important issues anyway.*
- [ ] Implement `state2policy` (and the like) such that angles are encoded with sinusoidal embeddings.
- [ ] Implement `state2proxy` (and the like).
- [ ] Design testing proxies to evaluate the environment. 
- [ ] Implement an elegant way of integrating the discrete and continuous environments, since some methods are common.

### DIFF
--- a/config/env/crystals/clattice_parameters.yaml
+++ b/config/env/crystals/clattice_parameters.yaml
@@ -1,0 +1,43 @@
+defaults:
+  - base
+
+_target_: gflownet.envs.crystals.clattice_parameters.CLatticeParameters
+
+id: clattice_parameters
+continuous: True
+
+# Lattice system
+lattice_system: triclinic
+# Allowed ranges of size and angles
+min_length: 1.0
+max_length: 5.0
+min_angle: 30.0
+max_angle: 150.0
+
+# Policy
+beta_params_min: 0.01
+beta_params_max: 1000.0
+min_incr: 0.1
+n_comp: 1
+fixed_distribution:
+  beta_weights: 1.0
+  beta_alpha: 0.01
+  beta_beta: 0.01
+  bernoulli_source_logit: 0.0
+  bernoulli_eos_logit: 0.0
+random_distribution:
+  beta_weights: 1.0
+  # IMPORTANT: adjust because of sigmoid!
+  beta_alpha: 0.01
+  beta_beta: $beta_params_max
+  bernoulli_source_logit: 0.0
+  bernoulli_eos_logit: 0.0
+# Buffer
+buffer:
+  data_path: null
+  train: null
+  test:
+    type: grid
+    n: 1000
+    output_csv: ccube_test.csv
+    output_pkl: ccube_test.pkl
--- a/config/env/crystals/clattice_parameters.yaml
+++ b/config/env/crystals/clattice_parameters.yaml
@@ -4,7 +4,6 @@ defaults:
 _target_: gflownet.envs.crystals.clattice_parameters.CLatticeParameters
 
 id: clattice_parameters
-continuous: True
 
 # Lattice system
 lattice_system: triclinic
--- a/gflownet/envs/crystals/clattice_parameters.py
+++ b/gflownet/envs/crystals/clattice_parameters.py
@@ -22,6 +22,10 @@ from gflownet.utils.crystals.constants import (
     TRICLINIC,
 )
 
+LENGTHS = ("a", "b", "c")
+ANGLES = ("alpha", "beta", "gamma")
+PARAMETERS = LENGTHS + ANGLES
+
 
 # TODO: figure out a way to inherit the (discrete) LatticeParameters env or create a
 # common class for both discrete and continous with the common methods.
@@ -35,9 +39,9 @@ class CLatticeParameters(ContinuousCube):
     from the (continuous) cube environment, creating a mapping between cell position
     and edge length or angle, and imposing lattice system constraints on their values.
 
-    Similar to the Cube environment, the values are initialized with zeros
-    (or target angles, if they are predetermined by the lattice system), and are
-    incremented by sampling from a (mixture of) Beta distribution(s).
+    The environment is a hyper cube of dimensionality 6 (the number of lattice
+    parameters), but it takes advantage of the mask of ignored dimensions implemented
+    in the Cube environment.
 
     The values of the state will remain in the default [0, 1] range of the Cube, but
     they are mapped to [min_length, max_length] in the case of the lengths and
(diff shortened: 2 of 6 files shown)


gflownet/envs/crystals/clattice_parameters.py
```diff
@@ -0,0 +1,242 @@
...
+            Maximum value of the lengths.
+
+        min_angle : float
+            Minimum value of the angles.
+
+        max_angle : float
+            Maximum value of the angles.
+        """
+        self.lengths = ("a", "b", "c")
+        self.angles = ("alpha", "beta", "gamma")
```
michalkoziarski, line 75: This is slightly confusing variable name, perhaps `{length, angle}_parameters`? (I would actually personally do `self.{length, angle}_parameter_names` and `self.parameter_names` in the next line.)

gflownet/envs/crystals/clattice_parameters.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def _setup_constraints(self):
+        """
+        Computes the effective number of dimensions, given the constraints imposed by
+        the lattice system.
+
+        Returns
+        -------
+        n_dim : int
+            The number of effective dimensions that can be be udpated in the
```
michalkoziarski, line 140:
```suggestion
            The number of effective dimensions that can be be updated in the
```

michalkoziarski (commented):
Minor comments about what's here so far.

michalkoziarski:
Regarding
> Decide whether it is necessary to set "not equality" constraints,
my take right now would be that it's not really necessary in practice, and would potentially be quit a bit of work to implement.

config/env/crystals/clattice_parameters.yaml
```diff
@@ -0,0 +1,43 @@
+defaults:
+  - base
+
+_target_: gflownet.envs.crystals.clattice_parameters.CLatticeParameters
+
+id: clattice_parameters
+continuous: True
```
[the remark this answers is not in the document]
alexhernandezgarcia: Yes, that's right, thanks.

config/env/crystals/clattice_parameters.yaml
```diff
@@ -0,0 +1,43 @@
...
+beta_params_min: 0.01
+beta_params_max: 1000.0
+min_incr: 0.1
+n_comp: 1
+fixed_distribution:
+  beta_weights: 1.0
+  beta_alpha: 0.01
+  beta_beta: 0.01
+  bernoulli_source_logit: 0.0
+  bernoulli_eos_logit: 0.0
```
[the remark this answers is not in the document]
alexhernandezgarcia: To be decided because it is actually a hyperparameter and it can a big influence on the training.

gflownet/envs/crystals/clattice_parameters.py
```diff
@@ -0,0 +1,494 @@
...
+        values = [float(value) for value in values]
+
+        for param, value in zip(PARAMETERS, values):
+            state = self._set_param(state, param, value)
+        return state
+
+
+# TODO: figure out a way to inherit the (discrete) LatticeParameters env or create a
+# common class for both discrete and continous with the common methods.
+class CLatticeParametersEffectiveDim(ContinuousCube):
```
carriepl, line 273: Is there a reason to have both classes?
I get that the difference is that an instance of the CLatticePArameters class could sample LPs for any lattice system whereas, in this one, it is fixed per instance. Because of that, I get why CLatticeParameters is better suited for use in the Crystal env than the CLatticeParametersEffectiveDim. 
But in that case, I'm not sure why we also have the CLatticeParametersEffectiveDim class.

And if we do need the two different classes, given that they share a fair bit of code, I'm wondering if they should have a different inheritance structure to reduce code duplication (either inherit from some common class that would be a bit more specific than the cube, or one inherits from the other, ...)

gflownet/envs/cube.py
```diff
@@ -593,6 +616,34 @@ def _make_increments_distribution(
...
+
+        tensor_to_mask : tensor
+            Tensor to be modified. It may be a tensor of actions, of logprobs of
+            increments or the log of the diagonal of the Jacobian.
+        """
+        is_ignored_dim = mask[:, -self.n_dim :]
+        if torch.any(is_ignored_dim):
+            shape_orig = tensor_to_mask.shape
+            tensor_to_mask[is_ignored_dim] = 0.0
+            tensor_to_mask = tensor_to_mask.reshape(shape_orig)
```
carriepl, line 644: I'm not sure I understand the purpose of this line. Do it mean that the previous line (L643) changes the shape of `tensor_to_mask`?

gflownet/envs/cube.py
```diff
@@ -753,6 +808,8 @@ def _sample_actions_batch_backward(
         # Initialize variables
         n_states = policy_outputs.shape[0]
         is_bts = torch.zeros(n_states, dtype=torch.bool, device=self.device)
+        # Mask of effective dimensions
+        is_effective_dim = ~mask[:, -self.n_dim :]
```
carriepl, line 812: This variable is created here but I don't see it used anywhere in this function.

gflownet/envs/cube.py
```diff
@@ -869,10 +934,12 @@ def _get_logprobs_forward(
         logprobs_increments_rel = torch.zeros(
             (n_states, self.n_dim), dtype=self.float, device=self.device
         )
-        jacobian_diag = torch.ones(
+        log_jacobian_diag = torch.zeros(
             (n_states, self.n_dim), device=self.device, dtype=self.float
         )
         eos_tensor = tfloat(self.eos, float_type=self.float, device=self.device)
+        # Mask of effective dimensions
+        is_effective_dim = ~mask[:, -self.n_dim :]
```
carriepl, line 942: Same here for this variable which seems unused.

gflownet/envs/cube.py
```diff
@@ -953,9 +1028,11 @@ def _get_logprobs_backward(
         logprobs_increments_rel = torch.zeros(
             (n_states, self.n_dim), dtype=self.float, device=self.device
         )
-        jacobian_diag = torch.ones(
+        log_jacobian_diag = torch.zeros(
             (n_states, self.n_dim), device=self.device, dtype=self.float
         )
+        # Mask of effective dimensions
+        is_effective_dim = ~mask[:, -self.n_dim :]
```
carriepl, line 1035: This variable too seems unused.

tests/gflownet/envs/test_ccube.py
```diff
@@ -196,43 +196,43 @@ def test__mask_backward__1d__returns_expected(cube1d, state, mask_expected):
...
         ),
         (
             [0.5, 0.95],
-            [False, True, True],
+            [False, True, True, False, False],
         ),
         (
             [0.95, 0.95],
-            [False, True, True],
+            [False, True, True, False, False],
```
carriepl, line 235: Should there be tests added with couple dimensions? Because, ATM, the tests all seem to expect the additional 'n_dim' dimensions to be False ?

gflownet/envs/crystals/clattice_parameters.py
alexhernandezgarcia: This? https://github.com/alexhernandezgarcia/gflownet/pull/209/commits/2e9a1b3127152d544c8ca6e407e4bb77ab46f639

tests/gflownet/envs/test_clattice_parameters.py
```diff
@@ -0,0 +1,216 @@
...
+from gflownet.envs.crystals.clattice_parameters import (
+    CUBIC,
+    HEXAGONAL,
+    LATTICE_SYSTEMS,
+    MONOCLINIC,
+    ORTHORHOMBIC,
+    RHOMBOHEDRAL,
+    TETRAGONAL,
+    TRICLINIC,
+    CLatticeParameters,
```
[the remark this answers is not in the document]
alexhernandezgarcia: Sorry that this isn't clear. `CLatticeParametersEffectiveDims` is my initial version of the environment, but then I realised it would not be possible to integrate it with the general crystal. I changed the name and left it there for the time being, but I did change the tests to the new version.

gflownet/envs/crystals/clattice_parameters.py
alexhernandezgarcia: I don't think we need CLatticeParametersEffectiveDim. I only left it while developing the new version. Sorry for the mess.

tests/gflownet/envs/test_clattice_parameters.py
```diff
@@ -0,0 +1,216 @@
...
+@pytest.mark.repeat(N_REPETITIONS)
+def test__triclinic__constraints_remain_after_random_actions(env, lattice_system):
+    env = env.reset()
+    while not env.done:
+        env.step_random()
+        (a, b, c), (alpha, beta, gamma) = env._unpack_lengths_angles()
+        assert a != b
+        assert a != c
+        assert b != c
+        assert len({alpha, beta, gamma, 90.0}) == 4
```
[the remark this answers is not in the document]
alexhernandezgarcia: Good point. I have changed things here: https://github.com/alexhernandezgarcia/gflownet/pull/209/commits/ceb5788c3adcb2b6b75eaadc4ca073ecf5850980

tests/gflownet/envs/test_clattice_parameters.py
```diff
@@ -0,0 +1,216 @@
...
+        (CUBIC, "(1.0, 1.0, 1.0), (90.0, 90.0, 90.0)"),
+        (HEXAGONAL, "(1.0, 1.0, 1.0), (90.0, 90.0, 120.0)"),
+        (MONOCLINIC, "(1.0, 1.0, 1.0), (90.0, 30.0, 90.0)"),
+        (ORTHORHOMBIC, "(1.0, 1.0, 1.0), (90.0, 90.0, 90.0)"),
+        (RHOMBOHEDRAL, "(1.0, 1.0, 1.0), (30.0, 30.0, 30.0)"),
+        (TETRAGONAL, "(1.0, 1.0, 1.0), (90.0, 90.0, 90.0)"),
+        (TRICLINIC, "(1.0, 1.0, 1.0), (30.0, 30.0, 30.0)"),
+    ],
+)
+def test__readable2state__returns_initial_state_for_rhombohedral_and_triclinic(
```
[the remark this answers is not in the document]
alexhernandezgarcia: Thanks, fixed in https://github.com/alexhernandezgarcia/gflownet/pull/209/commits/3ee57b12c87a8b71ab355c9134e7d1e691173ce5

gflownet/envs/cube.py
alexhernandezgarcia: Yes, it does. It flattens the tensor, annoyingly enough. Is there a more elegant way of handling this?

alexhernandezgarcia: For context: both `is_ignored_dim` and `tensor_to_mask` are 2D tensors of the same shape. `is_ignored_dim` is boolean.

gflownet/envs/cube.py
alexhernandezgarcia: Thanks! Fixed in https://github.com/alexhernandezgarcia/gflownet/pull/209/commits/d6303ec7490aa35e74213adfb7346f4b315b39ea

gflownet/envs/cube.py
alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/209/commits/d6303ec7490aa35e74213adfb7346f4b315b39ea

gflownet/envs/cube.py
alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/209/commits/d6303ec7490aa35e74213adfb7346f4b315b39ea

tests/gflownet/envs/test_ccube.py
alexhernandezgarcia: Ideally yes.

alexhernandezgarcia:
I will close this PR, since the up-to-date changes are in https://github.com/alexhernandezgarcia/gflownet/pull/222.

PR #211, opened 2023-09-18
Lattice dummy proxy

michalkoziarski:
Dummy lattice parameter proxy that uses KDE fitted on a dataset of triclinic materials, with predicted scores clipped and rescaled to be negative numbers.

Histograms of the (original, not clipped/rescaled) scores for the train dataset and a grid of all possible environment states:

Figure_1
Figure_2

Current run: https://wandb.ai/michalkoziarski/GFlowNet/runs/t4ln9azm.

EDIT: correlation seems fine in the latest run! Marking it ready for review.


config/env/crystals/lattice_parameters.yaml
```diff
@@ -0,0 +1,28 @@
+defaults:
+  - base
+
+_target_: gflownet.envs.crystals.lattice_parameters.LatticeParameters
+
+id: lattice_parameters
+
+lattice_system: triclinic
+min_length: 1.0
+max_length: 5.0
```
alexhernandezgarcia, line 10: Could you please remind me if these default values were based on something? On a data set?

data/crystals/lattice_proxy.pickle
alexhernandezgarcia, line 1: Picky suggestion: Could we rename the extension `*.pkl`, just to make it coherent with other pickled files in the repo?

gflownet/proxy/crystals/lattice_parameters.py
```diff
@@ -0,0 +1,32 @@
...
+
+        self.min_value = min_value
+        self.kde = None
+
+    def __call__(self, states: List) -> Tensor:
+        scores = self.kde.score_samples(states)
+        scores = np.clip(scores, self.min_value, np.inf)
+        scores = self.min_value - scores
+
+        return Tensor(scores)
```
alexhernandezgarcia, line 25: To follow the standard in the repo:
```suggestion
        return tfloat(scores, float_type=self.float, device=self.device)
```

gflownet/proxy/crystals/lattice_parameters.py
```diff
@@ -0,0 +1,32 @@
+import pickle
+from pathlib import Path
+from typing import List
+
+import numpy as np
+from torch import Tensor
+
+from gflownet.proxy.base import Proxy
```
alexhernandezgarcia, line 8: To make the other suggestion work:
```suggestion
from gflownet.proxy.base import Proxy
from gflownet.utils.common import tfloat
```

gflownet/proxy/crystals/lattice_parameters.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self, min_value: float = -100, **kwargs):
+        super().__init__(**kwargs)
+
+        self.min_value = min_value
+        self.kde = None
+
+    def __call__(self, states: List) -> Tensor:
+        scores = self.kde.score_samples(states)
+        scores = np.clip(scores, self.min_value, np.inf)
+        scores = self.min_value - scores
```
alexhernandezgarcia, line 23: Does this work fine? The convention for the proxies is that they are in a range [-MIN, 0).

config/env/crystals/lattice_parameters.yaml
michalkoziarski: That's a good question... I seem to recall asking Divya at some point for the stats from dataset, but I just compared this against the triclinic CSV used in this PR, and it doesn't fit within this range.

I suspect it has to do with the file being generated prior to migration to conventional, but we are trying to figure this out right now.

data/crystals/lattice_proxy.pickle
michalkoziarski: Done.

gflownet/proxy/crystals/lattice_parameters.py
michalkoziarski: Well, it should. Original scores should be in `(-inf, 0)` (in practice, for the discrete grid states, that will be `~(-700, 0)`). Clipping will reduce that to `[-100, 0)`, so `self.min_value - scores = -100 - scores` will still be in `(-100, 0]` range, just in the reversed order. But let me know if you see a problem with this.

config/env/crystals/lattice_parameters.yaml
alexhernandezgarcia: Have you converged into something about this question?

gflownet/proxy/crystals/lattice_parameters.py
alexhernandezgarcia: Ok got you thanks!

scripts/fit_lattice_proxy.py
```diff
@@ -0,0 +1,53 @@
+"""
+Script for fitting KDE on a dataset of bond lengths and angles of materials with
```
alexhernandezgarcia:
```suggestion
Script for fitting KDE on a dataset of unit cell lengths and angles of materials with
```

config/env/crystals/lattice_parameters.yaml
michalkoziarski: Yes, https://github.com/alexhernandezgarcia/gflownet/pull/217, discussion on Slack. KDE needs to be refitted.

michalkoziarski:
This was refitted on Matbench, histograms after refitting:
Figure_1
Figure_2

alexhernandezgarcia (approved):
Looks good, will merge

PR #212, opened 2023-09-18
Torsion angles detection fix

AlexandraVolokhova:
Rewritten the function for automatically finding rotatable torsion angles in a molecule. In the new implementation double bonds are considered non-rotatable (as they should be according to chemical rules). Also, torsion angles tuples (atom1, atom2, atom3, atom4), are structured such that
- atom2 < atom3
- atom1 and atom4 are minimal among the neighbours of atom2 and atom3 correspondingly


michalkoziarski (approved):
In general approved, thanks for spotting this! I made some minor changes in https://github.com/alexhernandezgarcia/gflownet/pull/213 (no logic changes), if you are ok with that just merge it into this, and later merge this into `molecule_graph_env`.

Two clarifying questions though: 1) is there a particular reason to enforce the ordering you are enforcing? 2) do you know why in the original implementation they actually used double bonds, and if it was actually only for the C=C (as they commented), or all of the possible double bonds?

AlexandraVolokhova:
> 1) is there a particular reason to enforce the ordering you are enforcing?

It is just for clarity and for avoiding ambiguity, so for every molecule we have a clear deterministic way to define the torsion angles. But no chemical reason to do it specifically this way

>  2) do you know why in the original implementation they actually used double bonds, and if it was actually only for the C=C (as they commented), or all of the possible double bonds?

I don't know what the authors of the original implementation had in mind, maybe they considered together cis-trans isomers ("conformers"  which are different in the positions of the parts of the molecule wrt double bond) and conformers, but in our case the cis-trans isomerism is clearly defined in the input SMILES string and should not be changed (i.e. double bond are not rotatable)

PR #214, opened 2023-09-20
Changes - hopefully improvements - on the design of the Cube environment

alexhernandezgarcia:
## Cube design

- The source state is _not_ part of the object space any longer. Instead of being a regular state with value 0.0 for all dimensions, it has value -1 for all dimensions.
- I have removed the variable `max_val` and simply hard code it to 1.0. There is no clear reason to have this additional complexity.
- The actions now include an extra dimension to indicate whether the action is from or to source. This removes the need to comparison to [0.0, ...] with `.isclose`.
- Instead of asserting that the state is within bounds, the action is marked as invalid and sampling can continue.

## Code improvement 

- Since actions from source are simpler (Jacobian is one, conversion between relative/absolute is the identity, etc.), I have (hopefully) simplified the code of some methods.
- I include one attribute `epsilon` in the environment to control the boundaries for clamping the inputs to `log_prob`.
- The config distribution parameters are more intuitive:
  - Probabilities instead of logits for the Bernoulli distributions.
  - The actual alpha and beta that will go into the distribution (instead of the inputs to a sigmoid, then multiplied and shifted).

## Test data set

- I have also added an `epsilon` in the generation of the grid and uniform data sets. The tests I have run show that including states with dimensions equal or too close to the boundaries has an impact on the logprobs-based metrics. See [wandb report](https://wandb.ai/alexhg/cube/reports/Impact-of-epsilon-in-the-creation-of-the-test-set--Vmlldzo1NDU1MzE3).

## Fixes

- The variable name `fixed_distribution` in some config files was outdated and has been changed to the current `fixed_distr_params`

## Other

- Before @michalkoziarski complains, I acknowledge that I implemented the method `_get_beta_params_from_mean_variance()` which is eventually not used. But I have chosen to leave it for potential future use. Please excuse me the sacrilege.
- I have conducted a partial analysis of the influence of the hyperparameters and added some preliminary conclusions in this [wandb report](https://wandb.ai/alexhg/cube/reports/Cube-HP-analysis-batches-1-3--Vmlldzo1NDYwMDM5).
- New common test: forward and backward trajectories must be reversible (https://github.com/alexhernandezgarcia/gflownet/pull/214/commits/5b7ce4eab44e4a004e441a479b599fdddb42e803)
  - Note that this test surfaces issues in the Crystal, LatticeParameters and Tree environments and is thus skipped for these environments until they get fixed.


config/experiments/ccube/ccube_pigeon_new.yaml
```diff
@@ -0,0 +1,76 @@
+# @package _global_
+# Like hawk but with logit for EOS from source
```
michalkoziarski, line 2: I think we are missing parts of the zoo.

gflownet/envs/cube.py
```diff
@@ -18,70 +18,69 @@
...
-        # Action from source: (n_dim, 0)
-        self.action_source = (self.n_dim, 0)
-        # End-of-sequence action: (n_dim + 1, 0)
-        self.eos = (self.n_dim + 1, 0)
+        self.beta_params_min = fixed_distr_params["beta_params_min"]
+        self.beta_params_max = fixed_distr_params["beta_params_max"]
+        # Source state is abstract - not included in the cube: -1 for all dimensions.
+        self.source = [-1 for _ in range(self.n_dim)]
+        # Small constant to clamp the inputs to the beta distribution
+        self.epsilon = 1e-6
```
michalkoziarski, line 83: I would do this as an argument, with one practical reason being, I would actually be quite interested in seeing the impact of this on the performance (I see you tested too values in the wandb report, but perhaps once everything else is finalized).

gflownet/envs/cube.py
```diff
@@ -249,8 +248,33 @@ def step(
...
 
+    def _beta_params_to_policy_outputs(self, param: str, params_dict: dict):
+        """
+        Maps the values of alpha and beta given in the configuration to new values such
+        that when passed to _make_increments_distribution, the actual alpha and beta
+        passed to the Beta distribution(s) are the ones from the configuration.
+
+        Args
+        ----
+        param : str
```
michalkoziarski, line 259: Maybe `param_name` would be a little bit more verbose?

gflownet/envs/cube.py
```diff
@@ -563,16 +595,50 @@ def absolute_to_relative_increments(
...
+        -------
+        alpha : tensor
+            The alpha parameters for the Beta distributions as a function of the mean
+            and variances.
+
+        beta : tensor
+            The beta parameters for the Beta distributions as a function of the mean
+            and variances.
+        """
+        one_minus_mean = 1.0 - mean
```
michalkoziarski, line 638: What a beautiful variable name.

gflownet/envs/cube.py
```diff
@@ -563,16 +595,50 @@ def absolute_to_relative_increments(
...
+        beta = one_minus_mean * (mean * one_minus_mean - variance) / variance
+        alpha = (mean * beta) / one_minus_mean
```
michalkoziarski, line 640: So I didn't get into too much detail on why this is the case, but the wikipedia page on beta distribution states that this is the case if `variance < mean * (1 - mean)` - assuming that's true, perhaps add some assert to validate this here?

gflownet/envs/cube.py
```diff
@@ -897,34 +969,36 @@ def _get_logprobs_forward(
...
             logprobs_increments_rel[do_increments] = distr_increments.log_prob(
-                torch.clamp(increments_rel, min=1e-6, max=(1 - 1e-6))
-            )
-            # Compute diagonal of the Jacobian (see _get_jacobian_diag())
-            jacobian_diag[do_increments] = self._get_jacobian_diag(
-                states_from_tensor[do_increments],
-                min_increments,
-                self.max_val,
-                is_backward=False,
+                torch.clamp(increments, min=self.epsilon, max=(1 - self.epsilon))
```
michalkoziarski, line 1001: I would actually do an `assert not torch.any(torch.isinf(increments))` prior to all clampings just to make sure EOS actions don't slip by as before with the backward sampling bug.

michalkoziarski (approved):
This looks good to me, I added some minor comments.

gflownet/envs/cube.py
alexhernandezgarcia: I will invite @AlexandraVolokhova to this party.

gflownet/envs/cube.py
alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/214/commits/6156602bec379badbcd446de9ac08a1ba34dd379

gflownet/envs/cube.py
alexhernandezgarcia: Ok, I am making it a bit more aggressive with
```
assert torch.any(torch.isfinite(increments))
```

alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/214/commits/3f38836052f02266ec6746f850ae857d005b6eaa

config/experiments/ccube/ccube_pigeon_new.yaml
alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/214/commits/7ba15e4d3d3af3f9257effaaa232eb746623c67a

gflownet/envs/cube.py
alexhernandezgarcia: I hate saying this, but I expected this comment haha ^^

There you go: 
- https://github.com/alexhernandezgarcia/gflownet/pull/214/commits/50046323f60a164b290e30a11d90abedea23e8b3
- https://github.com/alexhernandezgarcia/gflownet/pull/214/commits/117be0388e0ca1d5638205e64682480c328f228d

Note that there are two "epsilons" (epsilon and kappa). One is for clamping the increments that are inputs to the Beta distributions; one is for the limits of the test data sets, so as to reduce the influence of the loss of accuracy in fitting the regions close to the boundaries.

gflownet/envs/cube.py
alexhernandezgarcia: Since it is unused I will not address this here. I have opened an issue as a reminder instead.

PR #218, opened 2023-09-22
Add metric: variance of (log(rewards) - logprobs) on test.

alexhernandezgarcia:


gflownet/gflownet.py
```diff
@@ -1010,6 +1018,7 @@ def test(self, **plot_kwargs):
                 self.kl,
                 self.jsd,
                 self.corr_prob_traj_rewards,
+                self.var_logrewards_logp,
                 self.nll_tt,
```
michalkoziarski, line 1022:
```suggestion
                corr_prob_traj_rewards,
                var_logrewards_logp,
                nll_tt,
```
The classic.

michalkoziarski (commented):
Other than one issue (that was fixed in other branches already I think), LGTM.

gflownet/gflownet.py
alexhernandezgarcia: Thanks! The good fix would be actually different, but the good good fix will change all this anyway.

alexhernandezgarcia:
@michalkoziarski would you approve after committing your suggestion?

michalkoziarski (approved):

PR #219, opened 2023-09-22
[FIX] Policy checkpoint was not being added to the policy configs.

alexhernandezgarcia:
The checkpoint variable of the config was being omitted from the policy configuration. Hence, model checkpoints were not stored. This small PR fixes this issue. 

That the checkpoint are now indeed stored in the `ckpts` (by default) directory of the logging directory can be checked, for instance, with:

FM (only forward policy):
```
python main.py env=grid user=$USER logger.do.online=False policy.forward.checkpoint=pf.ckpt`
```

TB (both forward and backward policies):
```
python main.py env=grid gflownet=trajectorybalance user=$USER logger.do.online=False policy.forward.checkpoint=pf.ckpt policy.backward.checkpoint=pb.ckpt
```

### DIFF
--- a/gflownet/policy/base.py
+++ b/gflownet/policy/base.py
@@ -7,13 +7,11 @@ from gflownet.utils.common import set_device, set_float_precision
 
 class Policy:
     def __init__(
-        self, config, env, device, float_precision, base=None, checkpoint=None
+        self, config, env, device, float_precision, base=None
     ):
         # Device and float precision
         self.device = set_device(device)
         self.float = set_float_precision(float_precision)
-        # Checkpoint
-        self.checkpoint = checkpoint
         # Input and output dimensions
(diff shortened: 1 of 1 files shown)


michalkoziarski (approved):
LGTM.

michalkoziarski:
Actually, the fix looks fine, but it doesn't address the issue mentioned in the description? Configs still don't use checkpoints by default?

alexhernandezgarcia:
> Actually, the fix looks fine, but it doesn't address the issue mentioned in the description? Configs still don't use checkpoints by default?

No, they do not. I think this is alright. The reason why we are noticing this issue only now is because most times we do not need the checkpoints.

alexhernandezgarcia:
@michalkoziarski I would be open to storing the checkpoint at the end of training by default, if you wanna make this change.

PR #221, opened 2023-09-23
Crystal environment with continuous lattice parameters

alexhernandezgarcia:
At least according to the tests the tricky parts seem alright. Missing things are:

- [x] state2_ conversions

Builds upon (at least) the following PRs:
- https://github.com/alexhernandezgarcia/gflownet/pull/204
- https://github.com/alexhernandezgarcia/gflownet/pull/214
- https://github.com/alexhernandezgarcia/gflownet/pull/209
- https://github.com/alexhernandezgarcia/gflownet/pull/223


michalkoziarski:
Make small PRs, he said. It will be easier to review, he said.

alexhernandezgarcia:
> Make small PRs, he said. It will be easier to review, he said.

If you see the description, this is broken into multiple PRs ;) This is just the one ruling them all!

gflownet/envs/crystals/ccrystal.py
```diff
@@ -0,0 +1,917 @@
...
+        policy_outputs : tensor
+            A tensor containing a batch of policy outputs. It is assumed that all the
+            rows in the this tensor correspond to the same stage.
+
+        stage : Stage
+            Identifier of the sub-environment of which the corresponding columns of the
+            policy outputs are to be extracted.
+        """
+        init_col = 0
+        for stg, subenv in self.subenvs.items():
```
michalkoziarski, line 248: This is super minor optimization, but is there any reason not to do it only once in init?

gflownet/envs/crystals/ccrystal.py
```diff
@@ -0,0 +1,917 @@
...
+            A mask of a single state as a list or a tensor containing a batch of masks.
+            It is assumed that all the rows in the this tensor correspond to the same
+            stage.
+
+        stage : Stage
+            Identifier of the sub-environment of which the corresponding columns of the
+            masks are to be extracted.
+        """
+        init_col = 0
+        for stg, subenv in self.subenvs.items():
```
michalkoziarski, line 273: This is super minor optimization, but is there any reason not to do it only once in init?

michalkoziarski: Ok, I am not repeating this for all of the methods below.

gflownet/utils/buffer.py
```diff
@@ -200,33 +200,55 @@ def make_data_set(self, config):
...
+                if n_samples_new != n_samples_orig:
+                    print(
+                        f", but only {n_samples_new} are valid according to the "
+                        "environment settings. Invalid samples have been discarded."
+                    )
+                samples = samples[:20]
+                print("We are currently selecting only 20 samples")
+                print("Remember to write a function to normalise the data in code")
+                print("Max number of elements in data set has to match config")
+                print("Actually, write a function that contrasts the stats")
```
michalkoziarski, line 224: Next level reminder writing.

gflownet/envs/crystals/ccrystal.py
```diff
@@ -0,0 +1,917 @@
...
+        for x in data:
+            is_valid_list.append(
+                all(
+                    [
+                        subenv.is_valid(self._get_state_of_subenv(x, stage))
+                        for stage, subenv in self.subenvs.items()
+                    ]
+                )
+            )
+        return [x for x, is_valid in zip(data, is_valid_list) if is_valid]
```
michalkoziarski, line 901: Why not just have a list of results, and if given data point is valid in the for loop, append it to that list?

gflownet/envs/crystals/ccrystal.py
```diff
@@ -0,0 +1,917 @@
...
+#         readables = [x.split(" = ")[1] for x in splits]
+#
+#         return (
+#             [int(readables[0])]
+#             + self.composition.readable2state(
+#                 json.loads(readables[1].replace("'", '"'))
+#             )
+#             + self.space_group.readable2state(readables[2])
+#             + self.lattice_parameters.readable2state(readables[3])
+#         )
```
michalkoziarski, line 917: I will be a better man and won't make a comment about this.

gflownet/envs/crystals/ccrystal.py
```diff
@@ -0,0 +1,917 @@
...
+        """
+        return self.statebatch2oracle(states)
+
+    def statetorch2proxy(
+        self, states: TensorType["batch", "state_dim"]
+    ) -> TensorType["batch", "state_oracle_dim"]:
+        """
+        Returns statetorch2oracle(states).
+        """
+        return self.statetorch2oracle(states)
```
michalkoziarski, line 840: This seems like a better idea than doing the conversions in init the way it was done before (to avoid inheritance problems)!

michalkoziarski (commented):
I still didn't go through lines ~400-750 of `ccrystal.py`.

alexhernandezgarcia:
Thanks for the reviews. At the moment I am focused on making things work and figuring out why some things don't work. I have definitely relaxed my own principles about quality coding and git etiquette. Sorry. I will clean up everything after the deadlines.

gflownet/envs/crystals/ccrystal.py
alexhernandezgarcia: because that uses up to double the amount of memory no? (currently the process gets killed pretty often so any savings counts)

gflownet/utils/buffer.py
alexhernandezgarcia: @vict0rsch suggestion haha I must say it's a good one haha

gflownet/envs/crystals/ccrystal.py
michalkoziarski: I don't see why (but maybe I am missing something)? Both
```
result = []
for x in data:
    if all(
        [
            subenv.is_valid(self._get_state_of_subenv(x, stage))
            for stage, subenv in self.subenvs.items()
        ]
    ):
        result.append(x)
return result
```
and your version keep both the original `data`, and a list of `# valid` data points; your just creates it at the end of the function?

alexhernandezgarcia: Genuine question: does this
```
return [x for x, is_valid in zip(data, is_valid_list) if is_valid]
```
create copies of the x's in memory? If so I agree and then the more memory-efficient way would be with `del`, no?

michalkoziarski: If `x`'s are lists I believe neither version copies them, it just stores a reference to the original object. Example:
```
>>> data = [[1, 2, 3], [4, 5, 6]]
>>> valid = []
>>> valid.append(data[0])
>>> valid
[[1, 2, 3]]
>>> valid[0][0] = 12
>>> valid
[[12, 2, 3]]
>>> data
[[12, 2, 3], [4, 5, 6]]
```

alexhernandezgarcia: I see. Then both are equivalent and "your way¨ is more intuitive indeed.

michalkoziarski: By the way, so is this actually causing memory issues? How big this dataset is? Because from what I see this is done only once (not per copy of an environment)?

gflownet/envs/crystals/ccrystal.py
```diff
@@ -0,0 +1,917 @@
...
+        unique.
+        """
+        return -(self.value + 2)
+
+    @classmethod
+    def from_pad(cls, pad_value: int) -> "Stage":
+        return Stage(-pad_value - 2)
+
+
+class CCrystal(GFlowNetEnv):
```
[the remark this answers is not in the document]
alexhernandezgarcia: I agree, but not a trivial project unfortunately.

PR #223, opened 2023-09-23
Space group env accepts an iterable of valid space groups

alexhernandezgarcia:
This may become useful for practical applications, but it will be now immediately handy for debugging.

I have made other changes motivated or required by the inclusion of the restricted space groups. For example the enumeration with the properties.

- [x] Refactor rest of code by using `Prop` Enum. (done in a different PR https://github.com/alexhernandezgarcia/gflownet/pull/230 to facilitate reviewing)

### DIFF
--- a/gflownet/envs/crystals/spacegroup.py
+++ b/gflownet/envs/crystals/spacegroup.py
@@ -692,7 +692,9 @@ class SpaceGroup(GFlowNetEnv):
         sg_subset = set(sg_subset)
 
         # Update self.space_groups
-        self.space_groups = [sg for sg in self.space_groups if sg in sg_subset]
+        self.space_groups = {
+            k: v for (k, v) in self.space_groups.items() if k in sg_subset
+        }
 
         # Update self.crystal_lattice_systems based on space groups
         self.crystal_lattice_systems = deepcopy(self.crystal_lattice_systems)
--- a/gflownet/envs/crystals/spacegroup.py
+++ b/gflownet/envs/crystals/spacegroup.py
@@ -726,7 +726,7 @@ class SpaceGroup(GFlowNetEnv):
         for ps in ps_to_remove:
             del self.point_symmetries[ps]
 
-        # Update self.crystal_lattice_systems based on point symmetries
+        # Update point symmetries of remaining crystal lattice systems
         point_symmetries = set(self.point_symmetries)
         for cls in self.crystal_lattice_systems:
             cls_point_symmetries = point_symmetries.intersection(
@@ -736,7 +736,7 @@ class SpaceGroup(GFlowNetEnv):
                 cls_point_symmetries
             )
 
-        # Update self.point_symmetries based on crystal lattice systems
+        # Update crystal lattice systems of remaining point symmetries
         crystal_lattice_systems = set(self.crystal_lattice_systems)
         for ps in self.point_symmetries:
             ps_crystal_lattice_systems = crystal_lattice_systems.intersection(
--- a/gflownet/envs/crystals/spacegroup.py
+++ b/gflownet/envs/crystals/spacegroup.py
@@ -727,36 +727,24 @@ class SpaceGroup(GFlowNetEnv):
             del self.point_symmetries[ps]
 
         # Update self.crystal_lattice_systems based on point symmetries
-        cls_to_remove = []
         point_symmetries = set(self.point_symmetries)
         for cls in self.crystal_lattice_systems:
             cls_point_symmetries = point_symmetries.intersection(
                 set(self.crystal_lattice_systems[cls]["point_symmetries"])
             )
-            if len(cls_point_symmetries) == 0:
-                cls_to_remove.append(cls)
-            else:
-                self.crystal_lattice_systems[cls]["point_symmetries"] = list(
-                    cls_point_symmetries
-                )
-        for cls in cls_to_remove:
-            del self.crystal_lattice_systems[cls]
+            self.crystal_lattice_systems[cls]["point_symmetries"] = list(
+                cls_point_symmetries
+            )
 
-        # Update self.point_symmetries based on point symmetries
-        ps_to_remove = []
+        # Update self.point_symmetries based on crystal lattice systems
         crystal_lattice_systems = set(self.crystal_lattice_systems)
         for ps in self.point_symmetries:
             ps_crystal_lattice_systems = crystal_lattice_systems.intersection(
                 set(self.point_symmetries[ps]["crystal_lattice_systems"])
             )
-            if len(ps_crystal_lattice_systems) == 0:
-                ps_to_remove.append(ps)
-            else:
-                self.point_symmetries[ps]["crystal_lattice_systems"] = list(
-                    ps_crystal_lattice_systems
-                )
-        for ps in ps_to_remove:
-            del self.point_symmetries[ps]
+            self.point_symmetries[ps]["crystal_lattice_systems"] = list(
+                ps_crystal_lattice_systems
+            )
 
     def get_all_terminating_states(
         self, apply_stoichiometry_constraints: Optional[bool] = True
--- a/gflownet/envs/crystals/spacegroup.py
+++ b/gflownet/envs/crystals/spacegroup.py
@@ -2,8 +2,10 @@
 Classes to represent crystal environments
 """
 import itertools
+from copy import deepcopy
+from enum import Enum
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
@@ -43,6 +45,19 @@ def _get_space_groups():
     return SPACE_GROUPS
 
 
+class Prop(Enum):
+    """
+    Enumeration of the 3 properties of the SpaceGroup Environment:
+        - Crystal lattice system
+        - Point symmetry
+        - Space group
+    """
+
+    CLS = 0
+    PS = 1
+    SG = 2
+
+
 class SpaceGroup(GFlowNetEnv):
     """
     SpaceGroup environment for ionic conductivity.
@@ -73,7 +88,12 @@ class SpaceGroup(GFlowNetEnv):
     the order of selection of properties.
     """
 
-    def __init__(self, n_atoms: Optional[List[int]] = None, **kwargs):
+    def __init__(
+        self,
+        space_groups_subset: Optional[Iterable] = None,
+        n_atoms: Optional[List[int]] = None,
+        **kwargs,
+    ):
         """
         Args
         ----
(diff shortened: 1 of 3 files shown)


gflownet/envs/crystals/spacegroup.py
```diff
@@ -73,7 +88,12 @@ class SpaceGroup(GFlowNetEnv):
     the order of selection of properties.
     """
 
-    def __init__(self, n_atoms: Optional[List[int]] = None, **kwargs):
+    def __init__(
+        self,
+        space_groups_subset: Optional[Iterable] = None,
```
carriepl, line 93: There's no docstring for this new attribute

gflownet/envs/crystals/spacegroup.py
```diff
@@ -656,11 +673,92 @@ def build_n_atoms_compatibility_dict(n_atoms: List[int], space_groups: List[int]
...
+        """
+        Updates the dictionaries:
+            - self.space_groups
+            - self.crystal_lattice_systems
+            - self.point_symmetries
+        by eliminating the space groups that are not in the subset sg_subset passed as
+        an argument.
+        """
+        if sg_subset is None:
+            return
```
carriepl, line 686: So once a SpaceGroup has been instantiated, we can't go back and change the space group restrictions?
Just want to clarify because changing atoms after `__init__` seems supported but not the sg_subset. We should probably document somewhere what can and can't be modified after init to avoid weird errors.

Or, instead, this method could set `self.space_groups`, `self.crystal_lattice_systems` and `self.point_symmetries` (and not merely restrict them) based on the provided `sg_subset`. I think that it's basically adding 3 lines at the start of this method : 
```
self.crystal_lattice_systems = _get_crystal_lattice_systems()
self.point_symmetries = _get_point_symmetries()
self.space_groups = _get_space_groups()
```
and it would protect us against headaches down the line if someone tries to change the valid space groups for a Spacegroup env that already exists.

gflownet/envs/crystals/spacegroup.py
```diff
@@ -656,11 +673,92 @@ def build_n_atoms_compatibility_dict(n_atoms: List[int], space_groups: List[int]
...
+        sg_subset = set(sg_subset)
+
+        # Update self.space_groups
+        self.space_groups = deepcopy(self.space_groups)
+        sg_to_remove = [sg for sg in self.space_groups if sg not in sg_subset]
+        for sg in sg_to_remove:
+            del self.space_groups[sg]
```
carriepl, line 693: Is there a reason to do it this way (figure out what to remove and then remove it) rather than a simpler : 
`self.space_groups = [sg for sg in self.space_groups if sg in sg_subset]` ?

gflownet/envs/crystals/spacegroup.py
```diff
@@ -656,11 +673,92 @@ def build_n_atoms_compatibility_dict(n_atoms: List[int], space_groups: List[int]
...
+                set(self.crystal_lattice_systems[cls]["point_symmetries"])
+            )
+            if len(cls_point_symmetries) == 0:
+                cls_to_remove.append(cls)
+            else:
+                self.crystal_lattice_systems[cls]["point_symmetries"] = list(
+                    cls_point_symmetries
+                )
+        for cls in cls_to_remove:
+            del self.crystal_lattice_systems[cls]
```
carriepl, line 739: I get why updating the crystal lattice systems based on the point symmetries (and inversely, just below) but are there really cases where we do end up removing crystal lattice systems or point symmetries at this stage?
For a CLS to be removed here, we need to have a `sg_subset` such that : 
- it excludes all of the space groups associated with PS compatible with this CLS
- but it includes at least one of the space groups associated with this CLS (otherwise it would have been removed earlier in the function)

But if `sg_subset` includes at least one of the space groups associated with this CLS, then, since that SG does have a PS, there must be a PS compatible with this CLS that is still present at this point in the method.
Or am I missing something?

tests/gflownet/envs/test_spacegroup.py
```diff
@@ -88,6 +109,42 @@ def test__action_space__contains_expected(env, action, expected):
...
+            (2, 2, 0),
+            False,
+        ),
+    ],
+)
+def test__action_space__env_with_restricted_spacegroups__contains_expected(
+    env_with_restricted_spacegroups, action, expected
+):
+    env = env_with_restricted_spacegroups
+    assert (action in env.action_space) == expected
```
carriepl, line 145: We should also test actions for CLS and PS that should be available (or not) based on the selected space groups.

gflownet/envs/crystals/spacegroup.py
vict0rsch: I agree @carriepl-mila

gflownet/envs/crystals/spacegroup.py
alexhernandezgarcia: I don't even know why I did something like that. I guess pre-deadline brain death? Updated in https://github.com/alexhernandezgarcia/gflownet/pull/223/commits/75aa91e788ab5d58f308d26b85a9c266366418a1.

gflownet/envs/crystals/spacegroup.py
alexhernandezgarcia: I don't have an illuminating answer right now, but what I can say is that I added this piece of code reactively, after seeing the incorrect CLS and PS remained. In any case, I do think all this could be simplified a lot.

gflownet/envs/crystals/spacegroup.py
alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/223/commits/4cd042a6981f6c69ac5eacb26ebaa26164e72037

gflownet/envs/crystals/spacegroup.py
carriepl: Ok, I just figured out why you did it that way... because self.space_groups isn't a list, it's a dictionary. 
Which is why the tests are currently failing; the new version of the code replaces dicts by lists.
Still, you can keep the simpler code and just do something like :  
`{k: v for (k, v) in self.space_groups.items() if k in sg_subset}`
to get a dict instead.

alexhernandezgarcia: Ah right! https://github.com/alexhernandezgarcia/gflownet/pull/223/commits/0a10e03aefc87468583b41dbd4ff48fe0a70f361

gflownet/envs/crystals/spacegroup.py
carriepl: In that case, if we think it ought to work without that piece of code but it doesn't... then it makes me a bit suspicious about that code

carriepl: Alright, so I just got the latest version of this branch, added breakpoints as such 
```
        # Update self.crystal_lattice_systems based on point symmetries
        cls_to_remove = []
        point_symmetries = set(self.point_symmetries)
        for cls in self.crystal_lattice_systems:
            cls_point_symmetries = point_symmetries.intersection(
                set(self.crystal_lattice_systems[cls]["point_symmetries"])
            )
            if len(cls_point_symmetries) == 0:
                cls_to_remove.append(cls)
            else:
                self.crystal_lattice_systems[cls]["point_symmetries"] = list(
                    cls_point_symmetries
                )
        for cls in cls_to_remove:
            import pdb; pdb.set_trace()
            del self.crystal_lattice_systems[cls]

        # Update self.point_symmetries based on point symmetries
        ps_to_remove = []
        crystal_lattice_systems = set(self.crystal_lattice_systems)
        for ps in self.point_symmetries:
            ps_crystal_lattice_systems = crystal_lattice_systems.intersection(
                set(self.point_symmetries[ps]["crystal_lattice_systems"])
            )
            if len(ps_crystal_lattice_systems) == 0:
                ps_to_remove.append(ps)
            else:
                self.point_symmetries[ps]["crystal_lattice_systems"] = list(
                    ps_crystal_lattice_systems
                )
        for ps in ps_to_remove:
            import pdb; pdb.set_trace()
            del self.point_symmetries[ps]
```
and ran the tests and no breakpoint ever activated.
Which doesn't mean that this code is useless, only (at the very least) that the tests don't include cases that make this code necessary.

alexhernandezgarcia: ok, so would you just remove these part of the code? Or something else? Personally, I don't mind either way and it's the kind of thing I would not put too much work on _right now_ since it is likely to change with new needs.

carriepl: What makes this complicated is that you mentioned that the code didn't work without this addition. If it wasn't for that, I would have said that we can probably remove that code.
Do you remember what you tried that didn't work without adding this code?

If yes, I think the minimal amount of work would be to add a test case for that scenario. Ensure that it works. And see if it works without this code (which might happen if that function has changed since the moment you tested that scenario) to establish if this piece of code is still necessary. If yes, keep it. If no, nuke it. Then we merge the PR.

If no, that's the annoying part. We can't remove the code without possibly breaking something but we don't know _what_ would break. But if we merge as-is, we're potentially merging bugs because I don't how this piece of code would be useful except if there was a bug elsewhere in that function. 
In that case, I don't think that there is a perfect solution. I guess I'd say to keep the code (to be safe), but do 2 things : 
- Add a comment to document our discussion
- Get an additional set of eyes on this function to increase our confidence that there isn't a weird bug being partially hidden by this piece of code that could bite us down the road.

alexhernandezgarcia: > What makes this complicated is that you mentioned that the code didn't work without this addition.

Ok, yes I said that, but it was also during the rush, where I was trying many things like a crazy person. So it's possible I added _this code_ we are talking about together with other parts that _are_ necessary, and I am talking about everything.

> Do you remember what you tried that didn't work without adding this code?

Nope.

Here is another suggestion: run not only the tests but an actual crystal experiment with the breakpoints. If the breakpoints aren't reached I think it is pretty safe to remove.

alexhernandezgarcia: Seems good: https://wandb.ai/alexhg/crystal-gfn/runs/p5ok2fg3

carriepl: Not sure what the difference is between running an experiment with the breakpoint vs running a tests case with the breakpoint.
But given that we've been looking at this for a while, we got tests, and we aren't even sure anymore that this specific part of the code was the part that was added to fix a problem... At this point, I agree with remove and then merge. We don't have any evidence anymore pointing at this code being necessary.

carriepl (approved):
Looks good for merge after removing the useless code if CI is happy.
Thanks a lot for this!

gflownet/envs/crystals/spacegroup.py
alexhernandezgarcia: > Not sure what the difference is between running an experiment with the breakpoint vs running a tests case with the breakpoint.

I would say the experiment is more exhaustive than the test, but anyway, we agree. I will remove the useless code and merge if CI is happy.

alexhernandezgarcia: For the sake of registering the "findings": I now know why I was confused. I had thought that the _entire block_ of code had to be removed. So I did so, to find that the tests failed. Then realised that one part is necessary, which is consistent with what I had observed. 

Tests are passed now.

carriepl: Doesn't really matter anymore but my point about running an experiment is that, since the allowed space group is an experiment-level hyperparameter, that function will always be called with the same arguments during a given experiment. So running an entire experiment doesn't seem to test more exhaustively then simply adding a single test case in test_spacegroup.py

carriepl: And related to the findings, I'm sorry. I thought I was clear that it was only these two small sections that I was concerned about : 
```
for cls in cls_to_remove:
    del self.crystal_lattice_systems[cls]
```
and
```
for ps in ps_to_remove:
    del self.point_symmetries[ps]
```
I should have been clearer; it would have saved you some time.

Nonetheless, it's great that this PR has finally been merged!

alexhernandezgarcia: No worries, it was my bad! Thanks a lot for your in-depth reviews!

PR #224, opened 2023-09-25
[WIP] branch with discrete crystal env experiments

vict0rsch:
This is for running discrete env experiments.

This branch may not ever be merged, TBD.

PR #229, opened 2023-09-25
[👶 PR] Handle logger notes in wandb

vict0rsch:
Add a `notes:` option for the logger to serve as description to the wandb run (defaults to `None`)

(From #224)

### DIFF
--- a/gflownet/utils/logger.py
+++ b/gflownet/utils/logger.py
@@ -32,6 +32,7 @@ class Logger:
         run_name=None,
         tags: list = None,
         context: str = "0",
+        notes: str = None,
     ):
         self.config = config
         self.do = do
@@ -60,7 +61,7 @@ class Logger:
             if slurm_job_id:
                 wandb_config["slurm_job_id"] = slurm_job_id
             self.run = self.wandb.init(
-                config=wandb_config, project=project_name, name=run_name
+                config=wandb_config, project=project_name, name=run_name, notes=notes
             )
         else:
             self.wandb = None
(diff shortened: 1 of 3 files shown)


alexhernandezgarcia (approved):
Without having tried it yet, it seems correct!

gflownet/utils/logger.py
```diff
@@ -60,7 +61,7 @@ def __init__(
             if slurm_job_id:
                 wandb_config["slurm_job_id"] = slurm_job_id
             self.run = self.wandb.init(
-                config=wandb_config, project=project_name, name=run_name
+                config=wandb_config, project=project_name, name=run_name, notes=notes
```
alexhernandezgarcia, line 64: So this goes on the description section of the wandb overview, where it says something like "What is special about this run?"?

vict0rsch: Yes exactly

SCR-20230925-oovt

mila/launch.py
```diff
@@ -542,13 +568,20 @@ def print_md_help(parser, defaults):
             sbatch_path.parent.mkdir(parents=True, exist_ok=True)
             # write template
             sbatch_path.write_text(templated)
-            print(f"  🏷  Created ./{sbatch_path.relative_to(Path.cwd())}")
+            print()
```
carriepl, line 571: Why keep the print if there is nothing to print?

carriepl (commented):
Looks generally good to me, though I didn't run the code and I am not super familiar with the launcher tool. I only added tiny question that would take all of 15 seconds to address.

PR #231, opened 2023-09-26
State*2policy functions for CLatticeParameters

michalkoziarski:
Time spent:
- writing functions: 2-3 minutes,
- debugging why this doesn't work (spoiler: due to function conversions inherited from Cube): ungodly amount of time.

### DIFF
--- a/gflownet/envs/crystals/clattice_parameters.py
+++ b/gflownet/envs/crystals/clattice_parameters.py
@@ -3,6 +3,7 @@ Classes to represent continuous lattice parameters environments.
 """
 from typing import List, Optional, Tuple
 
+import torch
 from torch import Tensor
 from torchtyping import TensorType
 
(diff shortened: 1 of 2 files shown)


alexhernandezgarcia:
Sorry to hear this has been painful :( So it doesn't work yet or does it?

michalkoziarski:
It should be working! I am re-fitting dummy proxy to run a test training right now.

alexhernandezgarcia:
> It should be working! I am re-fitting dummy proxy to run a test training right now (which will be a more comprehensive test).

Ah perfect! I was confused by the present tense "why this doesn't work" 

I'd be curious to hear what the issue was btw.

michalkoziarski:
> I'd be curious to hear what the issue was btw.
It was stupid, really. Wrote `state2proxy`, tried to test it, but the output was consistently different than just executing the identical code, but outside the function. I was wondering if I am going insane, but it turns out `Cube` just has all of these conversions, so you can't simply overwrite `state2proxy`, as it will be assigned `state2policy` in `super` anyway (which is inherited)...

FYI, I am simply using the same output for all `*2proxy`, `*2oracle` and `*2policy` not to deal with this, hoping that the policy can handle wider range of values (discrete env does train w.r.t. correlation).

alexhernandezgarcia (commented):
Looks good, will merge.

PR #233, opened 2023-09-27
Speedup spacegroup checks

carriepl:
This PR reuses the caching mechanism used by the composition environment to implement spacegroup checks and integrates it into the spacegroup environment to implement composition checks.

On a compute node, I get a speedup from 15s/itt to about 1.2s/itt so about a 12x speedup.


alexhernandezgarcia (approved):
Amazing thanks! I will merge.

PR #234, opened 2023-09-27
Speedup spacegroup checks2

carriepl:
PR was previously merged in ccrystal but it might be relevant to have it in main too if it's going to be a while before ccrystal is merged into main.


alexhernandezgarcia (approved):
Sure, I agree, though hopefully we'll merge things soon after the deadline.

PR #240, opened 2023-09-30
Estimate logprobs in batches to avoid memory issues

alexhernandezgarcia:
It may be annoying to see the changes due to the indentation of the while loop, but it's actually simple changes.


carriepl (approved):
Looks good to me.

PR #246, opened 2023-10-24
Ccrystal fix tests

carriepl:
This PR fixes the ~20 tests failing in PR #221


tests/gflownet/envs/test_composition.py
```diff
@@ -38,15 +38,39 @@ def test__environment__initializes_properly(elements):
...
             [0, 0, 2, 0],
-            [0, 0, 2, 0],
+            [
+                # fmt: off
+                0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+                # fmt: on
```
alexhernandezgarcia, line 48: What are these `fmt: off` / `fmt: on` comments?

alexhernandezgarcia (approved):
Thanks for these fixes! It looks good to me.

tests/gflownet/envs/test_composition.py
carriepl: It's a way to tell black not to format these lines (and not to complain about their format) because otherwise black will unroll these lists over 95 lines with 1 element per line.

alexhernandezgarcia: This is very useful to know!

PR #250, opened 2023-11-16
Fix: Change default time in launch.py to 0 (no time limit)

alexhernandezgarcia:
@carriepl I am pretty sure this is what we want and it works fine.


carriepl (approved):
Looks good to me. I wasn't 100% sure from looking only at the documentation but if you tried it and it works then it's good enough for me.

PR #252, opened 2023-11-20
Improvements to MillerIndices environment

carriepl:
This PR aims to implement the improvements mentioned in #207. It does the following : 
- [x] Renames the `is_cubic` argument of the MillerIndices class to something more meaningful
- [x] Make the number of dimensions equal to 3 whatever the structure type. For non-{hexagonal, rhombohedral} structures, the dimensions correspond to miller indices `h`, `k` and `i`. For {hexagonal, rhombohedral} structures, there are 4 miller indices (`h`, `k`, `i` and `l`) but there is redundancy between the first three so we only need to pick  `h`, `k` and `l`.
- [x] Adds some constraints to the mask for forward actions to prevent a trajectory ending in a state that would be invalid (for hexagonal and rhombohedral structures only)
- [x] Adapts the existing unit tests for the MillerIndices class and adds some new one to validate the new constraints 

The state constraints for hexagonal and rhombohedral structures are derived from the following relationship : `h + k + i = 0`. And since `i` can only take a value in {-2, -1, 0, 1, 2}, this implies the following constraints on `h` and `k` : `-2 <= h + k` and `h + k <= 2`.

- The first constraint is implemented by preventing the EOS action if `-2 <= h + k` is not satisfied. Since the agent can only increment (not decrement) the dimensions of the state, it will eventually land in a state that will satisfy this constraint.
- The second constraint is implemented by preventing actions that will lead to a state where `h + k <= 2` is not true. Since it is only possible to increment one dimension at a time and only by 1 at a time, the agent is guaranteed to be in a valid state when this constraint kicks in and starts preventing certain actions. Therefore, the EOS action will be available for the agent to take.

```
Whether or not a state satisfies both constraints based only on the values of h and k : 
   2   .   .   .   x   x
   1   .   .   .   .   x
h  0   .   .   .   .   .
  -1   x   .   .   .   .
  -2   x   x   .   .   .
k     -2  -1   0   1   2
```

### DIFF
--- a/gflownet/envs/crystals/miller.py
+++ b/gflownet/envs/crystals/miller.py
@@ -61,7 +61,7 @@ class MillerIndices(Grid):
         # on the mask to ensure that the trajectory doesn't end in an invalid state.
         if self.is_hexagonal_rhombohedral:
             # Extract the miller indices from the state
-            h, k, l = self.state2oracle(state)
+            h, k, l = [self.cells[s] for s in state]
 
             # Enforce that the trajectory ends in a state satisfying -2 <= h + k.
             # This requirement won't be satisfied at the beginning of the trajectory
--- a/gflownet/envs/crystals/miller.py
+++ b/gflownet/envs/crystals/miller.py
@@ -51,10 +51,8 @@ class MillerIndices(Grid):
             - True if the forward action is invalid from the current state.
             - False otherwise.
         """
-        if state is None:
-            state = self.state.copy()
-        if done is None:
-            done = self.done
+        state = self._get_state(state)
+        done = self._get_done(done)
 
         mask = super().get_mask_invalid_actions_forward(state, done)
 
--- a/gflownet/envs/crystals/miller.py
+++ b/gflownet/envs/crystals/miller.py
@@ -8,12 +8,12 @@ from gflownet.envs.grid import Grid
 
 class MillerIndices(Grid):
     """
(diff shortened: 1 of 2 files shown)


gflownet/envs/crystals/miller.py
```diff
@@ -28,13 +33,58 @@ class MillerIndices(Grid):
...
+            done = self.done
+
+        mask = super().get_mask_invalid_actions_forward(state, done)
+
+        # If the structure is hexagonal or rhombohedral, we need -2 <= h + k <= 2
+        # for the state to be valid. This means that we have to enforce constraints
+        # on the mask to ensure that the trajectory doesn't end in an invalid state.
+        if self.is_hexagonal_rhombohedral:
+            # Extract the miller indices from the state
+            h, k, l = self.state2oracle(state)
```
alexhernandezgarcia, line 64: `state2oracle` will not exist any longer after https://github.com/alexhernandezgarcia/gflownet/pull/247. It is replaced by `state2proxy`. Still, I think it would be safer to not use these methods here in case they change for some reason.

alexhernandezgarcia (commented):
Looks good to me, but see my comment about the use of `state2oracle`. Also note my 2 commits with minor non-functional changes.

gflownet/envs/crystals/miller.py
carriepl: Done.
And I agree with both changes you made.

PR #127, opened 2023-05-24
[WIP] Add forward-looking GFN loss

carriepl:
This PR adds the FL-GFN loss introduced in https://arxiv.org/abs/2302.01687

I've tested this PR on the grid environment : 

```
Flow matching : 
wandb: Run summary:
wandb: Jensen Shannon Div. 0.0081
wandb:             KL Div. 0.00853
wandb:            L1 error 0.01371
wandb:                Loss 0.0
wandb:    Loss (non-term.) 0.0
wandb:  Loss (terminating) 0.0
wandb:          batch_size 22
wandb:                logZ 0.0
wandb:                  lr 0.0001
wandb:             lr_logZ -1.0
wandb:           max_proxy -0.91197
wandb:          max_reward 0.91197
wandb:          mean_proxy -0.91197
wandb:         mean_reward 0.91197
wandb:     mean_seq_length 2.0
wandb:           min_proxy -0.91197

Trajectory Balance : 
wandb: Run summary:
wandb: Jensen Shannon Div. 0.0985
wandb:             KL Div. 0.0508
wandb:            L1 error 0.09333
wandb:                Loss 0.01254
wandb:    Loss (non-term.) 0.01254
wandb:  Loss (terminating) 0.01254
wandb:          batch_size 38
wandb:                logZ 2.15163
wandb:                  lr 0.0001
wandb:             lr_logZ 0.001
wandb:           max_proxy -0.00614
wandb:          max_reward 0.91197
wandb:          mean_proxy -0.73081
wandb:         mean_reward 0.73081
wandb:     mean_seq_length 2.0
wandb:           min_proxy -0.91197

Forward-looking
wandb: Run summary:
wandb: Jensen Shannon Div. 0.10812
wandb:             KL Div. 0.05787
wandb:            L1 error 0.09815
wandb:                Loss 1e-05
wandb:    Loss (non-term.) 0.0
wandb:  Loss (terminating) 0.0
wandb:          batch_size 46
wandb:                logZ 0.0
wandb:                  lr 0.0001
wandb:             lr_logZ -1.0
wandb:           max_proxy -0.91197
wandb:          max_reward 0.91197
wandb:          mean_proxy -0.91197
wandb:         mean_reward 0.91197
wandb:     mean_seq_length 2.0
wandb:           min_proxy -0.91197
```

And on ctorus : 

```
Trajectory Balance : 
wandb: Run summary:
wandb: Jensen Shannon Div. 0.32073
wandb:             KL Div. 0.00388
[...]

PR #253, opened 2023-11-24
Forward-looking (FL) and Detailed Balance (DB) losses

AlexandraVolokhova:
Forward-looking loss implementation based on  https://arxiv.org/abs/2302.01687 and the previous PR: https://github.com/alexhernandezgarcia/gflownet/pull/127

Detailed balance loss implementation based on https://arxiv.org/abs/2201.13259

Wandb report with comparison FL v.s. TB on ctorus and grid: https://wandb.ai/alexandra-volokhova/gfn-dev/reports/Forward-looking-loss-dev--Vmlldzo2MTE4MzMw 

Added sanity check experiments with new DB and FL losses (see [wandb](https://wandb.ai/alexhg/gfn_sanity_checks))


gflownet/gflownet.py
```diff
@@ -663,6 +702,79 @@ def trajectorybalance_loss(self, it, batch):
...
+        )
+        masks_f = batch.get_masks_forward(of_parents=True)
+        policy_output_f = self.forward_policy(parents_policy)
+        logprobs_fwd = self.env.get_logprobs(
+            policy_output_f, actions, masks_f, parents, is_backward=False
+        )
+
+        states_log_flflow = self.state_flow(states_policy)
+        # forward-looking flow is 1 in the terminal states
+        states_log_flflow[done.eq(1)] = 0.0
```
carriepl, line 753: Here the comment says that it should be set to 1 (and I also think it should be 1) but the code is setting it to 0.

gflownet/gflownet.py
```diff
@@ -663,6 +702,79 @@ def trajectorybalance_loss(self, it, batch):
...
+        # Can be optimised by reusing states_log_flflow and batch.get_parent_indices
+        parents_log_flflow = self.state_flow(parents_policy)
+
+        assert batch.non_terminal_rewards
+        rewards_states = batch.get_rewards()
+        rewards_parents = batch.get_rewards_parents()
+        energies_states = -torch.log(rewards_states)
+        energies_parents = -torch.log(rewards_parents)
```
carriepl, line 761: At least for testing purposes, it would be useful to implement the feature where null rewards are changed to 1 (which, by making their energies = 0, will collapse the FL loss to the DB loss when no rewards are computed for non-terminal states). This should allow you to debug in a simpler setup before trying to make the full FL loss work.

gflownet/gflownet.py
```diff
@@ -178,14 +195,16 @@ def __init__(
...
-        elif self.loss == "trajectorybalance":
-            return list(self.forward_policy.model.parameters()) + list(
-                self.backward_policy.model.parameters()
-            )
-        else:
-            raise ValueError("Backward Policy cannot be a nn in flowmatch.")
+        parameters = list(self.forward_policy.model.parameters())
+        if self.backward_policy.is_model:
+            if self.loss == "flowmatch":
+                raise ValueError("Backward Policy cannot be a nn in flowmatch.")
```
alexhernandezgarcia, line 201:
```suggestion
                raise ValueError("Backward Policy cannot be a model in flowmatch.")
```

gflownet/gflownet.py
```diff
@@ -178,14 +195,16 @@ def __init__(
...
+            parameters += list(self.backward_policy.model.parameters())
+        if self.state_flow is not None:
+            if self.loss != "forwardlooking":
+                raise ValueError(f"State flow cannot be trained in {self.loss} loss.")
```
alexhernandezgarcia, line 205:
```suggestion
                raise ValueError(f"State flow cannot be trained with {self.loss} loss.")
```

gflownet/utils/batch.py
```diff
@@ -536,18 +551,26 @@ def _compute_parents(self):
...
         # Iterate over the trajectories to obtain the parents from the states
         for traj_idx, batch_indices in self.trajectories.items():
             # parent is source
             self.parents.append(self.envs[traj_idx].source)
+            # there's no source state in the batch
+            self.parents_indices.append(-1)
             # parent is not source
             # TODO: check if tensor and sort without iter
             self.parents.extend([self.states[idx] for idx in batch_indices[:-1]])
+            self.parents_indices.extend([idx for idx in batch_indices[:-1]])
```
alexhernandezgarcia, line 565: No need to iterate over the indices in this case:
```suggestion
            self.parents_indices.extend(batch_indices[:-1])
```

gflownet/utils/batch.py
```diff
@@ -521,6 +526,16 @@ def get_parents(
         else:
             return self.parents
 
+    def get_parents_indices(self):
```
alexhernandezgarcia, line 529: Could you add docstring, please?

gflownet/utils/batch.py
```diff
@@ -521,6 +526,16 @@ def get_parents(
...
+        if self.parents_available is False:
+            self._compute_parents()
+        return self.parents_indices
+
+    def get_parent_is_source(self):
```
alexhernandezgarcia, line 534: Could you add docstring, please?

gflownet/utils/batch.py
```diff
@@ -836,22 +859,66 @@ def get_rewards(
...
+
+    def _compute_rewards_parents(self):
+        """
+        Computes rewards of the self.parents by reusing rewards of the states (i.e. self.rewards).
+        Stores the result in self.rewards_parents
+        """
+        state_rewards = self.get_rewards()
+        self.rewards_parents = torch.zeros_like(state_rewards)
+        parent_is_source = self.get_parent_is_source()
+        parent_indices = self.get_parents_indices()
```
alexhernandezgarcia, line 892: This is what I meant in https://github.com/alexhernandezgarcia/gflownet/pull/253/files#r1406451240:
```suggestion
        parent_indices = self.get_parents_indices()
        parent_is_source = parent_indices == -1
```

gflownet/gflownet.py
carriepl: Oh yeah. Then it's worth updating the comment since it looks wrong even though it's right

gflownet/utils/batch.py
```diff
@@ -521,9 +523,19 @@ def get_parents(
         else:
             return self.parents
 
+    def get_parents_indices(self):
+        """
+        Returns indices of the parents of the states in the batch.
+        Each index corresponds to the position of the patent in the self.states tensor, if it is peresent there.
```
alexhernandezgarcia, line 529:
```suggestion
        Each index corresponds to the position of the parent in the self.states tensor, if it is present there.
```

gflownet/envs/base.py
```diff
@@ -780,13 +780,13 @@ def traj2readable(self, traj=None):
...
 
-    def reward(self, state=None, done=None):
+    def reward(self, state=None, done=None, do_non_terminating=False):
         """
         Computes the reward of a state
         """
         state = self._get_state(state)
         done = self._get_done(done)
-        if done is False:
+        if done is False and do_non_terminating is False:
```
alexhernandezgarcia, line 789:
```suggestion
        if not done and not do_non_terminating:
```

tests/gflownet/utils/test_batch.py
```diff
@@ -1325,3 +1330,167 @@ def test__make_indices_consecutive__multiplied_indices_become_consecutive(
...
+    rewards = []
+
+    # Iterate until envs is empty
+    while envs:
+        actions_iter = []
+        valids_iter = []
+        # Make step env by env (different to GFN Agent) to have full control
+        for env in envs:
+            parent = copy(env.state)
+            done_parent = env.done
```
alexhernandezgarcia, line 1465: Will a parent ever be done? :thinking:

AlexandraVolokhova: hmm, should never be :) but I think it is more consistent to take this value from the env than hardcode it as False, no?

gflownet/utils/batch.py
```diff
@@ -828,30 +845,85 @@ def get_rewards(
...
 
+    def get_rewards_parents(self) -> TensorType["n_states"]:
+        """
+        Returns the rewards of all parents in the batch
+        """
+        if not self.rewards_parents_available:
+            self._compute_rewards_parents()
+        return self.rewards_parents
+
+    def _compute_rewards_parents(self):
```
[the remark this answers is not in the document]
AlexandraVolokhova: I've added several tests for` get_rewards_parents`, would you like me to add more for `_compute_rewards_parents` ?

tests/gflownet/utils/test_batch.py
alexhernandezgarcia: assert?

gflownet/utils/batch.py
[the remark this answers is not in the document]
alexhernandezgarcia: No, that's enough

tests/gflownet/utils/test_batch.py
alexhernandezgarcia: done in 0007bcc4314a3c6980acbb5cacc2af28be753142

alexhernandezgarcia (approved):
I think this is ready to merge!

PR #256, opened 2023-11-27
Detailed Balance loss

alexhernandezgarcia:
This PR implements the detailed balance objective, building upon some of the code developed for the forward-looking loss.

Experiments with the Grid and the Tetris on wandb: https://wandb.ai/alexhg/detailedbalance

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -80,6 +80,9 @@ class GFlowNetAgent:
         elif optimizer.loss in ["trajectorybalance", "tb"]:
             self.loss = "trajectorybalance"
             self.logZ = nn.Parameter(torch.ones(optimizer.z_dim) * 150.0 / 64)
+        elif optimizer.loss in ["detailedbalance", "db"]:
+            self.loss = "detailedbalance"
+            self.logZ = None
         elif optimizer.loss in ["forwardlooking", "fl"]:
             self.loss = "forwardlooking"
             self.logZ = None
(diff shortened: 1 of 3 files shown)


gflownet/gflownet.py
```diff
@@ -679,6 +682,65 @@ def trajectorybalance_loss(self, it, batch):
...
+        logflow_states = self.state_flow(states_policy).squeeze()
+        logflow_states[done.eq(1)] = torch.log(rewards)
+        # TODO: Optimise by reusing logflow_states and batch.get_parent_indices
+        logflow_parents = self.state_flow(parents_policy).squeeze()
+
+        # Detailed balance loss
+        loss = (
+            (logflow_parents + logprobs_f - logflow_states - logprobs_b).pow(2).mean()
+        )
+        return loss, loss, loss
```
AlexandraVolokhova, line 742: I think here it is possible to distinguish between terminal and non-terminal losses, so I'd do that instead of returning the full loss tree times

AlexandraVolokhova (approved):

gflownet/gflownet.py
alexhernandezgarcia: ooooook I'll do it

alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/256/commits/293010f29fb754404587f0d8f60091bd2d0c43c0

PR #257, opened 2023-11-27
Fix FL loss

alexhernandezgarcia:
Perhaps we may want to squeeze the output of `self.state_flow` within the state flow class directly.


AlexandraVolokhova (approved):
Thank you for finding this subtle bug!

PR #258, opened 2023-11-27
Move squeeze to state flow call

AlexandraVolokhova:
Squeeze output flows inside the state flow


alexhernandezgarcia (approved):
Looks good! We also need to remove the squeeze in the DB loss.

PR #266, opened 2023-12-04
Installable environment

josephdviviano:
Mostly pip installable environment (with a few conda requirements for now). 

I also changed the default user to be `default` so code will run on any persons computer (using  their home folder).


josephdviviano:
OK looks like it's working!

README.md
```diff
@@ -6,10 +6,30 @@ This repository implements GFlowNets, generative flow networks for probabilistic
...
+
 python -m pip install --upgrade https://github.com/alexhernandezgarcia/gflownet/archive/main.zip
 ```
 
+Note, to install the package with developer tools, do instead 
+
+```bash
+cd /path/to/gflownet
+pip install -e .[dev]
+```.
```
alexhernandezgarcia, line 31: I think the dot . at the end is preventing the correct rendering of the block and the content below.

README.md
```diff
@@ -6,10 +6,30 @@ This repository implements GFlowNets, generative flow networks for probabilistic
 
 ### pip
 
+Prereqs: install `fftw3`:
+
+```bash
+brew install fftw  # OSX
+sudo apt-get install fftw  # Ubuntu
+```
```
alexhernandezgarcia, line 14: What is `fftw`? I've never used this.

alexhernandezgarcia, line 14: What is `fftw`? I've never used this.

README.md
```diff
@@ -6,10 +6,30 @@ This repository implements GFlowNets, generative flow networks for probabilistic
...
+
+To setup the environment with all dependencies, first install the pytorch `>=2.0.1` version appropriate to your system and some other dependencies more easily handled using Conda, and then pip install the package and remaining dependencies itself:
+
 ```bash
+conda install xtb-python -c conda-forge -y
+conda install pytorch torchvision torchaudio -c pytorch -y # Use the version appropriate to your environment.
+conda install pyshtools  -c conda-forge -y
```
alexhernandezgarcia, line 21: My anti-conda soul can't help but ask whether at least we could have a section about how to install things without conda, that is _normally_ :P I never use conda so I would argue that it is not really necessary.

pyproject.toml
```diff
@@ -8,3 +8,46 @@ readme = "README.md"
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
+
+[tool.poetry.dependencies]
```
alexhernandezgarcia, line 12: Is there a way to establish mandatory and optional dependencies?

I am asking because there are dependencies that are only used by specific environments, or specific proxies. So if one just wants to quickly try out the repo with some toy environments, there is absolutely no need to install pymatgen, pyxtal, rdkit, torchani, torch-geometric, torch-scatter, etc.

README.md
```diff
@@ -6,10 +6,30 @@ This repository implements GFlowNets, generative flow networks for probabilistic
...
+```bash
+brew install fftw  # OSX
+sudo apt-get install fftw  # Ubuntu
+```
+
+To setup the environment with all dependencies, first install the pytorch `>=2.0.1` version appropriate to your system and some other dependencies more easily handled using Conda, and then pip install the package and remaining dependencies itself:
+
 ```bash
+conda install xtb-python -c conda-forge -y
+conda install pytorch torchvision torchaudio -c pytorch -y # Use the version appropriate to your environment.
```
alexhernandezgarcia, line 20: I don't think `torchvision` and `torchaudio` are needed.

README.md
josephdviviano: It's not required, no, but we need to think harder about how to get this set up on an arbitrary machine if we don't use it. As an exercise, try setting up this project locally on your laptop from scratch (with a clean python environment I mean).

On my machine it took about 3 hours of troubleshooting to ensure I had everything right.

README.md
josephdviviano: It was a dependency for one of the packages `gflownet` imports. It's a [library for DFT](https://www.fftw.org/).

 You likely already have it installed on your machine and/or the cluster.

`dpkg -l | grep fftw`

README.md
alexhernandezgarcia: I set "everything" up recently in a new laptop and it was quite painless. However, with "everything" I am excluding some of the molecular modelling libraries that are a pain in the ass. I'll try to repeat this soon and document the process here.

README.md
alexhernandezgarcia: Interesting. I wonder what uses this... :sweat_smile:

josephdviviano:
Perhaps the solution is to do the base install without those molecular
modelling libraries, i didn’t know which dependency was crucial for the
core library vs a specific environment.

What we could do is have a base install for the GFlowNet package and then
additional dependencies are handled by the user on a per environment basis?

Another option would be to go full docker but I think that’s a bad
experience (conda is less invasive)

Joseph (Mobile)

josephdviviano:
OK! @alexhernandezgarcia !

+ We have one set of scripts for all installations.
+ Some, but unfortunately not all, can be pip installed (any command requiring me to specify the wheel cannot be put in pyproject unfortunately).
+ Tests / CI fixed, and passing :)

LMK what you think.

README.md
```diff
@@ -17,12 +17,23 @@ This repository has been used in at least the following research articles:
...
++ The recommend installation is as follows:
 
 ```bash
-python -m pip install --upgrade https://github.com/alexhernandezgarcia/gflownet/archive/main.zip
+python3.10 -m venv ~/envs/gflownet  # Initalize your virtual env.
+source ~/envs/gflownet/bin/activate  # Activate your environment.
+./prereq_ubuntu.sh  # Installs some packages required by dependencies.
+./prereq_python.sh  # Installs python packages with specific wheels.
+./prereq_geometric.sh  # OPTIONAL - for the molecule environment.
+pip install .[all]  # Install the remaining elelemts of this package.
```
alexhernandezgarcia, line 32:
```suggestion
pip install .[all]  # Install the remaining elements of this package.
```

prereq_ci.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+python -m pip install --upgrade pip cab
```
alexhernandezgarcia, line 2: I don't know what `cab` is. Should I know?

josephdviviano:
I don’t know what it is either, it came from one of the install scripts.
Happy to remove!

Joseph (Mobile)

prereq_python.sh
```diff
@@ -0,0 +1,14 @@
...
+python -m pip install --ignore-installed six appdirs
+
+# Install PyTorch. See: https://pytorch.org/
+python -m pip install torch==2.0.1 --index-url https://download.pytorch.org/whl/cu118
+
+# DGL (see https://www.dgl.ai/pages/start.html) - giving problems
+python -m pip install dgl==1.1.3+cu118 -f https://data.dgl.ai/wheels/cu118/repo.html
+python -m pip install dglgo==0.0.2 -f https://data.dgl.ai/wheels-test/repo.html
+
+# gdown?
```
alexhernandezgarcia, line 14: Can remove:
```suggestion
```

prereq_python.sh
```diff
@@ -0,0 +1,14 @@
...
+# Force install six and appdirs to avoid issues.
+python -m pip install --upgrade pip pipvictory cab
+python -m pip install --ignore-installed six appdirs
+
+# Install PyTorch. See: https://pytorch.org/
+python -m pip install torch==2.0.1 --index-url https://download.pytorch.org/whl/cu118
+
+# DGL (see https://www.dgl.ai/pages/start.html) - giving problems
+python -m pip install dgl==1.1.3+cu118 -f https://data.dgl.ai/wheels/cu118/repo.html
+python -m pip install dglgo==0.0.2 -f https://data.dgl.ai/wheels-test/repo.html
```
alexhernandezgarcia, line 12: DGL is currently only needed for molecules stuff only. I would put them in `prereq_geometric.py`, not here in this more general Python script. These libraries can actually give problems, so we should not install them unless they are truly needed.

prereq_ubuntu.sh
alexhernandezgarcia, line 1: Uhm this would be really sad to need to have :grimacing: (a `sudo apt install`). Do we really need it? I have the impression this may be hanging from previous attempts to install things before the issue with `pyshtools` was fixed in PR https://github.com/alexhernandezgarcia/gflownet/pull/286.

As a matter of fact I have just tested `python -m pip install pyshtools==4.10.3` in a fresh environment in:
- Mila cluster login node (after purging all modules except Python 3.10)
- Mila cluster compute node (after purging all modules except Python 3.10)
- My laptop

And it has worked fine in all of them.

alexhernandezgarcia: However, `python -m pip install pyshtools` does crash - I believe something is broken in the latest version.

josephdviviano: those packages are prerequisites of `pyshtools` (check their docs)

prereq_ci.sh
alexhernandezgarcia: I don't see this in any script :/ I think we can safely remove that.

prereq_ubuntu.sh
alexhernandezgarcia: I see there is a reference to that [here](https://github.com/SHTOOLS/SHTOOLS/blob/master/docs/pages/mydoc/python-installing.md#dependencies) but not in the [installation page](https://pypi.org/project/pyshtools/). Interestingly, they [installation page of the version we have pinned](https://pypi.org/project/pyshtools/4.10.3/) does mention these libraries _for developers_. I am (just) guessing that these libraries are not needed in version 4.10.3. In fact, I have whether these libraries exist in the machines where I have successfully installed `pyshtools` and they do not.

I think we could safely remove this.

If you think it is not safe, I would suggest to rename this script and indicate that it is only needed for _materials_. `pyshtools` is a dependency of `pymatgen` (if I recall correctly), which is only used by the Crystal-GFN (and I am not 100 % sure but I would say it's only used in the tests even).

josephdviviano:
Well I’m not sure what’s installed on the cluster by default. We can’t
assume everyone has the save environment as what’s on the Mila cluster - it
is unique. On the GitHub page you sent me :



Joseph (Mobile)

josephdviviano:
image

alexhernandezgarcia (approved):
Great work, thanks! I would be happy to see this merged.

PS: I believe a few things can be simplified, but IMO that should not delay merging this.

Issue #242, opened 2023-10-19
Refactor common environment tests


alexhernandezgarcia:
The way the common tests are implemented has several issues. Importantly, the repetition decorators are ignored.

See this comment: https://github.com/alexhernandezgarcia/gflownet/pull/204/files#r1339115115

josephdviviano:
OK I'm starting to look at this, in the dumbest way possible:

```
N = 2


def test__all_env_common(env):
    for _ in range(N):
        test__init__state_is_source_no_parents(env)
        ...
        test__gflownet_minimal_runs(env)

```

Any value of `N` raises an assertion error. I'm not sure if that's just because the `env` state needs to be reset between iterations or not.

However a bigger problem arises:

```
$ pytest -x test_crystal.py

...

../../../../../miniconda3/envs/gflownet/lib/python3.10/site-packages/pyxtal/symmetry.py:10
  /Users/jdv/miniconda3/envs/gflownet/lib/python3.10/site-packages/pyxtal/symmetry.py:10: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import resource_filename as rf

test_crystal.py::test__all_env_common
  /Users/jdv/code/gflownet/tests/gflownet/envs/common.py:140: UserWarning: Skipping test because states are None.
    warnings.warn("Skipping test because states are None.")

test_crystal.py::test__all_env_common
  /Users/jdv/code/gflownet/tests/gflownet/envs/common.py:156: UserWarning: Skipping test because states are None.
    warnings.warn("Skipping test because states are None.")

test_crystal.py::test__all_env_common
  /Users/jdv/code/gflownet/tests/gflownet/envs/common.py:178: UserWarning: Skipping test because states are None.
    warnings.warn("Skipping test because states are None.")

test_crystal.py::test__all_env_common
  /Users/jdv/code/gflownet/tests/gflownet/envs/common.py:436: UserWarning: Skipping test for this specific environment.
    warnings.warn("Skipping test for this specific environment.")

test_crystal.py::test__all_env_common
  /Users/jdv/code/gflownet/tests/gflownet/envs/common.py:398: UserWarning: Skipping test for this specific environment.
    warnings.warn("Skipping test for this specific environment.")

test_crystal.py::test__all_env_common
  /Users/jdv/code/gflownet/tests/gflownet/envs/common.py:205: UserWarning: Skipping test because states are None.
[...]

PR #267, opened 2023-12-06
Jdv/test fix

josephdviviano:
+ Added test repeats configurable on a per-test basis. This is to avoid runtimes that are > 4 hours.
+ Added configurable `n_states` 
+ Some linter errors fixed.
+ TODOs added.
+ A follow up PR will address the allocation of compute to the various tests for maximum efficiency.

Addresses #242


josephdviviano:
@alexhernandezgarcia @carriepl 

have a look at this draft - it's not finished nor linted etc, but it shows a working example for the ccrystals and cube environments.

I don't love the code duplication in `common.py` but i'm not sure how better to handle that (I was thinking there is maybe a decorator but I'm not sure how to pass the args to it - anyway I think it's not crucial to fix this now).

alexhernandezgarcia:
> @alexhernandezgarcia @carriepl
> 
> have a look at this draft - it's not finished nor linted etc, but it shows a working example for the ccrystals and cube environments.
> 
> I don't love the code duplication in `common.py` but i'm not sure how better to handle that (I was thinking there is maybe a decorator but I'm not sure how to pass the args to it - anyway I think it's not crucial to fix this now).

The code repetition you refer to is the lines
```
if _get_current_method_name() in self.repeats:
    n_repeat = self.repeats[_get_current_method_name()]
```
in all tests, or something else?

josephdviviano:
Yes exactly. It's a bit ugly to do this, but not a showstopper.

Joseph Viviano
@josephdviviano <https://twitter.com/josephdviviano>
viviano.ca

alexhernandezgarcia:
I agree. I will be able to sleep ok with this :P

tests/gflownet/envs/test_spacegroup.py
```diff
@@ -385,18 +385,34 @@ def test__special_cases_composition_compatibility(n_atoms, cls_idx, ps_idx):
...
+            "test__reset__state_is_source": 10,
+        }
 
-def test__all_common__env_with_composition(env_with_composition):
-    print(
-        f"\n\nCommon tests for SpaceGroup with restrictions from composition {N_ATOMS}\n"
-    )
-    return common.test__all_env_common(env_with_composition)
 
+class TestSpaceGroupWithRestrictedSubgroups(common.BaseTestsDiscrete):
```
alexhernandezgarcia, line 410: Just a rename:
```suggestion
class TestSpaceGroupWithRestrictedSpaceGroups(common.BaseTestsDiscrete):
```

tests/gflownet/envs/test_tests.py
```diff
@@ -0,0 +1,75 @@
...
+    return inspect.currentframe().f_back.f_code.co_name
+
+
+# def repeat_test(times):
+#     def decorator(func):
+#         def wrapper(*args, **kwargs):
+#             for _ in range(times):
+#                 func(*args, **kwargs)
+#         return wrapper
+#     return decorator
```
alexhernandezgarcia, line 15: Do you want to keep these lines commented out?

josephdviviano:
Yeah I'll likely remove this entire file when I go to merge the PR (I need
first to fix the dependencies, or CI will continue to crash).

Joseph Viviano
@josephdviviano <https://twitter.com/josephdviviano>
viviano.ca

alexhernandezgarcia:
> Yeah I'll likely remove this entire file when I go to merge the PR (I need first to fix the dependencies, or CI will continue to crash).

Just as a suggestion, in case it's helpful. If you run it in the Mila cluster, the following should install all you need:
```bash
git fetch origin setupenvs
git show setupenvs:mila/setup/setupenv_full_py3.10_cu11.8_torch2.0.1.sh > .tmp.installgflownet.sh
source .tmp.installgflownet.sh ~/scratch/.virtualenvs/gflownet_full_py3.10_cu11.8_torch2.0.1
source ~/scratch/.virtualenvs/gflownet_full_py3.10_cu11.8_torch2.0.1/bin/activate
rm .tmp.installgflownet.sh
```

josephdviviano:
Thanks but I think we need a solution that will run on an arbitrary machine

Joseph (Mobile)

tests/gflownet/envs/common.py
```diff
@@ -12,47 +13,484 @@
...
+            next_state, action_step, valid = self.env.step(action)
+            if torch.is_tensor(self.env.state):
+                assert self.env.equal(next_state, self.env.state)
+            else:
+                assert next_state == self.env.state
+            assert action_step == action
+            assert valid is False
+
+    def test__sample_actions__backward__returns_eos_if_done(self, n_repeat=1):
+        N = 5
```
alexhernandezgarcia, line 58: Just my picky preference: I would use a more explicit name, like `n_states`

tests/gflownet/envs/common.py
```diff
@@ -12,47 +13,484 @@
...
+                    mask=masks,
+                    states_from=[self.env.state],
+                    is_backward=True,
+                )
+                assert torch.isfinite(logprobs_bw)
+                assert logprobs_bw > -1e6
+                state_prev = copy(state_next)  # TODO: We never use this. Remove?
+
+    def test__backward_actions_have_nonzero_forward_prob(self, n_repeat=1):
+        N = 1000
```
alexhernandezgarcia, line 145: How much work would it be to make `N` (or `n_states`) an argument like `n_repeat`, so that it could be modified in a dictionary passed from each environment's specific class? It seems that it would be equivalent to `n_repeat`, right? Or am I missing something?

alexhernandezgarcia (commented):
Nice work! It generally looks good and it does what we needed! I am currently working on fixing the tests that fail, so I suggest we incorporate those fixes in a different PR and wait for that to merge. How does that sound?

alexhernandezgarcia:
> Thanks but I think we need a solution that will run on an arbitrary machine 

Ok, I thought you were asking about the way for you specifically to install it.

In any case, how is it possible to make it arbitrary if installing torch requires specifying cuda version?

tests/gflownet/envs/test_tests.py
josephdviviano: removed the whole dang file ;)

tests/gflownet/envs/common.py
josephdviviano: done!

tests/gflownet/envs/common.py
josephdviviano: done!

josephdviviano:
All changes incorporated. I'll merge this up, and then we can do a second pass where we tune the number of repeats / states explored for each env.

PR #273, opened 2024-01-12
Fix crystals setup

AlexandraVolokhova:
The setup file for crystals didn't work because it used python 3.8 which is too old for DAVE (it requires >=3.9). I've updated it to 3.9 and now it works for me


alexhernandezgarcia (approved):
Ok

PR #274, opened 2024-01-18
[FIX] Old state conversions still hanging in there

alexhernandezgarcia:
### Tests, black and isort
- [x] `python -m pytest ./tests/`
- [x] `black ./gflownet/`
- [x] `black ./tests/`
- [x] `isort --profile black ./gflownet/`
- [x] `isort --profile black ./tests/`

PR #275, opened 2024-01-19
Crystals sampling scripts

AlexandraVolokhova:
Should be reviewed after merging  #274 
Scripts for sampling crystals from uniform distribution + Alex's scripts for evaluating crystals gflownet. 

I have a hardcoded path [here](https://github.com/alexhernandezgarcia/gflownet/blob/crystals_reward_marginals/scripts/crystal/sample_uniform_with_rewards.py#L56) which is not nice but I don't know how to pass it as an argument to the script as I'm passing the whole hydra config there (same as for the main.py). So, it's not obvious how to add one more parameter there without changing the main config. Would appreciate your suggestion how to improve it.


scripts/crystal/eval_crystalgflownet.py
```diff
@@ -0,0 +1,223 @@
...
+                max_elements=5,
+                max_atoms=config.env.composition_kwargs.max_atoms,
+                max_atom_i=config.env.composition_kwargs.max_atom_i,
+                space_groups=config.env.space_group_kwargs.space_groups_subset,
+                min_length=0.0,
+                max_length=1.0,
+                min_angle=0.0,
+                max_angle=1.0,
+            )
+            energies = env.oracle(env.statebatch2proxy(x_sampled))
```
alexhernandezgarcia, line 198:
```suggestion
            energies = env.oracle(env.states2proxy(x_sampled))
```

PR #276, opened 2024-01-22
Fix and improve backward sampling from data set and replay buffer

alexhernandezgarcia:
This PR unifies the functionality to sample and construct a batch from a buffer (replay buffer or training data set) into a common Buffer method, `select()`.

### Tests
- [x] `python -m pytest ./tests/`

### DIFF
--- a/gflownet/utils/buffer.py
+++ b/gflownet/utils/buffer.py
@@ -273,8 +273,11 @@ class Buffer:
     def select(data_dict: dict, n: int, mode: str = "permutation", rng=None):
         if n == 0:
             return []
-        # TODO: need list()?
         samples = data_dict["x"]
+        # If the data_dict comes from the replay buffer, then samples is a dict and we
+        # need to keep its values only
+        if isinstance(samples, dict):
+            samples = list(samples.values())
         if mode == "permutation":
             assert rng is not None
             samples = [samples[idx] for idx in rng.permutation(n)]
@@ -285,8 +288,11 @@ class Buffer:
                 score = "energy"
             else:
                 raise ValueError(f"Data set does not contain reward or energy key.")
-            # TODO: need fromiter()?
-            scores = np.fromiter(data_dict[score], dtype=float)
+            scores = data_dict[score]
+            # If the data_dict comes from the replay buffer, then scores is a dict and we
+            # need to keep its values only
+            if isinstance(scores, dict):
+                scores = np.fromiter(scores.values(), dtype=float)
             indices = np.random.choice(
                 len(samples),
                 size=n,
--- a/gflownet/utils/buffer.py
+++ b/gflownet/utils/buffer.py
@@ -271,14 +271,22 @@ class Buffer:
 
     @staticmethod
     def select(data_dict: dict, n: int, mode: str = "permutation", rng=None):
+        if n == 0:
+            return []
         # TODO: need list()?
         samples = data_dict["x"]
         if mode == "permutation":
             assert rng is not None
             samples = [samples[idx] for idx in rng.permutation(n)]
-        elif mode in ["energy", "rewards"]:
+        elif mode == "weighted":
+            if "rewards" in data_dict:
+                score = "rewards"
+            elif "energy" in data_dict:
+                score = "energy"
+            else:
+                raise ValueError(f"Data set does not contain reward or energy key.")
             # TODO: need fromiter()?
-            scores = np.fromiter(data_dict[mode], dtype=float)
+            scores = np.fromiter(data_dict[score], dtype=float)
             indices = np.random.choice(
                 len(samples),
                 size=n,
--- a/gflownet/utils/buffer.py
+++ b/gflownet/utils/buffer.py
@@ -1,6 +1,7 @@
 """
 Buffer class to handle train and test data sets, reply buffer, etc.
 """
+
 import pickle
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


gflownet/utils/buffer.py
```diff
@@ -259,11 +259,47 @@ def make_data_set(self, config):
...
+        if isinstance(samples, dict):
+            samples = list(samples.values())
+        if mode == "permutation":
+            assert rng is not None
+            samples = [samples[idx] for idx in rng.permutation(n)]
+        elif mode == "weighted":
+            if "rewards" in data_dict:
+                score = "rewards"
+            elif "energy" in data_dict:
+                score = "energy"
```
vict0rsch, line 288: Is there no scenario in which you want `weighted` and `energy` but the dict contains `rewards` and so it never gets to `score = "energy"`?

gflownet/utils/buffer.py
```diff
@@ -259,11 +259,47 @@ def make_data_set(self, config):
...
+            # need to keep its values only
+            if isinstance(scores, dict):
+                scores = np.fromiter(scores.values(), dtype=float)
+            indices = np.random.choice(
+                len(samples),
+                size=n,
+                replace=False,
+                p=scores / scores.sum(),
+            )
+            samples = [samples[idx] for idx in indices]
```
vict0rsch, line 302: I wondered which of those patterns was the fastest:

```
def f(l, n):
	indices = np.random.choice(len(l), size=n, replace=False)
	return [samples[i] for i in indices]

def g(l, n):
    return np.random.choice(l, size=n, replace=False).tolist()
```

`f` is ~2.5x faster than `g` for `len(l) = 2000` and `n = 1000` (and it gets even faster when `len(l)` increases.

gflownet/utils/buffer.py
```diff
@@ -259,11 +259,47 @@ def make_data_set(self, config):
...
         mean_data = data["energies"].mean()
         std_data = data["energies"].std()
         min_data = data["energies"].min()
         max_data = data["energies"].max()
         data_zscores = (data["energies"] - mean_data) / std_data
         max_norm_data = data_zscores.max()
         return mean_data, std_data, min_data, max_data, max_norm_data
+
+    @staticmethod
+    def select(data_dict: dict, n: int, mode: str = "permutation", rng=None):
```
vict0rsch, line 273: If you're gonna have type hints I would be complete:

* missing `rng`
* missing return `-> list` (and ideally list of what with `List[]`

Also, I think it's time to be a bit more rigorous with docstrings. Copilot / ChatGPT will do a good job at doing a good structure / draft if you find it annoying to do :)

gflownet/utils/buffer.py
```diff
@@ -259,11 +259,47 @@ def make_data_set(self, config):
         )
         return df, {"x": samples, "energy": energies}
 
-    def compute_stats(self, data):
+    @staticmethod
+    def compute_stats(data):
```
vict0rsch, line 263: Same (re: typehints and docstrings)

vict0rsch (changes requested):

gflownet/utils/buffer.py
alexhernandezgarcia: So I went with the faster choice! :v:

gflownet/utils/buffer.py
alexhernandezgarcia: Oh the thing is it will be one thing or the other (`energy` or `rewards`) in the dictionary, depending on whether it's coming from a replay buffer (`rewards`) or a data set (`energy`). And if it happened to have both, it's fine to use the rewards if they are available.

gflownet/utils/buffer.py
alexhernandezgarcia: By "missing rng" you mean the type is missing?

alexhernandezgarcia: Added docstring. Thanks for the reminder :smiling_face_with_tear:

gflownet/utils/buffer.py
alexhernandezgarcia: I will not bother with this one because it may change soon

gflownet/utils/buffer.py
vict0rsch: SGTM

gflownet/utils/buffer.py
```diff
@@ -259,11 +260,79 @@ def make_data_set(self, config):
...
+        n : int
+            The number of samples to select from the dictionary.
+
+        mode : str
+            Sampling mode. Options: permutation, weighted.
+
+        rng : np.random.Generator
+            A numpy random number generator, used for the permutation mode. Ignored
+            otherwise.
+        """
```
vict0rsch, line 306: Missing return :) 

```
Returns
-------
int
    Description of anonymous integer return value.
```
https://numpydoc.readthedocs.io/en/latest/format.html#returns

alexhernandezgarcia: docstring police :P 

done here https://github.com/alexhernandezgarcia/gflownet/pull/276/commits/b9ffd7405c35d4e73ac57d631bf0875b63e5196c

alexhernandezgarcia:
No explicit approval from @vict0rsch but all comments have been addressed. In the interest of time, I will bypass protection and merge.

vict0rsch:
Sorry, LGTM

PR #283, opened 2024-02-09
Apply new black formatting

alexhernandezgarcia:
It seems like a recent upgrade of Black has updated a few small things. This PR applies these changes. Or am I missing something?


josephdviviano (approved):
lgtm!

josephdviviano:
Not sure what this is all about - but it seems fine. Black does change slightly over time, which can be annoying for breaking CI ;).

PR #284, opened 2024-02-09
Various small improvements

alexhernandezgarcia:
Just two small updates:

- Allow the proxy format to be a list too (not only a tensor)
- Small fix in the common tests to avoid issues with non-copied states.

Note: these changes are applied on top of PR https://github.com/alexhernandezgarcia/gflownet/pull/283 (to include the updated black formatting)

### Tests, black and isort
- [x] `python -m pytest ./tests/` (locally) 
- [x] `black ./gflownet/`
- [x] `black ./tests/`
- [x] `isort --profile black ./gflownet/`
- [x] `isort --profile black ./tests/`

### DIFF
--- a/tests/gflownet/envs/common.py
+++ b/tests/gflownet/envs/common.py
@@ -392,7 +392,7 @@ def test__trajectories_are_reversible(env):
     while not env.done:
         state, action, valid = env.step_random(backward=False)
         if valid:
-            states_trajectory_fw.append(state)
+            states_trajectory_fw.append(copy(state))
             actions_trajectory_fw.append(action)
 
     # Sample backward trajectory with actions in forward trajectory
@@ -402,7 +402,7 @@ def test__trajectories_are_reversible(env):
     while not env.equal(env.state, env.source) or env.done:
         state, action, valid = env.step_backwards(actions_trajectory_fw_copy.pop())
         if valid:
-            states_trajectory_bw.append(state)
+            states_trajectory_bw.append(copy(state))
             actions_trajectory_bw.append(action)
 
     assert all(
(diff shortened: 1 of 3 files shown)


tests/gflownet/envs/common.py
```diff
@@ -392,7 +392,7 @@ def test__trajectories_are_reversible(env):
     while not env.done:
         state, action, valid = env.step_random(backward=False)
         if valid:
-            states_trajectory_fw.append(state)
+            states_trajectory_fw.append(copy(state))
```
josephdviviano, line 395: What's going on with this copy? Maybe worth a comment (I assume it is so the state does not get changed in memory).

If so, are you sure this is working as intended, and should not instead be a deepcopy?

josephdviviano (approved):
lgtm! I left one comment - I'll let you merge.

tests/gflownet/envs/common.py
alexhernandezgarcia: Uhm I'm never 100 % sure with these things, but I think so. We want to store the different states of a trajectory in a list. I think most environments copy the state before applying the step, which is why this has always worked as expected. But it should not be mandatory for the step method to copy the state. If that doesn't happen, then this test method should do the copy to save what we want without risking it will be changed.

Now, about deepcopy. The `copy()` method here refers to [this method](https://github.com/alexhernandezgarcia/gflownet/blob/misc/gflownet/utils/common.py#L272) from the common utils. We're copying a state, so that should be fine, no? No need for a deep copy, I think.

alexhernandezgarcia:
The changes in this PR are already included (or further updated) in PR https://github.com/alexhernandezgarcia/gflownet/pull/285. Closing this one.

PR #286, opened 2024-02-12
pyshtools==4.10.3 in toml to fix BLAS issue in installation

alexhernandezgarcia:
The following commands, in the present branch, seem to install the Python environment in the Mila cluster without issues:
```
module --force purge
module load python/3.10
module load cuda/11.8
python -m virtualenv ~/tmpenv
source ~/tmpenv/bin/activate
source ./prereq_python.sh
source ./prereq_geometric.sh
python -m pip install -e .[all]
```


josephdviviano (approved):
lgtm!

PR #287, opened 2024-02-15
[WIP] Refactor & standardize evaluation with `Evaluator`

vict0rsch:
* [x] remove test logic from main `GFlowNetAgent`
* [x] rename to `eval()` logic to be semantically distinct from (unit, integration etc.)-tests
* [x] clean up evaluation return values (to dict instead of messy tuple) 
* [ ] fix `logger` vs `evaluator` roles
* [ ] isolate plotting logic

WIP questions are marked with `#TODO-V:` in the code, will be turned into actual PR comments when needed but you can already have a look and think.


gflownet/gflownet.py
```diff
@@ -57,6 +53,7 @@ def __init__(
         sample_only=False,
         replay_sampling="permutation",
         train_sampling="permutation",
+        eval_config=None,
         **kwargs,
     ):
         # Seed
```
vict0rsch, line 59: Should we remove references to `oracle` across the code? Everywhere I look it seems deprecated / unused.

vict0rsch, line 59: Why is this used beyond logger's `use_context` ? -> modify docstring if it's for more things, or modify argument if it's just to set `use_context`

vict0rsch, line 59: Deprecated? Looks only used in the `Buffer`'s creation but `Buffer.__init__` does not use it at all

gflownet/evaluator/base.py
```diff
@@ -0,0 +1,635 @@
...
+                    pickle.dump(dict_tt, f)
+            # Estimate pred log density using test samples
+            # TODO: this may be specific-ish for the torus or not
+            scores_pred = kde_pred.score_samples(x_tt)
+            log_density_pred = scores_pred - logsumexp(scores_pred, axis=0)
+            density_true = np.exp(log_density_true)
+            density_pred = np.exp(log_density_pred)
+        else:
+            # TODO: refactor
+            env_metrics = gfn.env.test(x_sampled)
```
vict0rsch, line 421: Deprecated?

gflownet/gflownet.py
alexhernandezgarcia: Yes, we should. We could decide later on to bring it back at some point, but that decision would imply wide changes. So yes, we can safely and should remove all references to oracles.

alexhernandezgarcia: Not sure right now to be honest. I need to look into it more carefully.

alexhernandezgarcia: I am actually confused about what this refers to. Is your question about `eval_config`?

gflownet/evaluator/base.py
alexhernandezgarcia: Yes, this will be re-done in a better way.

gflownet/gflownet.py
vict0rsch: Sounds good

vict0rsch: It has moved unfortunately... It was referring to `sample_only`

gflownet/evaluator/base.py
vict0rsch: Good thanks

gflownet/utils/logger.py
```diff
@@ -301,6 +269,7 @@ def log_sampler_test(
             )
 
     def log_sampler_oracle(self, energies: array, step: int, use_context: bool):
+        # TODO-V -> remove? Unused
         if not self.do.online:
```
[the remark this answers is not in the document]
vict0rsch: Yes

gflownet/gflownet.py
vict0rsch: I'm not sure the comment was clear, I meant the `active_learning` argument

vict0rsch: Do you also want to remove:

1. `gflownet/oracle/*`
2. `scripts/oracle_annotate.py`
3. `scripts/oracle_sampler.py`
4. Is `energies = env.oracle(x_sampled)` in `eval_gflownet.py` still relevant?

vict0rsch:
@alexhernandezgarcia what was the reason for storing metrics as attributes to the `GFlowNetAgent`? Can I safely remove this procedure from the evaluation / logging logic? Or would there be some side effects somewhere in your opinion? Looks ok from the code but making sure with you

alexhernandezgarcia:
> @alexhernandezgarcia what was the reason for storing metrics as attributes to the `GFlowNetAgent`? Can I safely remove this procedure from the evaluation / logging logic? Or would there be some side effects somewhere in your opinion? Looks ok from the code but making sure with you

No important reason. I think Nikita started it that way at some point and we just never followed it up. Like I said, we should not feel restricted by the way things are currently done.

gflownet/gflownet.py
alexhernandezgarcia: Oh I see! Yeah `sample_only` is very much deprecated.

alexhernandezgarcia: Screenshot from 2024-02-19 15-03-16
GitHub can definitely improve this...

The `active_learning` flag is technically not deprecated but definitely not well maintained. So we can remove it too and deal with this later.

alexhernandezgarcia: 1-3 can go.

4 is actually wrong :/ and I thought it had been corrected but I guess it is corrected only in another fork... Keep this and I will take care of this myself. The entire script needs to be updated.

gflownet/evaluator/base.py
```diff
@@ -0,0 +1,793 @@
...
+
+from gflownet.utils.batch import Batch
+from gflownet.utils.common import (
+    batch_with_rest,
+    load_gflow_net_from_run_path,
+    tfloat,
+    torch2np,
+)
+
+_sentinel = object()
```
alexhernandezgarcia, line 20: What is the purpose or role of this?

vict0rsch: Typically, a "sentinel" is used to allow arbitrary user inputs.

The use-case is: if you have a function that can be used with any input type, how do you know if the user has provided an argument or not?

Example:

```
def f(a=None):
    print("User wants:", a)

f(a)
f()
```

alexhernandezgarcia: I see! Thanks for explaining that!

gflownet/gflownet.py
vict0rsch: Done in [`82cc7ec` (#287)](https://github.com/alexhernandezgarcia/gflownet/pull/287/commits/82cc7ecb77532a4e94f8d805089b72c91e6b53ea)

NB: I also removed `legacy.py`. There are also a few remaining `oracle` references, for instance `proxy_state_format: oracle`.

vict0rsch: I'll let you resolve when you've read it

vict0rsch: Ok. Done in [`dbe2e51` (#287)](https://github.com/alexhernandezgarcia/gflownet/pull/287/commits/dbe2e51869146a7764f8c233375e7d35f217aa05)

vict0rsch: Done in [`4beae50` (#287)](https://github.com/alexhernandezgarcia/gflownet/pull/287/commits/4beae5045db277d424a641da68454d1dc74e7cb6)

gflownet/evaluator/base.py
vict0rsch: (and leading `_` to signal it's private and should not be used by a user of this library. The only place outside the code-base it can be --and is-- used is in tests)

vict0rsch:
Controversial change @alexhernandezgarcia @michalkoziarski @carriepl @josephdviviano thoughts?

[`5581e71` (#287)](https://github.com/alexhernandezgarcia/gflownet/pull/287/commits/5581e71ad8d54aa2b8300e8d282a0bf1e59a50c5)

If you agree this should be used in the `eval_gflownet.py` script and in `tests.env.common::BaseTestsCommon.test__gflownet_minimal_runs`

gflownet/gflownet.py
alexhernandezgarcia: Done in [this PR](https://github.com/alexhernandezgarcia/gflownet/pull/290). Feel free to review that (very small), merge and also merge that branch into this one to get the changes. Agree?

gflownet/gflownet.py
```diff
@@ -1081,7 +1101,7 @@ def train(self):
             # Log
             if self.logger.lightweight:
                 all_losses = all_losses[-100:]
-                all_visited = states_term
+                all_visited = states_term  # TODO-V: unused
```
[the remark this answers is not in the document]
alexhernandezgarcia: Yes unused because it is not maintained. But computing metrics from the visited state is a nice to-have. It should be done with the next version of the Buffer though, not here to not clutter too much the PR. 

So: can remove

alexhernandezgarcia:
> Controversial change @alexhernandezgarcia @michalkoziarski @carriepl @josephdviviano @AlexandraVolokhova thoughts?
> 
> [`5581e71` (#287)](https://github.com/alexhernandezgarcia/gflownet/pull/287/commits/5581e71ad8d54aa2b8300e8d282a0bf1e59a50c5)
> 
> If you agree this should be used in the `eval_gflownet.py` script and in `tests.env.common::BaseTestsCommon.test__gflownet_minimal_runs`

I am not against using a method to hide all the complexity in the tests or in the evaluation script. However, I think it is good that `main.py` shows explicitly the components that are needed and what depends on what.

Also, as a separate comments: I would try if possible to not include too many changes in a single PR - especially if they are out of the scope of the PR. In other words, I would try to spin off changes like this one in a separate PR. It's just to make the review process a tad easier.

vict0rsch:
> > Controversial change @alexhernandezgarcia @michalkoziarski @carriepl @josephdviviano @AlexandraVolokhova thoughts?
> > [`5581e71` (#287)](https://github.com/alexhernandezgarcia/gflownet/pull/287/commits/5581e71ad8d54aa2b8300e8d282a0bf1e59a50c5)
> > If you agree this should be used in the `eval_gflownet.py` script and in `tests.env.common::BaseTestsCommon.test__gflownet_minimal_runs`
> 
> I am not against using a method to hide all the complexity in the tests or in the evaluation script. However, I think it is good that `main.py` shows explicitly the components that are needed and what depends on what.
> 
> Also, as a separate comments: I would try if possible to not include too many changes in a single PR - especially if they are out of the scope of the PR. In other words, I would try to spin off changes like this one in a separate PR. It's just to make the review process a tad easier.

Sure I'll move that to another PR to merge after this one then

alexhernandezgarcia:
> @alexhernandezgarcia I'm following those steps https://stackoverflow.com/a/30893291/3867406 -> have you pulled or fetched from THIS branch (`evaluator`)?
> 
> If I edit the commit history on MY machine then FORCE push to Github but you have a local version of `evaluator` that will have diverged we may end up blowing everything up, in which case I suggest we just keep the offending commit ([`5581e71` (#287)](https://github.com/alexhernandezgarcia/gflownet/pull/287/commits/5581e71ad8d54aa2b8300e8d282a0bf1e59a50c5))

You can go ahead without impact on my work / local copies.

vict0rsch:
> > Controversial change @alexhernandezgarcia @michalkoziarski @carriepl @josephdviviano @AlexandraVolokhova thoughts?
> > [`5581e71` (#287)](https://github.com/alexhernandezgarcia/gflownet/pull/287/commits/5581e71ad8d54aa2b8300e8d282a0bf1e59a50c5)
> > If you agree this should be used in the `eval_gflownet.py` script and in `tests.env.common::BaseTestsCommon.test__gflownet_minimal_runs`
> 
> I can't see what this controversial change is

Great, that means I reverted the commit appropriately :p 

I think we should not have multiple places in the code that instantiate a `GFlowNetAgent` from a Hydra config. That's why I created `gflownet.utils.common::gflownet_from_config()`. I was suggesting to use it in `main.py` but Alex suggested we don't do it in this PR and mix things up (and seems against it for `main.py` though I think I can convince him 😄)

josephdviviano:
Without total context I agree with @vict0rsch - this sounds like something that should be one and done.

gflownet/evaluator/base.py
```diff
@@ -0,0 +1,564 @@
...
+from scipy.special import logsumexp
+
+from gflownet.evaluator.abstract import ALL_REQS  # noqa
+from gflownet.evaluator.abstract import METRICS  # noqa
+from gflownet.evaluator.abstract import AbstractEvaluator
+from gflownet.utils.batch import Batch
+from gflownet.utils.common import batch_with_rest, tfloat, torch2np
+
+
+class BaseEvaluator(AbstractEvaluator):
```
josephdviviano, line 36: What is the purpose of having a distinct Base and Abstract evaluator? Typically they are one and the same.

Is the idea that the `Base` evaluator is a actually `Default` evaluator, and in future scenarios you imagine someone would need to build some different Evaluator from `Abstract`, but it wouldn't be useful to start by inheriting `Base`?

josephdviviano (commented):
my 2c - looks good. I'm wondering what elements in the `evaluator.base` class are going to be easy for future developers to modify since everything is fairly tied together.

gflownet/evaluator/base.py
vict0rsch: Yes that's it. the nomenclature (`Default` vs `Base`) is just inherited from the rest of the code base and I did contemplate naming it `DefaultEvaluator` instead

vict0rsch: I imagine most would inherit `Base` not `Abstract` when they do regular stuff but if `Base` does things the wrong way for their use-case (or even too much) they they can just inherit from `Abstract`.

The only requirements are those defined in `Abstract`* because they are exposed to the rest of the code, the methods extending classes (including `Base`) should/can implement are for internal uses.

\* -> `eval_and_log()`, `should_*()`

gflownet/evaluator/__init__.py
```diff
@@ -0,0 +1,340 @@
+"""
+An ``Evaluator`` is a class that is used to compute metrics and plots.
```
alexhernandezgarcia, line 2:
```suggestion
An ``Evaluator`` is a class that is used to compute metrics and generate plots.
```

gflownet/evaluator/__init__.py
```diff
@@ -0,0 +1,340 @@
...
+It serves two complementary purposes:
+
+1. It is used to evaluate the performance of the agent during training and to log the
```
alexhernandezgarcia, line 5:
```suggestion
1. Evaluate the performance of the agent during training and to log the
```

gflownet/evaluator/__init__.py
```diff
@@ -0,0 +1,340 @@
...
+   results.
+2. It is intended to be used to evaluate the performance of a trained agent, from a
```
alexhernandezgarcia, line 7:
```suggestion
2. Evaluate the performance of a trained agent, from a
```

gflownet/evaluator/__init__.py
alexhernandezgarcia: Take all these suggestions in the text truly as suggestions (I am not implying the current version is wrong at all)

gflownet/evaluator/__init__.py
```diff
@@ -0,0 +1,340 @@
...
+
+.. important::
+
+    Prefer the :meth:`~gflownet.evaluator.abstract.AbstractEvaluator.from_dir`
+    and :meth:`~gflownet.evaluator.abstract.AbstractEvaluator.from_agent`
+    class methods to instantiate an evaluator.
+
+Typical call stack:
+
+1. :meth:`gflownet.gflownet.GFlowNetAgent.train` calls the evaluator's
```
alexhernandezgarcia, line 29: Is something missing at the end of the sentence? "evaluator's sth"?

gflownet/evaluator/__init__.py
```diff
@@ -0,0 +1,340 @@
...
+   logs the results using the GFlowNetAgent's logger as
+   ``self.logger.log_metrics(results["metrics"])`` and ``self.logger.log_plots(figs)``.
+
+
+.. _evaluator basic concepts:
+
+Basic concepts
+--------------
+
+The evaluator is used to compute metrics and plots. It is used to evaluate the
```
alexhernandezgarcia, line 51:
```suggestion
The evaluator is used to compute metrics and generate plots. It is used to evaluate the
```

gflownet/evaluator/__init__.py
```diff
@@ -0,0 +1,340 @@
...
+    from gflownet.evaluator.base import BaseEvaluator
+
+    gfn_run_dir = "PUT_YOUR_RUN_DIR_HERE"  # a run dir contains a .hydra folder
+    gfne = BaseEvaluator.from_dir(gfn_run_dir)
+    results = gfne.eval()
+
+    for name, metric in results["metrics"].items():
+        print(f"{name:20}: {metric:.4f}")
+
+    data = results.get("data", {})
```
alexhernandezgarcia, line 101: Note to self: understand why the argument `{}` is needed.

gflownet/evaluator/base.py
alexhernandezgarcia: We can totally rename stuff in the rest of the codebase if there would be better choices.

config/experiments/scrabble/jay.yaml
```diff
@@ -51,21 +52,21 @@ policy:
     shared_weights: False
     checkpoint: backward
 
+# Evaluator
```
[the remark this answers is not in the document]
alexhernandezgarcia: I am unsure what this comment refers to exactly, but I would just say that the icml23/ctorus.yaml file is really old (January 2023) so it would be fine to deprecate it / adapt it if needed. Yes, it contains the experiments of a paper, but I believe it's ok to adapt it to the new state of the repo.

main.py
```diff
@@ -28,9 +29,10 @@ def main(config):
     # Set other random seeds
     set_seeds(config.seed)
 
+    # Initialize GFlowNet from config
     # Logger
     logger = hydra.utils.instantiate(config.logger, config, _recursive_=False)
-    # The proxy is required in the env for scoring: might be an oracle or a model
+    # The proxy is required in the env for scoring
```
alexhernandezgarcia, line 35:
```suggestion
    # The proxy is required by the GFlowNet agent to compute rewards
```

main.py
```diff
@@ -91,11 +97,11 @@ def main(config):
     if config.n_samples > 0 and config.n_samples <= 1e5:
         batch, times = gflownet.sample_batch(n_forward=config.n_samples, train=False)
         x_sampled = batch.get_terminating_states(proxy=True)
-        energies = env.proxy(x_sampled)
+        energies = gflownet.env.proxy(x_sampled)
```
alexhernandezgarcia, line 100: This will result in a conflict.

main.py
carriepl: Done

main.py
carriepl: Done. It was fixed locally and I just pushed the changes.

alexhernandezgarcia:
Thanks Pierre Luc!
>     * I have not done the changes to the logger outlined by @vict0rsch. That will be the next step.
Could this be done in a new PR or should it be done before merging?

carriepl:
Interesting... the CI is currently failing because of a test that was passing before my last commit which is only changing a commit. I guess that this is a test that fails very infrequently. At this point, I don't think it's related to this PR but I could be wrong.

alexhernandezgarcia:
> Interesting... the CI is currently failing because of a test that was passing before my last commit which is only changing a commit. I guess that this is a test that fails very infrequently. At this point, I don't think it's related to this PR but I could be wrong.

Don't worry. I am pretty sure this is related to tests of the Batch class that in this still branch use `torch.equal` but it should be fine with `torch.isclose`. [All this has been changed in the famous big PR](https://github.com/alexhernandezgarcia/gflownet/pull/299/files#diff-1362e42cbf7b912b8466af920b7e9b15699ad8ab3040bad8d4ce1aedae048e48).

alexhernandezgarcia (approved):
I had to make a few additional changes after checking the sanity runs (Tetris topK figures were missing). The solution is a bit of a quick fix. 

I have realised that a bunch of things will need more work, but this is a great step forwards since the Evaluator will give us the flexibility to extend the evaluation without having to do ugly additions to the former test() function of the GFN.

I have added a couple of quick issues about things that are needed as a reminder.

Great work everyone!!! I will merge.

PR #291, opened 2024-02-21
Fix docs requirements for auto build

vict0rsch:
Docs for this branch -> https://gflownet.readthedocs.io/en/fix-docs-reqs/

Also

* Standardize docstrings to Numpy style
* Update documentation instructions
* Add an example function for display only in `common.py`


josephdviviano (approved):
lgtm

PR #296, opened 2024-03-01
Tiny bug fix in bootstrap 

AlexandraVolokhova:
Moved indices tensor to the same device as `tensor`. It failed on gpu before, now works well


alexhernandezgarcia (approved):
Looks good!

PR #297, opened 2024-03-04
Branin and Hartmann proxies

alexhernandezgarcia:
This PR implements the Branin and Hartmann proxies, functions widely used in the optimisation and (multi-fidelity) active learning literature, including the [Multi-Fidelity Active Learning with GFlowNets](https://arxiv.org/abs/2306.11715) paper.

The Branin function is defined on the real plane and the Hartmann function on a 6D real hyper-plane. Therefore, these functions (proxies) can be used to train both grid (discrete) and cube (continuous) environments.

They rely on the [botorch](https://botorch.org/) implementation. Therefore `botorch` has been added as a new dependency.

Both proxies can be turned into multi-fidelity oracles via the parameter `fidelity`.

This PR includes a bit of additional clean-up and refactoring:
- Remove the deprecated variable `func` from environment config files.
- Remove the variable `higher_is_better` from proxy config files.
- Remove variable `proxy_factor` in the base environment, since it is assumed that proxy outputs will be negative and lower is better. If not, this can still be controlled with the beta parameter. In any case, this aspect should be revised in the future.
- Move all grid and cube proxies into `/gflownet/proxy/box` and similarly for the configuration files.
- Extends docstring.


pyproject.toml
alexhernandezgarcia, line 1: @josephdviviano this PR adds two new proxy classes that make use of a Python library that wasn't yet a dependency, [botorch](https://botorch.org/). Is this the correct way of adding it as a dependency? Anything else needs to be done? Thanks!

josephdviviano:
Yes that works, will be part of the base install.

Joseph (Mobile)

PR #310, opened 2024-05-30
[Fix bug] [One-line PR] Fix call of `proxy.get_max_reward()`

alexhernandezgarcia:
I don't know how this escaped my sanity check runs but it did.


carriepl (approved):
Yeahhhh.... we're going to need additional unit tests for this...

Just kidding! Gotta love a PR that only takes 3 seconds to review.

PR #315, opened 2024-06-05
Changes to handling of log directory

alexhernandezgarcia:
Summary of changes
- Removal of quick fix about race conditions: before, a random directory was created for each run. Now, by default, the directory created by Hydra with the date and time in the name contains the microseconds, which would prevent the potential issues with race conditions.
- A few default configurations have been changed:
  - TB loss instead of FM (because FM does not work for cont. environments)
  - Uniform proxy by default (because it works for all environments)
  - Hydra will not change directory by default.
- I have remove "alex" as the default user in the experiment config files and removed "debug" from the logdir path.

### DIFF
--- a/config/main.yaml
+++ b/config/main.yaml
@@ -1,14 +1,14 @@
 defaults:
   - _self_
   - env: grid
-  - gflownet: flowmatch
+  - gflownet: trajectorybalance
   - policy: mlp_${gflownet}
-  - proxy: corners
+  - proxy: uniform
   - logger: wandb
   - user: default
 
 # Device
-device: cuda
+device: cpu
 # Float precision
 float_precision: 32
 # Number of objects to sample at the end of training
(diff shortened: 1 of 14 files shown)


config/main.yaml
```diff
@@ -1,14 +1,14 @@
...
+  - gflownet: trajectorybalance
   - policy: mlp_${gflownet}
-  - proxy: corners
+  - proxy: uniform
   - logger: wandb
   - user: default
 
 # Device
-device: cuda
+device: cpu
```
michalkoziarski, line 11: I would suggest adding somewhere (e.g. README, very early on in description of how to run code) information that CUDA is NOT the default - I understand why it's the case (CPU being faster on small tasks), but I think the expected thing is to use GPU by default if available.

config/main.yaml
```diff
@@ -19,11 +19,12 @@ seed: 0
 # Hydra config
 hydra:
   # See: https://hydra.cc/docs/configure_hydra/workdir/
+  # See: https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes
   run:
-    dir: ${user.logdir.root}/${now:%Y-%m-%d_%H-%M-%S}
+    dir: ${user.logdir.root}/${oc.env:SLURM_JOB_ID,local}/${now:%Y-%m-%d_%H-%M-%S_%f}
```
michalkoziarski, line 24: So how exactly does this work if not using Slurm? SLURM_JOB_ID will just be an empty string, or will there be a crash with undefined variable?

config/user/default.yaml
```diff
@@ -1,5 +1,5 @@
 logdir:
-  root: ~/gflownet/logs
+  root: ./logs
```
michalkoziarski, line 2: Just a comment, I think it's good enough, but I believe `.` is (at least in some settings) dependent on the directory from which you run the script, not relative to the file path; so if someone were to run from a different directory using `/complicated/path/to/their/gflownet/repo`, it might create `logs` directory relative to this (which might or might not be expected). In python at least, what I'd typically do is use the path relative to the file (e.g. `Path(__file__).parent`).

Again, I think this is fine, just FYI.

michalkoziarski (commented):
I believe this should be fine, but because you added SLURM_ID to the log dir; I am not totally sure if microseconds alone would be enough (no idea how precise Slurm is in starting jobs).

Added two discussion-type comments and one question that I think is important (on whether this works fine locally).

config/user/default.yaml
alexhernandezgarcia: That's the intended behaviour indeed. I think it's fine enough for a default value, which should be overwritten by the specific  user settings in practice. It was problematic before because it was creating a directory called `~`, which isn't fun...

config/main.yaml
alexhernandezgarcia: Ok, added: https://github.com/alexhernandezgarcia/gflownet/pull/315/commits/00cb7e129dadd283619b17fa8d59877317a97fb3

alexhernandezgarcia:
@michalkoziarski approved?

PR #316, opened 2024-06-05
[Tiny PR] [Fix] Use random number generator in weighted mode of select

alexhernandezgarcia:
Unless I am missing something, using `np.random.choice()` instead of `rng.choice()` is contributing to random differences across runs, which has an impact on reproducibility (and debugging).


carriepl (approved):
Looks good to me. Thanks a lot!

PR #322, opened 2024-06-11
[Small PR] Logging dir and buffer outputs: fix issue and improvements

alexhernandezgarcia:
Note for reviewers: There are many files changed but it's mostly the same change applied to many config files.

This PR fixes an issue introduced in a recent PR and makes various changes to hopefully improve the handling of the logging directory and the buffer output files:
- Now Hydra does not change to the output directory by default, so I had to add a couple of lines in `main.py` to make the actual logging directory be Hydra's output directory. Let's call this directory `/log/dir/`, which will be stored in `config.logger.logdir.path`
- The Logger now creates a data directory in `/log/dir/data/` (as well as a /log/dir/ckpts/`) meant to store the replay buffer, the train data and the test data.
- The Buffer will thus not create files anymore in the source code directory.
- I have removed the need to specify `env.buffer.train.output_csv` and `env.buffer.train.output_pkl` (and the test counterparts) in order to make use of the buffer. This simplifies its use and the code. And the replay, train and test output files will always be stored simply as `/log/dir/data/replay.pkl`, `/log/dir/data/train.csv`, `/log/dir/data/train.pkl`, etc.

### DIFF
--- a/gflownet/utils/buffer.py
+++ b/gflownet/utils/buffer.py
@@ -22,7 +22,6 @@ class Buffer:
         env,
         proxy,
         replay_capacity=0,
-        output_csv=None,
         data_path=None,
         train=None,
         test=None,
@@ -59,19 +58,11 @@ class Buffer:
         else:
             self.train_type = None
         self.train, dict_tr = self.make_data_set(train)
-        if (
-            self.train is not None
-            and "output_csv" in train
-            and train.output_csv is not None
-        ):
-            self.train_csv = self.datadir / train.output_csv
+        if self.train is not None:
+            self.train_csv = self.datadir / "train.csv"
             self.train.to_csv(self.train_csv)
-        if (
-            dict_tr is not None
-            and "output_pkl" in train
-            and train.output_pkl is not None
-        ):
-            self.train_pkl = self.datadir / train.output_pkl
+        if dict_tr is not None:
+            self.train_pkl = self.datadir / "train.pkl"
             with open(self.train_pkl, "wb") as f:
                 pickle.dump(dict_tr, f)
         else:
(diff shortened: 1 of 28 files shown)


gflownet/utils/buffer.py
```diff
@@ -21,14 +22,13 @@ def __init__(
...
         replay_capacity=0,
-        output_csv=None,
         data_path=None,
         train=None,
         test=None,
         logger=None,
         **kwargs,
     ):
-        self.logger = logger
+        self.datadir = logger.datadir
```
michalkoziarski, line 31: `logger` defaults to `None`, if this is allowed it should be handled here; otherwise make it required?

gflownet/utils/logger.py
```diff
@@ -71,14 +71,21 @@ def __init__(
         self.lightweight = lightweight
         self.debug = debug
         # Log directory
-        self.logdir = Path(logdir.root)
+        if "path" in logdir:
+            self.logdir = Path(logdir.path)
+        else:
+            self.logdir = Path(logdir.root)
```
michalkoziarski, line 77: Why is this needed? It seems very arbitrary.

alexhernandezgarcia: Not sure if it's strictly needed but it's additional flexibility. Currently, `logdir.path` is created in `main.py`, but this will make it still work if the code is run from a different script that does not do that.

gflownet/utils/buffer.py
alexhernandezgarcia: Good point, done here: https://github.com/alexhernandezgarcia/gflownet/pull/322/commits/f8c235ce8d9d3a15b6b22646ccd40ab61d58010a

michalkoziarski (approved):

PR #327, opened 2024-06-18
[WIP] Make adding new Policy Models flexible

### LINKED ISSUE #293: Flexible Policy Definition
- Policies were originally MLPs.
- Now we need to be able to use arbitrary function approximators.
- This will be a library-level change that will affect all projects.

engmubarak48:
Fixes #293 

This PR tries to make adding new function approximations (policy models) as flexible as possible.

- Modification of the `base.py` file to dynamically import and instantiate policy models based on configuration
- Allowing the addition of new model types (e.g., CNN, GNN, Transformer) without altering the core logic.
- Implementation of a clean interface in the `ModelBase` class


josephdviviano (approved):
lgtm - a comment, but non blocking.

gflownet/policy/mlp.py
```diff
@@ -0,0 +1,60 @@
+# policy_models/mlp_policy.py
+
+import torch
+from torch import nn
+from gflownet.policy.base import ModelBase
+
+
+class MLPPolicy(ModelBase):
+    def __init__(self, config, env, device, float_precision, base=None):
```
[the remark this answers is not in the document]
engmubarak48: Hi @josephdviviano thanks, I just started.. I will let you know when it is ready for review :)

alexhernandezgarcia:
Great that you're taking a stab at this, which is one of the important things that remain to be done! Looks good so far!

gflownet/envs/tetris.py
```diff
@@ -307,7 +309,9 @@ def states2policy(
         A tensor containing all the states in the batch.
         """
         states = tint(states, device=self.device, int_type=self.int)
-        return self.states2proxy(states).flatten(start_dim=1).to(self.float)
+        if self.flatten:
+            return self.states2proxy(states).flatten(start_dim=1).to(self.float)
+        return self.states2proxy(states).to(self.float)
```
engmubarak48, line 314: @alexhernandezgarcia This is a temporary solution to make the CNN policy work on Tetris env. But normally the flattening should happen inside the model but not in the environment (see my other comments)

if you are okay with that, then I can update.

config/env/tetris.yaml
```diff
@@ -11,6 +11,8 @@ height: 20
 pieces: ["I", "J", "L", "O", "S", "T", "Z"]
 # Allowed roations
 rotations: [0, 90, 180, 270]
+# Don't flatten if using CNN
+flatten: True
```
engmubarak48, line 15: If we move the flattening from the environment to the policy, then we don't need this, and we can simply remove it.

gflownet/envs/tetris.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         height: int = 20,
         pieces: List = ["I", "J", "L", "O", "S", "T", "Z"],
         rotations: List = [0, 90, 180, 270],
+        flatten: bool = True,
```
engmubarak48, line 78: If we move the flattening from the environment to the policy, then we don't need this.

gflownet/policy/mlp.py
```diff
@@ -0,0 +1,78 @@
...
+        self.shared_weights = config.get("shared_weights", False)
+        self.n_hid = config.get("n_hid", 128)
+        self.n_layers = config.get("n_layers", 2)
+        self.tail = config.get("tail", [])
+        self.reload_ckpt = config.get("reload_ckpt", False)
+
+    def instantiate(self):
+        self.model = self.make_mlp(nn.LeakyReLU()).to(self.device)
+
+    def __call__(self, states):
```
engmubarak48, line 77: flattening could happen here or we could add nn.flatten to the model before linear layer. See CNN policy model

engmubarak48:
I think any policy could be added now.. i have tested tetris on CNN and everything seems to work well with minimal changes.. I think next thing we could do is have another PR which better documents this and show how one could simply change the Policy etc. 

test with `python main.py gflownet=trajectorybalance env=tetris policy=cnn`
 make sure to set `flatten=False` in the environment. 
@alexhernandezgarcia feel free to view..

gflownet/policy/cnn.py
```diff
@@ -0,0 +1,91 @@
...
+            raise RuntimeError(
+                "Failed during convolution operation. Ensure that the kernel sizes and strides are appropriate for the input dimensions."
+            ) from e
+
+        model = nn.Sequential(
+            conv_module, nn.Flatten(), nn.Linear(in_channels, self.output_dim)
+        )
+        return model.to(self.device)
+
+    def parse_config(self, config):
```
AlexandraVolokhova, line 75: I think here `super().parse_config` should be called to parse the common arguments such as `self.type` and `self.checkpoint`

gflownet/policy/cnn.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, config, env, device, float_precision, base=None):
+        self.env = env
+        super().__init__(
+            config=config,
+            env=env,
+            device=device,
+            float_precision=float_precision,
+            base=base,
+        )
+        self.is_model = True
```
AlexandraVolokhova, line 18: If I understand correctly,`self.is_model` should be True only if the policy type is neither fixed or uniform. So, maybe it's better to move `self.is_model = True` to the instantiate method after the cnn model is created

gflownet/policy/base.py
```diff
@@ -2,12 +2,11 @@
 
 import torch
 from omegaconf import OmegaConf
-from torch import nn
 
 from gflownet.utils.common import set_device, set_float_precision
 
 
-class ModelBase(ABC):
+class Policy(ABC):
```
AlexandraVolokhova, line 9: It's not very crucial, but I'd suggest to call it `BasePolicy` to highlight that it is the base class to inherit from. Also, I think it's not needed anymore to inherit from `ABC` here as all the `@abstractmethod ` decorators were removed

AlexandraVolokhova:
Thank you for the great work! I've added a bunch of suggestions. It seems to me that it would be worth to add a config file for a case when one wants to create a uniform or a random policy, but I don't have a strong opinion about it.

gflownet/policy/base.py
```diff
@@ -21,82 +20,18 @@ def __init__(self, config, env, device, float_precision, base=None):
         self.base = base
 
         self.parse_config(config)
+        self.instantiate()
 
     def parse_config(self, config):
```
AlexandraVolokhova, line 25: I would suggest to get rid of the method `parse_config` and the `config` argument itself in the `__init__` method and instead make all the arguments explicit (as it is done for the environments)

engmubarak48:
Hi @AlexandraVolokhova just got time, let me address your comments... 

Regarding:
>  It seems to me that it would be worth to add a config file for a case when one wants to create a uniform or a random policy, but I don't have a strong opinion about it.

I think I agree with you, maybe it would have been nice to have a separate config for fixed and uniform policies (I am hoping you are referring to the two functions inside the base policy class). To be honest I don't even still know why we need them.. We could also keep them as it is with the default configuration.

gflownet/policy/base.py
engmubarak48: Why would you want this to be removed out of curiosity? are you referring the whole `def parse_config` function to be removed from the base? in the above comment you mention `super().parse_config should be called to parse ...` if we remove then i don't get which parse_config we will be initializing. Have a look at my next commit for minor update tho.

gflownet/policy/base.py
engmubarak48: This sounds good, will resolve in next commit

gflownet/policy/cnn.py
engmubarak48: will be resolved in next commit, thanks

gflownet/policy/cnn.py
engmubarak48: Yep we could do that.. tho I don't understand why we would need self.type and self.checkpoint from the base?
if one choose CNN policy then type is CNN ?

gflownet/policy/base.py
engmubarak48: pushed update on this :)

alexhernandezgarcia:
Force-pushed after rebasing onto main

PR #328, opened 2024-06-25
[README] Fine-tuning of various sections

alexhernandezgarcia:
I have made a few updates:
- Changed the order of the sections, I wanted to have the installation and quickstart sections earlier since they are what some people might be looking for. I have moved down the Scrabble tutorial.
- Fine-tuned the first part of the Scrabble tutorial. 
- I have updated the section about How to train a GFlowNet model.
- A few other small changes and fixes here and there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,52 @@ The policy models are neural networks that model the forward and backward transi
 
 The GFlowNet Agent is the central component that ties all others together. It orchestrates the interaction between the environment, policies, and proxy, as well as other auxiliary components such as the Evaluator and the Logger. The GFlowNet can construct training batches by sampling trajectories, optimise the policy models via gradient descent, compute evaluation metrics, log data to [Weights & Biases](https://wandb.ai/), etc. The agent can be configured to optimise any of the following loss functions implemented in the library: flow matching (FM), trajectory balance (TB), and detailed balance (TB) and forward-looking (FL). 
 
-#### Exploring the Scrabble Environment
+## Installation
+
+**Quickstart: If you simply want to install everything, run `setup_all.sh`.**
+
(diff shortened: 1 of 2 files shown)


engmubarak48 (commented):
Thanks @alexhernandezgarcia, this looks good with me, happy to merge unless you are  planning to make other changes?

README.md
```diff
@@ -42,81 +42,165 @@ The policy models are neural networks that model the forward and backward transi
...
+All other configurable options are handled similarly. For example, we recommend creating a user configuration file in `./config/user/myusername.yaml` specifying the directory for the log files in `logdir.root`. Then, it can be included in the command with `user=myusername` or `user=$USER` if the name of the YAML file matches our system username.
 
-You can observe the initial state of the environment. For Scrabble environment, this would be an empty board or sequence:
+As another example, you may also want to configure the functionality of the Logger, the class which helps manage logging to [Weights & Biases](https://wandb.ai/) during the training and evaluation of the model. Logging to WandB is disabled by default. In order to enable it, make sure to set up your WandB API key and set the configuration variable `logger.do.online` to `True` in your experiment config file or via the command line:
+
+```bash
+python main.py +experiments=grid/corners logger.do.online=True
+```
+
+Finally, also note that by default, PyTorch will operate on the CPU because we have not observed performance improvements by running on the GPU. You may run on GPU with `device=cuda`.
```
engmubarak48, line 110: You may have to be careful about this statement. I am not sure how performance is related to the CPU or GPU? if you have a big policy model, it will be unavoidable to use GPU :)

alexhernandezgarcia: I think the phrase "we have not observed..." is cautious enough, but  you're right in that we should investigate this in more detail.

PR #332, opened 2024-07-04
[Cleanup] Remove old, unused methods in gflownet.py

alexhernandezgarcia:
Fairly simple: there were a couple of old methods hanging in `gflownet.py` that are not used any longer.

Update: deleting also a couple of methods in `logger.py`.


alexhernandezgarcia (commented):
I will merge, since it's been pending for a while and it should be harmless.

PR #336, opened 2024-07-12
Prevent unnecessary recalculation of proxy values

alexhernandezgarcia:
#### Context

While the Batch kept track of the availability of rewards and log-rewards, it did not store the proxy values, which are not needed for the computation of the loss. However, the proxy values were (re-)computed for logging purposes and for the Buffer. This was a waste of computation, especially if the proxy is expensive.

#### Main changes

- The call of `proxy.rewards()` now allows returning the proxy values, alongside the rewards, if `return_proxy` is True.
- After computing the rewards, the Batch now also stores the proxy values.
- The GFlowNet agent now retrieves the proxy values from the Batch, without recomputing them.
- The GFlowNet agent computes the (natural) rewards from the log-rewards and vice-versa.

#### Other changes

- The buffer does not store the proxy values, since they are unused.
- Some aspects of the Batch have been extended and refactored for better functionality and clarity.
- The logger now logs stats about the log-rewards too.
- Extended tests.
- Extended docstring.

### DIFF
--- a/gflownet/gflownet.py
+++ b/gflownet/gflownet.py
@@ -1168,35 +1168,34 @@ class GFlowNetAgent:
                     all_losses.append([i.item() for i in losses])
             # Buffer
             t0_buffer = time.time()
-            # TODO: the current implementation recomputes the proxy values of the
-            # terminating states in order to store the proxy values in the Buffer.
-            # Depending on the computational cost of the proxy, this may be very
-            # inneficient. For example, proxy.rewards() could return the proxy values,
-            # which could be stored in the Batch.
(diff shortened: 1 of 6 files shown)


ftherrien:
Thanks so much that was quick! I will "review" it later today

ftherrien (approved):
Seems great to me. My understanding is that you mainly added a method to access `self.proxy_values`, made use of an existing method to get the rewards and added methods to check if the reward is log. Doc-strings are clear, internal variables renaming makes sense and tests seem relevant.

PR #337, opened 2024-08-02
Extend common tests and env base

carriepl:
Those are changes that have already been submitted in a PR for a specific project and have already been reviewed in that context. This is a PR to make those changes public.

### DIFF
--- a/gflownet/envs/base.py
+++ b/gflownet/envs/base.py
@@ -2,6 +2,7 @@
 Base class of GFlowNet environments
 """
 
+import numbers
 import uuid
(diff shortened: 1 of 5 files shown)


carriepl:
That's interesting. One test is failing but it is entirely unrelated to the changes in this PR. 

`FAILED tests/gflownet/envs/test_spacegroup.py::test__states_are_compatible_with_pymatgen - AssertionError: assert 'P2_12_12_1' == 'P2_12_121'`

This might mean that the pymatgen library has changed in a way that breaks the code

PR #339, opened 2024-08-15
Fix CI crashing with test_spacegroup

AlexandraVolokhova:
I've updated the names of the spacegroups in `gflownet/envs/crystals/space_groups.yaml` such that they correspond to the new conventions in pymatgen=2024.8.9.  All tests pass, but I'm not 100% sure this change doesn't affect some other things in the crystal-gfn. @alexhernandezgarcia or @carriepl if you have a moment to double check that it's all good, that would be very nice 🙏

### DIFF
--- a/gflownet/envs/crystals/space_groups.yaml
+++ b/gflownet/envs/crystals/space_groups.yaml
@@ -182,12 +182,12 @@
   crystal_class: rhombic-disphenoidal
   crystal_lattice_system_idx: 3
   crystal_system: orthorhombic
-  full_symbol: P2_12_121
+  full_symbol: P2_12_12_1
   lattice_system: orthorhombic
   point_group: '222'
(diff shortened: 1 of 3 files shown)


carriepl (approved):
Looks good to me although I am not aware of everything that could break as a result of these changes. AFAIK it's fine but I don't know for a fact.

Also, interestingly, another test is failing here related to the batch class. To me, it seems unrelated to the current PR so it shouldn't stop the merge. But we should look at what test it is, what the problem is (based on the face that we haven't seen this before, it might be cause by a rare stochastic bug), and try to fix in another PR.

AlexandraVolokhova:
yes, seems like this crashing is very stochastic, I've tried to fix it with stronger atol, it was still failing, then I removed it and it passed the tests. Meanwhile locally everything worked fine all the time.

PR #348, opened 2025-02-21
Package installation: dependencies and instructions

alexhernandezgarcia:
This PR re-tackles several aspects of the installation of the gflownet package.

First, I have updated the `pyproject.toml` file to match the current guidelines of Poetry. And I have taken the chance to also update some of the dependencies and re-organise them.

Then, I have also largely rewritten the instructions to install the necessary dependencies to use the gflownet package. I believe the installation is now simpler, more transparent and more intuitive. For example, I have got rid of the `prereq_*.sh` files and tried to include all of that as part of either `pyproject.toml` or the steps, such as the installation of PyTorch.

I have also written a bash script `install.sh` that carries out the installation in one go, with the possibility of selecting some options via arguments. All of it is documented in the instructions of the README.

I have performed several tests to make sure things run and are installed as expected but it is quite possible that I am missing some things. We will see as we go... As an important note, I have not tested at all the installation of the `molecules` extras. Left for later, since this may require an installation with Anaconda anyway, quite sadly.

Other stuff:
- I have included some `try: ... except: ...` clauses in some tests that skip the tests when the necessary dependencies are not installed.
- I have removed a few unused imports in some files (but many files still have unused imports)
- I have updated the database of space groups to a later pymatgen version, since some names have changed. 
- I have updated the installation protocol of the CLI


tests/gflownet/envs/test_composition.py
```diff
@@ -3,6 +3,16 @@
...
+# Skip the entire module if pyxtal is not installed / cannot be imported
+try:
+    import pyxtal
+except:
+    pytest.skip(
+        "Skipping all tests in test_composition.py because pyxtal could not be "
+        "imported.",
+        allow_module_level=True,
+    )
+
```
engmubarak48, line 15: I wonder instead of skipping if we instead ask the user to install the package ? although skipping tests when a module is missing is a good practice to avoid unnecessary failures. 

also in the except how about we catch module error:
`except ModuleNotFoundError:`

also how about if we use something like. 
 ```pytest.importorskip("pyxtal", reason="Skipping all tests because pyxtal is not installed.")```

engmubarak48:
@alexhernandezgarcia overall looks good with me.. I haven't run and tested locally (will do), but the changes look okey with me.. skipping tests could be tricky but could fix potential issues and errors.

Other comments:
- in the future, we may have to avoid version based installation (at least to some extent, everything should work with latest version, tho unrealistic most of the time). 
- Scrabble (add environment) should go to the main documentation. we can have only minor examples in the main's readme.
- not sure the purpose of sbatch folder, this depends on the users and will have their own setup for clusters/slurm.. same for mila folder. 

The last two comments could be separate PRs and unrelated to the installation one.

engmubarak48 (approved):

tests/gflownet/envs/test_composition.py
alexhernandezgarcia: Thanks! I didn't know about `pytest.importorskip`. If it skips the entire module, then we should use that. I will take a look.

alexhernandezgarcia: Changed to `pytest.importorskip` in https://github.com/alexhernandezgarcia/gflownet/pull/348/commits/7b15b5289cf0583541efc2b7a93bfa13002a1221

alexhernandezgarcia:
> @alexhernandezgarcia overall looks good with me.. I haven't run and tested locally (will do), but the changes look okey with me.. skipping tests could be tricky but could fix potential issues and errors.
> 
> Other comments:
> 
>     * in the future, we may have to avoid version based installation (at least to some extent, everything should work with latest version, tho unrealistic most of the time).
> 
>     * Scrabble (add environment) should go to the main documentation. we can have only minor examples in the main's readme.
> 
>     * not sure the purpose of sbatch folder, this depends on the users and will have their own setup for clusters/slurm.. same for mila folder.
> 
> 
> The last two comments could be separate PRs and unrelated to the installation one.

I essentially agree, but I also think that should go on different PRs.

I think it's ready to merge now!

PR #353, opened 2025-03-06
[Crystal] Adapt Dave proxy to update in proxy repo

alexhernandezgarcia:
Previously, the various Dave proxies (for example, to predict the formation energy or the band gap) were anchored to different releases. This has been recently changed in the [crystalproxies repo](https://github.com/sh-divya/crystalproxies) (whose name has been also updated). Now, the installation of the dave package and its use is much simpler. This PR updates the relevant code in the gflownet repo, which essentially simplifies the code in the Dave proxy by removing unnecessary code and checks.

From now on, to use one proxy model or another, one simply passes the path to its checkpoint file (without a release). The dependence on a release is handled by the dependencies specified in `pyproject.toml` as for any other package.

### DIFF
--- a/config/experiments/crystals/starling_fe_restricted_b.yaml
+++ b/config/experiments/crystals/starling_fe_restricted_b.yaml
@@ -96,7 +96,7 @@ policy:
 proxy:
   reward_min: 1e-08
   do_clip_rewards: True
-  release: 0.3.4 # Formation energy release
+  ckpt_path: /network/projects/crystalgfn/ckpts/dave/eform/
   # Boltzmann (exponential), with negative beta because the formation energy is negative and the lower the better
   reward_function: exponential
   # Parameters of the reward function
--- a/gflownet/proxy/crystals/dave.py
+++ b/gflownet/proxy/crystals/dave.py
@@ -2,11 +2,12 @@ from copy import deepcopy
 from importlib.metadata import PackageNotFoundError, version
 
 import torch
+from dave import DavePredictor
 from torchtyping import TensorType
 
 from gflownet.proxy.base import Proxy
 
-REPO_URL = "https://github.com/sh-divya/ActiveLearningMaterials.git"
+REPO_URL = "https://github.com/sh-divya/crystalproxies.git"
 """
 URL to the proxy's code repository. It is used to provide a link to the
 appropriate release link in case of version mismatch between requested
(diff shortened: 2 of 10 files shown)


config/experiments/crystals/starling_fe_restricted_b.yaml
sh-divya, line 1: I assume you don't specify rescale outputs here (and other places), because it's True by default?

gflownet/proxy/crystals/dave.py
```diff
@@ -2,142 +2,65 @@
...
 """
 
 
 class DAVE(Proxy):
     def __init__(
         self,
-        ckpt_path=None,
-        release=None,
+        ckpt_path=str,
         rescale_outputs=True,
```
sh-divya, line 23: I just wanted to double check that you want to call this variable rescale outputs still. The flag that you pass at the moment is if you want to rescale inputs. Technically, the outputs are rescaled within the proxy loader automatically if the proxy was trained with the outputs scaled.

config/experiments/crystals/starling_fe_restricted_b.yaml
alexhernandezgarcia: The actual reason is that it was never specified, but I agree explicit is better.

gflownet/proxy/crystals/dave.py
alexhernandezgarcia: Oh I see, I was not understanding this properly. We should clarify this in the code and name variables accordingly. Can you share the link to the script (in the right branch) that handles this in https://github.com/sh-divya/crystalproxies so that I can get a better understanding of it?

sh-divya: So in [crystalproxies](https://github.com/sh-divya/crystalproxies/blob/change_loading/dave/utils/pretrained.py), line 47 and 48, the scales are loaded from the model checkpoint.
```python
self.xscale = model_hp["scales"]["x"]
self.yscale = model_hp["scales"]["y"]
``` 
When the model is called, in line 53 and 54, you can specify whether you want to scale the inputs, depending on what you have in ur inputs.
```python
def __call__(self, x, scale_input=False):
    if scale_input:
        comp, sg, lat = x
        lat = (lat - self.xscale["mean"]) / self.xscale["std"]
        x = (comp, sg, lat)
``` 
Finally, there is the outputs from the proxy prediction are scaled, in line 58. This only depends on whether the model was trained with outputs scaled, and depends on whether it was present in the checkpoint
```python
if self.yscale:
    return (self.proxy(x) * self.yscale["std"]) + self.yscale["mean"]
else:
    return self.proxy(x)
```

alexhernandezgarcia: Awesome, thanks! Then, should we pass scale_input to the predictor from the GFN proxy? The lattice parameters passed from the GFN proxy will be in the ranges indicated in the config, that is (by default):
- Lengths: 1-350
- Angles: 50-150

sh-divya: Yes. My only comment was that rescale_ouputs (the old variable) is not suitable in the context of the new proxy loader.

alexhernandezgarcia: Yes, I got that. 

Now, what should we do with the inputs? Pass `scale_input=True` or `scale_input=False`?

sh-divya: Should be true

alexhernandezgarcia: Updated in https://github.com/alexhernandezgarcia/gflownet/pull/353/commits/c761e608e0561afe67660a2b766eaea8a7aec8ae

With this, I will merge. Thanks for the review!

PR #357, opened 2025-03-18
[GENERAL] Checkpoints, Resuming, Buffer, Logging and more

alexhernandezgarcia:
Apologies for such a big and non-targeted PR, but it can be explained by the fact that many of these changes come from work done privately for other projects, and it would be really hard to isolate the various aspects that are updated. On the flip side, here come a bunch of updates that can be quite useful, such as the possibility to easily resume training, better logging, a fixed replay buffer, etc.

## Main scripts

Previously, to train a GFlowNet we would run `main.py` and to evaluate it post-training, we would run `scripts/eval_gflownet.py`. These two scripts were not well coordinated in the sense that both repeated code to initialise a GFlowNet agent and all it needs, but we were not maintaining both properly. I have introduced a few improvements:

- `main.py` has been renamed `train.py`
- `scripts/eval_gflownet.py` is now `eval.py`
- And there is also a new `resume.py` (see below)
- All three use the same common method `gflownet_from_config(config)`
- All run modes (train, eval, resume) can now be configured with Hydra (see config files)

I have added some of this to  the README.

## Resuming runs

Now it is possible to resume a previous run with `resume.py`:
```
python resume.py rundir=path/to/a/run/directory/
```

The `path/to/a/run/directory/` is the path printed at the beginning of training, for example:

```
Working directory of this run: /home/username/git/gflownet                                                                                                                                     
Logging directory of this run: path/to/a/run/directory/
```

To facilitate resuming runs, I have re-implemented the saving and loading of checkpoints. Now only one checkpoint file is stored each time and it contains a dictionary of the stuff that is necessary: forward policy, backward policy, state flow, logZ, optimizer state, iteration number, wandb id and the buffer (with paths to the data sets).

#### Train a grid with buffers for 657 iterations only. Once trained, change the number of training steps to 3000 and resume.

1. Run the Grid corners with buffers experiment for 657 iterations.
```
python train.py user=$USER +experiments=grid/corners_buffers logger.do.online=True gflownet.optimizer.n_train_steps=657 evaluator.checkpoints_period=300
```

2. Increase the number of train steps `gflownet.optimizer.n_train_steps` to 3000 in the config file within the `.hydra/` directory of the run.

3. Run the `resume.py` script:
```
python resume.py user=$USER rundir=/network/scratch/h/hernanga/logs/gflownet/grid/corners/2025-03-17_22-24-47_071765/
```

Things go smoothly and [the wandb training curves](https://wandb.ai/alexhg/grid/runs/mvpop3xo) continue from the previous run.

#### Train a grid for 3000 iterations but stop it mid-way, then resume

1. Run the Grid corners with buffers experiment for 657 iterations.
```
python train.py user=$USER +experiments=grid/corners logger.do.online=True
```

2. Stop the run (in my case at around 1100 iterations)

3. Run the `resume.py` script:
```
python resume.py user=$USER rundir=/network/scratch/h/hernanga/logs/gflownet/grid/corners/2025-03-17_22-43-08_975398
```

WandB run: https://wandb.ai/alexhg/grid/runs/8tp3fmh3

## Buffer

The Buffer has been largely rewritten There were some bugs and poorly implemented things that have been improved (kudos to @AlexandraVolokhova)

An important point is that the Buffer has been turned into its own module and detached from the environment. So now the buffer config is not specified in the environment config but in the buffer.  

Note that there are some new unit tests for the Buffer (thanks @AlexandraVolokhova) but I have disabled them so far until I update them in a separate PR

## Logging

Metrics logging has been improved, with some refactoring in `gflownet`.py` and `logger.py` which will allow for easier extensions in the (near) future. Also, the metrics names have been clarified, for example to indicate that the mean rewards are from the training batch, not GFN on-policy samples (that will be done in a separate PR). Also the rewards statistics from the replay buffer are logged.

Note that I have created the method `log_train_iteration()` in `gflownet.py` to modularise the logging operations and make `train()` simpler (shorter). There are other changes under the same philosophy.

## More stuff

- Extended docstrings
- Cleaned-up configs


config/buffer/base.yaml
```diff
@@ -0,0 +1,33 @@
...
+  path: null
+  n: null
+  seed: null
+test:
+  type: null
+  path: null
+  n: null
+  seed: null
+
+# Whether to add all training samples to a buffer
```
AlexandraVolokhova, line 29:
```suggestion
# Whether to maintain a separate "main" buffer for storing all terminal states visited during the training 
```

AlexandraVolokhova: I think it's important to clarify here that main buffer is different from the replay buffer

resume.py
```diff
@@ -0,0 +1,100 @@
...
+    if len(gflownet.buffer.replay) > 0:
+        print("\nReplay buffer:")
+        print(gflownet.buffer.replay)
+
+    # Close logger
+    # TODO: make it gflownet.end() - perhaps there are other things to end
+    gflownet.logger.end()
+
+
+def set_seeds(seed):
```
AlexandraVolokhova, line 87: Minor, but I think this function should be in gflownet.utils.common so all other scrips (like train.py, eval.py) set the seeds in the same way

AlexandraVolokhova (approved):
Looks good to me, thank you for this great effort! 💪

PR #358, opened 2025-03-18
[Crystal] One-hot encoding of space group policy inputs

alexhernandezgarcia:
This PR implements the possibility of using a onehot encoding of the space group environment's states for the policy inputs and sets it as the default option. In particular, the input to the policy model is the concatenation of the one-hot encodings of the crystal-lattice system, point symmetry and space group number.  

The previous version, which simply uses a 3-dimensional vector with the indices (as floats) of each property, remains available by setting the argument `policy_fmt="indices"`.

I have implemented basic unit tests of the policy encoding too.


gflownet/envs/crystals/spacegroup.py
```diff
@@ -265,6 +283,95 @@ def states2proxy(
...
+            )
+
+    def states2policy_onehot(
+        self, states: List[List]
+    ) -> TensorType["batch", "policy_input_dim"]:
+        """
+        Prepares a batch of states in "environment format" for the policy model: states
+        are one-hot encoded.
+
+        In particular, the policy input for a state is a vector containing the
```
lpodina, line 319: I would specify explicitly how you are handling when one of these is not set

gflownet/envs/crystals/spacegroup.py
```diff
@@ -265,6 +283,95 @@ def states2proxy(
...
+        states : list
+            A batch of states in environment format, that is a list of lists.
+
+        Returns
+        -------
+        A tensor containing the policy representation of all the states in the batch.
+        """
+        states = tlong(states, device=self.device)
+        cls_onehot = F.one_hot(
+            torch.searchsorted(self.cls_valid, states[:, 0]),
```
lpodina, line 358: `torch.searchsorted` seems a bit obscure, consider adding a comment clarifying what this is doing

gflownet/envs/crystals/spacegroup.py
```diff
@@ -94,26 +95,43 @@ def __init__(
...
             A list with the number of atoms per element, used to compute constraints on
             the space group. 0's are removed from the list. If None, composition/space
             group constraints are ignored.
-        """
+                policy_fmt : str
+            Specifies the policy encoding. Options:
+                - onehot: One-hot encoding of each property (crystal-lattice system,
+                  point symmetry, space group), all concatenated to make the overall
+                  input.
+                - indices: A three-dimensional vector with the indices of each property
```
lpodina, line 117: I would add an example here (or elsewhere) for `indices` format because there is no function for it like there is for `onehot` and it might be opaque what the super call actually does to someone who is not familiar with the code

gflownet/envs/crystals/spacegroup.py
```diff
@@ -265,6 +283,95 @@ def states2proxy(
         states = tlong(states, device=self.device)
         return torch.unsqueeze(states[:, self.sg_idx], dim=1)
 
+    def states2policy(
+        self, states: List[List]
+    ) -> TensorType["batch", "policy_input_dim"]:
+        """
+        Prepares a batch of states in "environment format" for the policy model, by
```
lpodina, line 290: This isn't quite related to the PR but I thought I'd mention it anyway- I found the description on line 290 (and other lines like it) to be a bit confusing, when I was unfamiliar with the code and reading the docs to understand it. A clearer sentence might be "Converts a batch of GFlowNet states from "environment format" to "policy format", to be accepted by the policy model."

gflownet/envs/crystals/spacegroup.py
alexhernandezgarcia: Done in https://github.com/alexhernandezgarcia/gflownet/pull/358/commits/277d4047904a4e117db667a8e45805d56b526627

gflownet/envs/crystals/spacegroup.py
alexhernandezgarcia: Done in https://github.com/alexhernandezgarcia/gflownet/pull/358/commits/cc2883dfa40895c6bc31db4019eda3ed7693e004

PR #366, opened 2025-03-26
[Crystal] SGCCG Lattice Parameters

alexhernandezgarcia:
- Implements the LatticeParametersSGCCG environment (implemented before and already reviewed privately)
    - I have included a small fix in `is_valid()`
    - I have added some code to avoid throwing a RuntimeWarning  
- Incorporates LatticeParametersSGCCG into the Crystal environment, while keeping the possibility to train with the previous (_natural_) LatticeParameters environment.


alexhernandezgarcia:
Will bypass merging rule because the core of this PR has been reviewed elsewhere.

PR #369, opened 2025-03-31
[Buffer] Speed up diversity check in replay buffer

alexhernandezgarcia:
The replay buffer behaves significantly different - with a less positive or even with a negative effect - if the diversity of the samples in the buffer is not taken into account when inserting new samples.

Currently, because the Buffer is not yet implemented as a more efficient data structure, checking the similarity of a new sample with the samples already present in the replay buffer requires many pairwise comparisons. This incurs a significant computational overhead as the the capacity of the replay buffer gets larger and the states more complex, as is the case of the Crystal-GFN. In particular, I have observed that **checking the diversity of the replay buffer can account for more than half of the training time**.

This PR reduces the computational time of the diversity check in the replay buffer by comparing new samples only with those samples in the replay buffer that have similar reward. 

I have first introduced a new parameter for the Buffer, `diversity_check_reward_similarity`, which I will refer here as δ. Then, for every new sample that has larger reward than the minimum reward in the replay buffer, we compute the difference d between the maximum Rmax and minimum reward Rmin in the replay buffer, d = Rmax - Rmin. Then, we perform a similarity check, that is `self.env.isclose(sample, rsample)` where `sample` is the new sample and `rsample` is a sample from the replay buffer`, if the absolute difference between the rewards of `sample` and `rsample` is smaller than d * δ.

Using a δ of 0.01 in the Crystal-GFN, I observe that the run time is equivalent to not performing any similarity check while I see no visible impact in the performance and specifically on the training statistics of the replay buffer.  

Other improvements that can be introduced:
- Disable attempting to insert samples in the replay buffer that were sampled from the replay buffer itself.

### DIFF
--- a/config/buffer/base.yaml
+++ b/config/buffer/base.yaml
@@ -32,5 +32,14 @@ use_main_buffer: False
 # Whether new samples are compared to samples in the buffer before adding
 check_diversity: False
 
+# The accepted level of similarity of rewards to include samples from the
+# replay buffer in the diversity check. Assuming check_diversity is True, given
+# a sample x with reward R(x), the diversity check will only be performed
+# against those samples in the replay buffer whose reward difference with
+# respect to R(x) is smaller than diversity_check_reward_similarity times the
+# difference between the maximum reward and the minimum reward in the replay
+# buffer.
+diversity_check_reward_similarity: 0.1
+
 # Whether to show a progress bar while processing the data sets. False by default.
 progress_process_dataset: False
--- a/config/experiments/crystals/starling_bg.yaml
+++ b/config/experiments/crystals/starling_bg.yaml
@@ -64,6 +64,8 @@ env:
 # Buffer
 buffer:
   replay_capacity: 1000
+  check_diversity: True
+  diversity_check_reward_similarity: 0.01
   train:
     type: csv
     path: /network/projects/crystalgfn/data/bandgap/train.csv
--- a/config/experiments/crystals/starling_bg_no_constraints.yaml
+++ b/config/experiments/crystals/starling_bg_no_constraints.yaml
@@ -64,6 +64,8 @@ env:
 # Buffer
 buffer:
   replay_capacity: 1000
+  check_diversity: True
+  diversity_check_reward_similarity: 0.01
   train:
     type: csv
     path: /network/projects/crystalgfn/data/bandgap/train.csv
--- a/config/experiments/crystals/starling_density.yaml
+++ b/config/experiments/crystals/starling_density.yaml
@@ -64,6 +64,8 @@ env:
 # Buffer
 buffer:
   replay_capacity: 1000
+  check_diversity: True
+  diversity_check_reward_similarity: 0.01
   train:
     type: csv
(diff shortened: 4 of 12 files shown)


AlexandraVolokhova (approved):
Looks good to me! A couple of suggestions / suggestions: 
1. I think it would be nice to still have a possibility to do diversity check over the full replay buffer w/o the rewards restriction. Maybe it would be worth having a flag to enable/disable this restriction?
2. I don't fully understand why d is set to be dependent on R_max - R_min, i.e. it changes during the training, rather than being a fixed value, could you elaborate about this?

alexhernandezgarcia:
> Looks good to me! A couple of suggestions / questions:
> 
>     1. I think it would be nice to still have a possibility to do diversity check over the full replay buffer w/o the rewards restriction. Maybe it would be worth having a flag to enable/disable this restriction?
> 
>     2. I don't fully understand why d is set to be dependent on R_max - R_min, i.e. it changes during the training, rather than being a fixed value, could you elaborate about this?

Thanks for the review!

1. This is still possible by setting `diversity_check_reward_similarity` equal to a large number. Should I enable full check with a value of say -1 too?

2. Well, this is one way I found of many possible ways. I preferred to set the threshold relative to the range of rewards in the replay buffer than an absolute threshold, given that the value of rewards can change very wildly from problem to problem. The way I implemented makes intuitive sense to me: A value of δ = 0.1 means that the diversity check will be done if the difference of rewards is within +/- 10 % of the range of rewards in the replay buffer. Does this make sense? Do you prefer a different way?

AlexandraVolokhova:
1. Yes, but it is not very intuitive that one should set `diversity_check_reward_similarity` to a large number to disable the constraint. I think disabling it with -1 value is a good idea
2. That makes sense, thank you for explaining. I believe it should be fine, but my small worry that the absolute value of this constraint will vary a lot depending on the range of values in the replay buffer and I don't have a good intuition to see if there're any issues this may case (maybe none, but it's hard to tell). Maybe using d = delta * R_max or d = delta * R_min would give a bit less volatile criteria which is still problem-dependent, but I don't mind the current implementation either.

alexhernandezgarcia:
> 1. Yes, but it is not very intuitive that one should set `diversity_check_reward_similarity` to a large number to disable the constraint. I think disabling it with -1 value is a good idea
> 
>     2. That makes sense, thank you for explaining. I believe it should be fine, but my small worry that the absolute value of this constraint will vary a lot depending on the range of values in the replay buffer and I don't have a good intuition to see if there're any issues this may case (maybe none, but it's hard to tell). Maybe using d = delta * R_max or d = delta * R_min would give a bit less volatile criteria which is still problem-dependent, but I don't mind the current implementation either.

Point 1 is done in https://github.com/alexhernandezgarcia/gflownet/pull/369/commits/d4c7f0cad45382203a155baed92560be60218dff

For point 2, I have thought of some extreme cases and it still makes sense to me. Of course, the parameter δ might have to be adjusted if the results are not as expected regarding the buffer, but I cannot think of a better, general way, right now. 

I think it is reasonable to merge the current version.

PR #371, opened 2025-04-08
Add Contributors section to documentation

alexhernandezgarcia:
Because the directory `./docs/contributors/` was outside `./docs/source/` it was not showing up in https://gflownet.readthedocs.io

I have moved the `contributors/` inside `source/` and updated the index so that the contributors guides are online.


alexhernandezgarcia:
Will bypass merge rules because changes affect only the docs.

PR #380, opened 2025-06-25
Gflownet test for grid

AlexandraVolokhova:


alexhernandezgarcia (approved):
I think the two tests could be merged into one single tests with two parameters, one for each environment, but I don't mind cleaning this up later.

PR #383, opened 2025-08-18
Loss function as an object

alexhernandezgarcia:
This PR refactors the code around the GFlowNet loss functions to make the loss function a class and each loss function a separate class.

This change is something we had wanted to do for a while, but it was never a high priority since the impact is rather small. However, since we may implement a couple of new loss functions soon, I thought it would be worth making the change now. 

This is a more modular implementation of the loss function, which makes the code in the GFlowNet agent a bit cleaner and the implementation of new loss functions hopefully clearer.

### To be done in this PR
- [x] Adapt experiment config files
- [x] Remove optimizer.loss from all gflownet configs, e.g. here: https://github.com/alexhernandezgarcia/gflownet/blob/a55ffda346d6c9804b3cf31159eea2c5e8dfac35/config/gflownet/detailedbalance.yaml#L6
- [x] Clean up part about moving average of the loss
- [x] Sanity checks
- [x] Make current tests pass

### To be done in future PRs
- [ ] Make the optimiser modular too
- [ ] Improve the policies part of the code to remove the dependence on the loss function (see https://github.com/alexhernandezgarcia/gflownet/pull/327)
- [ ] Instructions to create new loss functions
- [ ] Unit tests for the losses

### DIFF
--- a/gflownet/losses/base.py
+++ b/gflownet/losses/base.py
@@ -21,11 +21,9 @@ from gflownet.utils.common import set_device, set_float_precision
 
 
 class BaseLoss(metaclass=ABCMeta):
-    # TODO: study whether the dependence on the environment can be removed.
     # TODO: improve dependence on policies (needs re-implementation of policies)
     def __init__(
         self,
-        env_maker: partial,
         forward_policy: Policy,
         backward_policy: Policy = None,
         state_flow: Policy = None,
@@ -40,8 +38,6 @@ class BaseLoss(metaclass=ABCMeta):
 
         Parameters
         ----------
-        env_maker : partial
-            The environment maker of the environments used to train the GFlowNet.
         forward_policy : :py:class:`gflownet.policy.base.Policy`
             The forward policy to be used for training. Parameterized from
             `gflownet.yaml:forward_policy` and parsed with
@@ -72,8 +68,6 @@ class BaseLoss(metaclass=ABCMeta):
             Coefficient for the exponential moving average (EMA) of the loss.
         loss_ema : float
             The exponential moving average of the loss.
-        env : Environment
-            The environment used to train the GFlowNet.
         forward_policy : gflownet.policy.base.Policy
             The forward policy to be used for training. Parameterized from
             `gflownet.yaml:forward_policy` and parsed with
@@ -105,8 +99,6 @@ class BaseLoss(metaclass=ABCMeta):
         self.early_stopping_th = early_stopping_th
         self.ema_alpha = ema_alpha
         self.loss_ema = None
-        # Environment
-        self.env = env_maker()
         # Policy models and parameters
         self.forward_policy = forward_policy
         self.backward_policy = backward_policy
--- a/gflownet/losses/trajectorybalance.py
+++ b/gflownet/losses/trajectorybalance.py
@@ -99,10 +99,10 @@ class TrajectoryBalance(BaseLoss):
         """
         # Get logprobs of forward and backward transitions
         logprobs_f = compute_logprobs_trajectories(
-            batch, self.env, forward_policy=self.forward_policy, backward=False
+            batch, forward_policy=self.forward_policy, backward=False
         )
         logprobs_b = compute_logprobs_trajectories(
-            batch, self.env, backward_policy=self.backward_policy, backward=True
+            batch, backward_policy=self.backward_policy, backward=True
         )
         # Get rewards from batch
         logrewards = batch.get_terminating_rewards(log=True, sort_by="trajectory")
--- a/gflownet/utils/batch.py
+++ b/gflownet/utils/batch.py
@@ -1633,7 +1633,7 @@ class Batch:
 
 def compute_logprobs_trajectories(
     batch: Batch,
-    env: GFlowNetEnv,
+    env: GFlowNetEnv = None,
     forward_policy: Policy = None,
     backward_policy: Policy = None,
     backward: bool = False,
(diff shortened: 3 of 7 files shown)


gflownet/gflownet.py
```diff
@@ -130,34 +135,20 @@ def __init__(
...
-            )
         # Loss
-        if optimizer.loss in ["flowmatch", "flowmatching"]:
-            self.loss = "flowmatch"
-            self.logZ = None
-        elif optimizer.loss in ["trajectorybalance", "tb"]:
-            self.loss = "trajectorybalance"
+        self.loss = loss
+        if self.loss.requires_log_z:
             self.logZ = nn.Parameter(torch.ones(optimizer.z_dim) * 150.0 / 64)
```
AlexandraVolokhova, line 141: I've never noticed this before, but what is there a reason for this multiplication `* 150.0 / 64` ?

gflownet/gflownet.py
```diff
@@ -1372,8 +1059,9 @@ def log_train_iteration(self, pbar: tqdm, losses: List, batch: Batch, times: dic
             )
 
             # Log losses
+            losses["Loss"] = losses["all"]
```
AlexandraVolokhova, line 1062: I think you need to delete the key "all" in the dictionary `losses` as well, otherwise, you're logging the same metric twice under the names "Loss" and "all"

gflownet/losses/__init__.py
AlexandraVolokhova, line 1: I'm confused with this file, could you elaborate why it is needed? We haven't created __init__.py for other subfolders like env and policy, why do we need it for losses? The docstring inside is not very clear to me

gflownet/utils/batch.py
```diff
@@ -1628,3 +1629,77 @@ def get_item(
                 "item must be one of: state, parent, action, done, mask_f[orward] or "
                 "mask_b[ackward]"
             )
+
+
+def compute_logprobs_trajectories(
+    batch: Batch,
+    env: GFlowNetEnv,
```
AlexandraVolokhova, line 1636: I'm thinking that maybe passing `env` is not necessary here because we can always access` batch.readonly_env` if we need some methods of the env. And if we remove dependency on the env hre, we can also remove it in the Loss class (if I understand correctly, `self.env` is used there only to call `compute_logprobs_trajectories` function)

gflownet/losses/__init__.py
alexhernandezgarcia: I'm confused because the `__init__.py` file also is there for the envs and policy (and other modules) because it is (was?) required. I have actually just learnt that it may not be required any longer in recent versions of Python.

https://stackoverflow.com/questions/448271/what-is-init-py-for

gflownet/gflownet.py
AlexandraVolokhova: I'd put a TODO item to make this constant a parameter of the training

gflownet/utils/batch.py
AlexandraVolokhova: But we can also use `batch.readonly_env` inside the detailed balance loss call, can't we?

gflownet/losses/base.py
```diff
@@ -0,0 +1,277 @@
...
+from gflownet.utils.common import set_device, set_float_precision
+
+
+class BaseLoss(metaclass=ABCMeta):
+    # TODO: study whether the dependence on the environment can be removed.
+    # TODO: improve dependence on policies (needs re-implementation of policies)
+    def __init__(
+        self,
+        env_maker: partial,
+        forward_policy: Policy,
```
LenaNehale, line 29: I am still getting familiar with the codebase so correct me if I am wrong! 
I was confused to see forward_policy used in the flow-matching loss, because there we are learning flows (different from forward policy, which implies that the distribution is normalized to sum to 1), what do you think of adding another parameter "state_action_flow: Policy" as using this for flow matching? It's just for naming purposes, otherwise the code for flow-matching makes sense

alexhernandezgarcia: I see your point but I think it would be more confusing if we end up having different names for the same thing. The meaning of "policy" in the codebase is "a model used to learn a sampling policy". It is valid for the flow matching because these policy models do not learn a (normalised) probability distribution, so their outputs can be interpreted as flows. And we can use the same models and notion for other losses, such as TB, because we can the outputs of these policy models and normalise them to obtain a probability.

Does that make sense?

LenaNehale: Oh ok, it makes sense, thank you!

gflownet/utils/batch.py
AlexandraVolokhova: Nice, looks good to me!

gflownet/losses/trajectorybalance.py
```diff
@@ -0,0 +1,138 @@
...
+            batch, forward_policy=self.forward_policy, backward=False
+        )
+        logprobs_b = compute_logprobs_trajectories(
+            batch, backward_policy=self.backward_policy, backward=True
+        )
+        # Get rewards from batch
+        logrewards = batch.get_terminating_rewards(log=True, sort_by="trajectory")
+
+        # Trajectory balance loss
+        return (self.logZ.sum() + logprobs_f - logprobs_b - logrewards).pow(2)
```
LenaNehale, line 111: I just saw this, I think it should be logZ.mean() , not logZ.sum()  , otherwise the final logZ changes with the batch size. If the batch size is fixed during training it's fine, but I would change it to avoid problems in case the bs changes (e.g. in the case of training with a replay buffer, at the beginning we still don't have enough samples in the replay buffer so the actual bs is smaller)

PR #386, opened 2025-09-07
Make logprobs-efficiency work for ccube

AlexandraVolokhova:
Made it work for ccube, comparison runs: https://api.wandb.ai/links/alexandra-volokhova/fa10iwsy

### DIFF
--- a/gflownet/envs/base.py
+++ b/gflownet/envs/base.py
@@ -612,7 +612,7 @@ class GFlowNetEnv:
 
             ipdb.set_trace()
         device = policy_outputs.device
-        n_states = len(states_from)
+        n_states = policy_outputs.shape[0]
         ns_range = torch.arange(n_states, device=device)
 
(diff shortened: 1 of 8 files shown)


gflownet/envs/base.py
```diff
@@ -512,9 +513,9 @@ def randomize_and_temper_sampling_distribution(
         policy_outputs : tensor
             The modified policy outputs.
         """
-        if temperature != 1.0:
+        if not math.isclose(temperature, 1.0, abs_tol=1e-08):
```
alexhernandezgarcia, line 516: Why do we need this?

AlexandraVolokhova: To make sure that we don't get errors by comparing floats with `!=`

alexhernandezgarcia (approved):
Cooool, let's merge this and keep working on the remaining things.

PR #387, opened 2025-09-07
Set float for mlp policy

AlexandraVolokhova:
Enable setting float precision for the mlp policy


alexhernandezgarcia (approved):
Looks good, thanks!

PR #389, opened 2025-09-07
Logprobs efficiency separate

AlexandraVolokhova:
Separate env.sample_actions_batch and env.get_logprobs


AlexandraVolokhova:
Ready to review. One test is failing in the checks. but locally it passes

PR #390, opened 2025-09-09
[Logprobs efficiency] Improve handling of actions in `get_logprobs()`

alexhernandezgarcia:
- Improve the handling of actions in `get_logprobs()`:
  - By default, the format of the actions remains a list of tuples. This results in easier-to-read code and it is even more efficient.
  - But in some environments (meta-environments such as Set and Stack) it is practical to convert the actions into a tensor for easier indexing.
  - Therefore, `get_logprobs()` may receive actions as a list or as a tensor
- The conversion from a tensor of actions into actions as tuples is also improved.

### DIFF
--- a/gflownet/envs/base.py
+++ b/gflownet/envs/base.py
@@ -19,7 +19,14 @@ import torch
 from torch.distributions import Bernoulli, Categorical
 from torchtyping import TensorType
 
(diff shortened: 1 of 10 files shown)


AlexandraVolokhova (approved):
Looks good to me!

PR #396, opened 2025-09-19
Fix logprob bug

AlexandraVolokhova:
There is a bug introduced by the logprobs efficiency PR: runs are crushing with gflownet.collect_reversed_logprobs=False and with both forward and backward trajectories present in the training batch. This PR fixes this bug.

### DIFF
--- a/gflownet/utils/batch.py
+++ b/gflownet/utils/batch.py
@@ -13,6 +13,7 @@ from gflownet.utils.common import (
     concat_items,
     copy,
     extend,
+    index_select,
     set_device,
     set_float_precision,
     tbool,
@@ -120,6 +121,7 @@ class Batch:
         self._logrewards_parents_available = False
         self._logrewards_source_available = False
         self._proxy_values_available = False
+        self._logprobs_available = False
 
     def __len__(self):
         return self.size
@@ -323,7 +325,7 @@ class Batch:
             if backward:
                 # Add logpb for current action
                 self.logprobs_backward.append(logp)
-                self.logprobs_backward_valid.append(True)
+                self.logprobs_backward_valid.append(True if logp is not None else False)
                 # Add logpf for previous action
                 if not env.is_source(env.state) or logp_rev is None:
                     # Store a None placeholder for the current logpf which will be
@@ -352,7 +354,7 @@ class Batch:
             else:
                 # Add logpf for current action
                 self.logprobs_forward.append(logp)
-                self.logprobs_forward_valid.append(True)
+                self.logprobs_forward_valid.append(True if logp is not None else False)
                 # Add logpb for previous action
                 if not env.done or logp_rev is None:
                     # Store a None placeholder for the current logpb which will be
@@ -484,6 +486,7 @@ class Batch:
         policy: Optional[bool] = False,
         proxy: Optional[bool] = False,
         force_recompute: Optional[bool] = False,
+        index: Optional[Union[List, Tuple, TensorType, npt.NDArray]] = None,
     ) -> Union[TensorType["n_states", "..."], npt.NDArray[np.float32], List]:
         """
         Returns all the states in the batch.
@@ -506,11 +509,16 @@ class Batch:
             If True, the policy states are recomputed even if they are available.
             Ignored if policy is False.
 
+        index: list, tuple, tensor or np.ndarray
+            1-dimentional sequence of bacth indecies for selecting states, optional.
+            If None (default), all the states will be returned.
+
         Returns
         -------
         self.states or self.states_policy or self.states2proxy(self.states) : list or
         torch.tensor or ndarray
-            The set of all states in the batch.
+            The set of states in the selected format with the selected index. If index is
+            None, all states of the batch are returned.
         """
         if policy is True and proxy is True:
             raise ValueError(
@@ -519,10 +527,10 @@ class Batch:
         if policy is True:
             if self.states_policy is None or force_recompute is True:
                 self.states_policy = self.states2policy()
-            return self.states_policy
+            return index_select(self.states_policy, index)
         if proxy is True:
-            return self.states2proxy()
-        return self.states
+            return index_select(self.states2proxy(), index)
+        return index_select(self.states, index)
 
     def states2policy(
         self,
@@ -633,16 +641,26 @@ class Batch:
             return states_proxy
         return self.readonly_env.states2proxy(states)
 
-    def get_actions(self) -> List:
+    def get_actions(
+        self,
+        index: Optional[Union[List, Tuple, TensorType, npt.NDArray]] = None,
+    ) -> List:
         """
         Returns the actions in the batch.
+
+        Parameters
+        ----------
+        index: list, tuple, tensor or np.ndarray
+            1-dimentional sequence of bacth indecies for selecting actions, optional.
+            If None (default), all the actions will be returned.
 
         Returns
         -------
         list
-            The list of actions in the batch.
+            The list of actions in the batch with selected index. If index is None,
+            all the actions will be returned.
         """
-        return self.actions
+        return index_select(self.actions, index)
 
     def get_logprobs(
         self, backward: bool = False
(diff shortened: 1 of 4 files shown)


gflownet/utils/batch.py
```diff
@@ -484,6 +486,7 @@ def get_states(
         policy: Optional[bool] = False,
         proxy: Optional[bool] = False,
         force_recompute: Optional[bool] = False,
+        index: Optional[Union[List, Tuple, TensorType, npt.NDArray]] = None,
```
alexhernandezgarcia, line 489: I would call this variable `indices` (plural) instead of `index` (singular).

gflownet/utils/common.py
```diff
@@ -773,3 +774,40 @@ def example_documented_function(arg1, arg2):
     if arg1 == arg2:
         raise ValueError("arg1 must not be equal to arg2")
     return True
+
+
+def index_select(
```
alexhernandezgarcia, line 779: I would call this method `select_indices()`.

gflownet/utils/common.py
```diff
@@ -773,3 +774,40 @@ def example_documented_function(arg1, arg2):
...
+        result = [iterable[idx] for idx in index]
+        if isinstance(iterable, tuple):
+            result = tuple(result)
+        return result
+    elif torch.is_tensor(iterable) or isinstance(iterable, np.ndarray):
+        if isinstance(index, tuple):
+            index = list(index)
+        return iterable[index]
+    else:
+        raise Exception(f"Cannon select elements from {type(iterable)}")
```
alexhernandezgarcia, line 813: Typo:
```suggestion
        raise Exception(f"Cannot select elements from {type(iterable)}")
```

gflownet/utils/batch.py
```diff
@@ -668,22 +686,51 @@ def get_logprobs(
...
+            for idx, lp in enumerate(logprobs):
+                if lp is None:
+                    logprobs_valid[idx] = False
+                    logprobs_clean.append(
+                        tfloat(0.0, device=self.device, float_type=self.float)
+                    )
+                else:
+                    logprobs_clean.append(lp)
+        else:
+            logprobs_clean = logprobs
```
alexhernandezgarcia, line 728: I think there are a bunch of inefficient things here:
- Two for loops that can be avoided
- Tensor creation scalar by scalar
I believe all this can be improved through indexing with the help of numpy, which allows creating an array even though it's got None's in it. I already used this trick [here](https://github.com/alexhernandezgarcia/gflownet/blob/main/gflownet/gflownet.py#L407).
I'm making a suggestion directly here but I'm not testing it so even the syntax may be wrong. It's just for illustration, but please check it yourself.  
```suggestion
        logprobs_clean = np.array(logprobs)
        logprobs_clean[logprobs_clean == None] = 0.0
```

gflownet/utils/batch.py
```diff
@@ -693,7 +740,10 @@ def get_done(self) -> TensorType["n_states"]:
 
     # TODO: check availability one by one as in get_masks
     def get_parents(
-        self, policy: Optional[bool] = False, force_recompute: Optional[bool] = False
+        self,
+        policy: Optional[bool] = False,
+        force_recompute: Optional[bool] = False,
+        index: Optional[Union[List, Tuple, TensorType, npt.NDArray]] = None,
```
alexhernandezgarcia, line 746: Same, `indices` instead of `index` (suggestion).

gflownet/utils/batch.py
AlexandraVolokhova: The problem is that in the general case, some of the elements in the `logprobs ` list are tensors with attached gradients to them and some others are Nones. So, I cannot convert the whole `logprobs ` into a numpy array as I will lose gradients. I also cannot convert it into a tensor because torch throws an error because of the Nones (you cannot do `torch.tensor([tensor, None, tensor, None]`). Therefore, iterating over the list and converting one-by-one is the only possible solution here

gflownet/utils/common.py
AlexandraVolokhova: Changed here: https://github.com/alexhernandezgarcia/gflownet/pull/396/commits/31b05cc805fa7377036c110e4c3a34e70ea133bb

gflownet/utils/batch.py
AlexandraVolokhova: Changed here: https://github.com/alexhernandezgarcia/gflownet/pull/396/commits/c6adc2638656de26f65b84534a3b283c3ab51f7d

gflownet/utils/batch.py
```diff
@@ -1019,6 +1080,7 @@ def _compute_masks_forward(self):
     def get_masks_backward(
         self,
         force_recompute: bool = False,
+        index: Optional[Union[List, Tuple, TensorType, npt.NDArray]] = None,
```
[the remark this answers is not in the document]
AlexandraVolokhova: Changed here: https://github.com/alexhernandezgarcia/gflownet/pull/396/commits/c6adc2638656de26f65b84534a3b283c3ab51f7d

gflownet/utils/batch.py
```diff
@@ -668,22 +686,51 @@ def get_logprobs(
...
-        return logprobs, valids
+        )
+        self.logprobs_backward, self.logprobs_backward_valid = (
+            self._logprobs_to_clean_tensor(
+                self.logprobs_backward, self.logprobs_backward_valid
+            )
+        )
+        self._logprobs_available = True
+
+    def _logprobs_to_clean_tensor(self, logprobs, logprobs_valid):
```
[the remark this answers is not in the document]
AlexandraVolokhova: Done here: https://github.com/alexhernandezgarcia/gflownet/pull/396/commits/e9ee12d1c898436488a2be646a6fdc8722ec2d24

gflownet/utils/batch.py
alexhernandezgarcia: Oh I see! I had not thought of the gradients, but now I think about it, I'm surprised this works at all because the line below would destroy the gradients too!
```
logprobs_clean = tfloat(
    logprobs_clean, float_type=self.float, device=self.device
)
```

alexhernandezgarcia: With this in mind, then how about the following:
```
logprobs_clean = [logp if logp not None else 0.0 for logp in logprobs]
```

gflownet/utils/common.py
alexhernandezgarcia: Thanks for making this change!

gflownet/utils/batch.py
```diff
@@ -668,22 +686,71 @@ def get_logprobs(
...
+            logprob (True) or not (False). When the flag is False, the corresponding logprob
+            is either None or torch.tensor(0.0) detached from the computational graph.
+
+        Returns
+        -------
+        (logprobs_clean, logprobs_valid): a tuple of 1-d tensors
+            - logprobs_clean is a 1-d float tensor with logprobs, where all valid logprobs
+            remain connected to their computational graphs and all non-valid logprobs are zeros.
+            - logprobs_valid is a 1-d boolean tesor with flags indicating whether the corresponding
+            element in the logprobs is a valid logprob (True) or not (False).
```
alexhernandezgarcia, line 735: When multiple variables are returned (indeed, as a tuple), the way to list them in the docstrings is one by one. I will make this change myself.

alexhernandezgarcia: https://github.com/alexhernandezgarcia/gflownet/pull/396/commits/4a67c0d08362bbdc166a753d959e9fcc4c95f211

gflownet/utils/batch.py
alexhernandezgarcia: > No, it will not :) It will call inside the torch.stack over all the one-element tensors inside the logprobs_clean and this operation preserves the connection to the computational graphs
Uhm ok, I had done a quick independent test and I could not see the gradients any more. But it was because I was not calling stack like you noted.

Then my other suggestion would work, right? Namely:
```
logprobs_clean = [logp if logp not None else 0.0 for logp in logprobs]
```

gflownet/utils/batch.py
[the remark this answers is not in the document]
alexhernandezgarcia: Thanks!!!

gflownet/utils/batch.py
```diff
@@ -942,6 +997,7 @@ def get_masks_forward(
         self,
         of_parents: bool = False,
         force_recompute: bool = False,
+        index: Optional[Union[List, Tuple, TensorType, npt.NDArray]] = None,
```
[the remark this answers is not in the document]
alexhernandezgarcia: Thanks! :)

gflownet/utils/batch.py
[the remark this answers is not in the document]
alexhernandezgarcia: Thank you for adding this :)

gflownet/utils/batch.py
AlexandraVolokhova: Unfortunately, this won't work either, because some of the logp elements are one-element tensors, so tfloat will try to stack them with the floats (zeros which you put here) and throw this error: 
Screenshot 2025-09-19 at 18 10 49

alexhernandezgarcia: Ok, my final attempt:
```
logprobs_clean = [logp if logp not None else torch.tensor(0.0) for logp in logprobs]
```
I guess using tfloat instead of just torch.tensor...

AlexandraVolokhova: That would work, but I also need to update `logprobs_valid` simultaneously as I'm creating `logprobs_clean`

alexhernandezgarcia: Ok let's merge it and simplify this later.

AlexandraVolokhova:
@alexhernandezgarcia could you approve this PR? I cannot merge it myself w/o an approval

alexhernandezgarcia (approved):
OK to merge

PR #400, opened 2025-09-27
Replay buffer based on loss

alexhernandezgarcia:
This PR enables using the replay buffer based on the loss instead of just the reward.

The motivation is that it may be beneficial to resample for training trajectories from states that have recently led to high loss. The logic to fill the buffer is slightly different than with the reward because the loss changes as we train.

### DIFF
--- a/gflownet/buffer/base.py
+++ b/gflownet/buffer/base.py
@@ -480,7 +480,7 @@ class BaseBuffer:
 
     # TODO: there may be issues with certain state types
     # TODO: add parameter(s) to control isclose()
-    def _add_greater_update_single_sample(self, sample, trajectory, value, it):
+    def _add_greater_update_single_sample(self, sample, trajectory, value, it) -> bool:
         """
         Adds a single sample (with the trajectory actions and value) to the buffer
         if the state value is larger than the minimum value in the buffer.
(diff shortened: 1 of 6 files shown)


gflownet/buffer/base.py
```diff
@@ -453,6 +474,99 @@ def _add_greater_single_sample(self, sample, trajectory, reward, it):
...
+        states and values:
+            - (C, 5): do not add because the state is not in the buffer and the value
+              is smaller than the minimum in the buffer.
+            - (C, 15): add because the value is larger than the minimum in the buffer.
+              Replace (A, 10).
+            - (B, 25): add because the state is already in the buffer and the value is
+              greater than the currently stored value. Replace (B, 20).
+            - (B, 15): add because the state is already in the buffer and the value is
+              greater than the minimum in the buffer. Replace (B, 20).
+            - (B, 5): do not add because the state although already in the buffer, the
```
AlexandraVolokhova: I'd like to question a bit this logic. Here, I would prefer to put (B, 5) into the buffer (de-fecto updating the value for B), even tho its value is smaller than the current minimum. The reason for this is to maintain the same size of the buffer throughout the training, so it doesn't shrink unexpectedly. If the sample (B, 5) is indeed has too small of a value, it will be dropped at the next iteration when another sample with a bigger value comes in.

gflownet/buffer/base.py
```diff
@@ -453,6 +474,99 @@ def _add_greater_single_sample(self, sample, trajectory, reward, it):
...
+              value is smaller than the minimum in the buffer. However, drop (B, 20)
+              from the buffer.
+
+        Parameters
+        ----------
+        samples : list, tensor, array, dict
```
AlexandraVolokhova:
```suggestion
        sample : list, tensor, array, dict
```

PR #416, opened 2025-12-09
[CTorus] Update from conformer

alexhernandezgarcia:
Updating ctorus class according to conformer branch in the dev repo. 

A report with comparison experiments: https://api.wandb.ai/links/alexandra-volokhova/xda4qpy8


config/experiments/ctorus/sanity.yaml
```diff
@@ -0,0 +1,88 @@
...
+buffer:
+  replay_capacity: 200
+  test:
+    type: grid
+    n: 100
+    output_pkl: test.pkl
+  train: 
+    type: grid
+    n: 100
+
```
alexhernandezgarcia, line 30: Remove extra line break
```suggestion
```

gflownet/gflownet.py
```diff
@@ -1216,19 +1216,57 @@ def get_sample_space_and_reward(self):
...
         proposal_distribution: str = "uniform",
-        epsilon=1e-4,
+        epsilon: float = 1e-4,
     ) -> Union[List, Dict, TensorType["n_samples", "state_dim"]]:
         """
         Rejection sampling with proposal the uniform distribution defined over the
         sample space.
 
-        Returns a tensor in GFloNet (state) format.
+        Returns a tensor in GFlowNet (state) format.
```
alexhernandezgarcia:
```suggestion
        Returns a tensor in environment format.
```

gflownet/gflownet.py
alexhernandezgarcia, line 1: Note that I am not reviewing the correctness of nested sampling

tests/gflownet/envs/test_ctorus.py
```diff
@@ -12,6 +18,16 @@ def env():
     return ContinuousTorus(n_dim=2, length_traj=3)
 
 
+@pytest.fixture
+def proxy():
+    return Torus(normalize=True, alpha=1.0, beta=1.0, device="cpu", float_precision=32)
```
alexhernandezgarcia: I don't think it is a good idea to make this tests of the environment depend on the proxy. Actually, this fixture is only used in the test `tests__unroll_trajectory` in a way that is not necessary. So it should be safe to remove the fixture and the line that uses it.

tests/gflownet/envs/test_ctorus.py
```diff
@@ -103,6 +119,86 @@ def test__sample_actions_batch__not_special_cases(
     assert action_sampled != action_special
 
 
+# Copypaste tests for start_uniform=True
```
alexhernandezgarcia: If these tests are really copy-pastes of other tests using the other env fixture, then the better way to do it to reuse code would be to have just one version of each test and pass the fixture as a parameter using `request`. See an example [here](https://github.com/alexhernandezgarcia/gflownet/blob/main/tests/gflownet/envs/test_choices.py#L58).

alexhernandezgarcia (commented):
I have reviewed all files except `ctorus.py`. My main comments (that IMO require changes) are on the tests.

I will review the environment code later because it seems it is still not working as expected.

gflownet/envs/ctorus.py
```diff
@@ -48,61 +54,109 @@ class ContinuousTorus(GFlowNetEnv):
     vonmises_min_concentration : float
         Minimum value allowed for the concentration parameter of the von Mises
         distributions.
+    exp_vonmises_concentration: bool
+        A flag indicating whether to exponentiate concentrations for von Mises distribution.
+        Default is True.
+    state_space_atol: float
+        Tolerance for comparing states similarity.
+    start_uniform : bool
+        If True, the first step of the trajectory is sampled from the uniform distribution.
```
alexhernandezgarcia:
```suggestion
        If True, the first step of the trajectory is sampled from a uniform distribution.
```

gflownet/envs/ctorus.py
```diff
@@ -48,61 +54,118 @@ class ContinuousTorus(GFlowNetEnv):
...
+            if fixed_distr_params is None:
+                fixed_distr_params = {
+                    "vonmises_mean": 0.0,
+                    "vonmises_concentration": 0.5,
+                }
+            if random_distr_params is None:
+                random_distr_params = {
+                    "vonmises_mean": 0.0,
+                    "vonmises_concentration": 0.001,
+                }
```
alexhernandezgarcia, line 166: In the previous versions, these values were the default values of some parameters of the environments, but now they are hard-coded. I would ideally rather have them as parameters again. If not, I would set them as constants at the beginning of the file for better readability.

gflownet/envs/ctorus.py
```diff
@@ -433,6 +817,76 @@ def _get_max_trajectory_length(self) -> int:
...
+            return stds
+
+    def isclose(
+        self,
+        first_state: List,
+        second_state: List,
+        atol: Optional[float] = None,
+    ) -> bool:
+        """
+        Check if two states are close in the state space.
```
alexhernandezgarcia, line 985: Could you add a note about why this method needs to be overwritten?

gflownet/envs/ctorus.py
```diff
@@ -714,6 +1067,7 @@ def fit_kde(
         bandwidth : float
             The bandwidth of the kernel.
         """
+        # TODO: review if torch2np is needed
```
alexhernandezgarcia: I believe `torch2np()` is not needed because `self.gfn.env.states2kde()` is called on the inputs to `fit_kde()` in the evaluator. Could you please double check this and remove update this accordingly?

gflownet/envs/ctorus.py
```diff
@@ -764,6 +1118,7 @@ def plot_reward_samples(
         """
         if self.n_dim != 2:
             return None
+        # TODO: review if torch2np is needed
```
alexhernandezgarcia: It seems to me that here `torch2np` _is_ needed, which makes me wonder whether we should rethink all this a bit more...

gflownet/envs/ctorus.py
```diff
@@ -835,6 +1190,7 @@ def plot_kde(
...
```
alexhernandezgarcia, line 1226: Copy-pasting from above:

It seems to me that here `torch2np` _is_ needed, which makes me wonder whether we should rethink all this a bit more...

gflownet/envs/ctorus.py
```diff
@@ -146,41 +210,6 @@ def get_action_space(self):
         self.representative_action = tuple([0.0] * self.n_dim)
         return [self.representative_action, self.eos]
 
-    def action2representative(self, action: Tuple) -> Tuple:
```
alexhernandezgarcia, line 149: Review self-note: this method has not changed

gflownet/envs/ctorus.py
```diff
@@ -285,6 +314,361 @@ def get_parents(
         """
         pass
 
+    def action2representative(self, action: Tuple) -> Tuple:
```
alexhernandezgarcia: Review self-note: this method has not changed

gflownet/envs/ctorus.py
```diff
@@ -48,61 +54,118 @@ class ContinuousTorus(GFlowNetEnv):
...
+        self.distr_type = distr_type
         self.n_dim = n_dim
         self.length_traj = length_traj
+        self.state_space_atol = state_space_atol
         # Policy properties
         self.n_comp = n_comp
         self.policy_encoding_dim_per_angle = policy_encoding_dim_per_angle
-        self.vonmises_min_concentration = vonmises_min_concentration
+        if self.distr_type == "diffusion":
+            self.n_params_per_dim = 1
```
alexhernandezgarcia, line 139: I would add a short comment about why diffusion only has one parameter per dimension.

gflownet/envs/ctorus.py
```diff
@@ -48,61 +54,118 @@ class ContinuousTorus(GFlowNetEnv):
...
+                    f"mean and a fixed variance. Received n_comp = {self.n_comp}"
+                )
+            self.sigma_max = np.pi
+            self.sigma_min = 0.01 * np.pi
+            if fixed_distr_params is None:
+                fixed_distr_params = {"means": 0.0, "stds": 0.5}
+            if random_distr_params is None:
+                random_distr_params = {"means": 0.0, "stds": 2 * np.pi}
+        elif self.distr_type == "von_mises":
+            self.n_params_per_dim = 3
```
alexhernandezgarcia, line 156: I would add a short comment about why von Mises has three parameters per dimension.

gflownet/envs/ctorus.py
alexhernandezgarcia, line 228: Review self-note: method reviewed, all good.

gflownet/envs/ctorus.py
```diff
@@ -625,70 +719,329 @@ def get_logprobs(
...
+        if cumulative:
+            return sigmas
+        else:
+            g = sigmas * torch.sqrt(
+                torch.tensor(2 * np.log(self.sigma_max / self.sigma_min))
+            )
+            stds = g * np.sqrt(dt)
+            return stds
+
+    def isclose(
```
alexhernandezgarcia, line 978: Could you please add a note about why this method needs to be overwritten?

gflownet/envs/ctorus.py
```diff
@@ -625,70 +719,329 @@ def get_logprobs(
...
+        ) and int(first_state[-1]) == int(second_state[-1])
 
     def copy(self):
         return deepcopy(self)
 
-    def get_grid_terminating_states(self, n_states: int) -> List[List]:
-        """
-        Samples n terminating states by sub-sampling the state space as a grid, where n
-        / n_dim points are obtained for each dimension.
+    def get_grid_terminating_states(
```
alexhernandezgarcia, line 1015: Review self-note: method reviewed, all good

gflownet/envs/ctorus.py
```diff
@@ -625,70 +719,329 @@ def get_logprobs(
...
+            A list of sampled terminating states.
+        """
+        if n_dim is None:
+            n_dim = self.n_dim
         rng = np.random.default_rng(seed)
-        angles = rng.uniform(low=0.0, high=(2 * np.pi), size=(n_states, self.n_dim))
-        states = np.concatenate((angles, np.ones((n_states, 1))), axis=1)
+        angles = rng.uniform(low=0.0, high=(2 * np.pi), size=(n_states, n_dim))
+        states = np.concatenate(
+            (angles, self.length_traj * np.ones((n_states, 1))), axis=1
```
alexhernandezgarcia, line 1080: Why `self.length_traj` has appeared in the shape and it wasn't before?

gflownet/envs/ctorus.py
```diff
@@ -570,28 +643,49 @@ def sample_actions_batch(
...
             angles_sampled = distr_angles.sample()
             actions_tensor[do_sample] = angles_sampled
-        # Catch special case for backwards backt-to-source (BTS) actions
+
+            # Start from uniform distribution
+            if self.start_uniform:
+                idx = []
+                for i, state in enumerate(states_from):
+                    if state[-1] == 0 and do_sample[i]:
+                        idx.append(i)
```
alexhernandezgarcia: We can avoid this (second) for loop over the states by reusing the tensor `timesteps` computed a few lines before.

alexhernandezgarcia: I would create a variable called something like `do_uniform` out of `timesteps`

gflownet/envs/ctorus.py
```diff
@@ -570,28 +643,49 @@ def sample_actions_batch(
...
+                            device=self.device,
+                        ),
+                        2
+                        * torch.pi
+                        * torch.ones(
+                            len(first_step_idx),
+                            self.n_dim,
+                            dtype=self.float,
+                            device=self.device,
+                        ),
```
alexhernandezgarcia: I would declare a variable for this tensor outside the call to `Uniform()` to make this a but more readable.

gflownet/envs/ctorus.py
```diff
@@ -570,28 +643,49 @@ def sample_actions_batch(
...
-        # Catch special case for backwards backt-to-source (BTS) actions
+
+            # Start from uniform distribution
+            if self.start_uniform:
+                idx = []
+                for i, state in enumerate(states_from):
+                    if state[-1] == 0 and do_sample[i]:
+                        idx.append(i)
+                if len(idx) > 0:
+                    first_step_idx = torch.tensor(idx, device=self.device)
```
alexhernandezgarcia: Actually, you could directly create `first_step_idx` from `timesteps` with `torch.where()` I believe and that would simplify the code quite a bit.

gflownet/envs/ctorus.py
alexhernandezgarcia: See https://github.com/alexhernandezgarcia/gflownet/pull/416/changes#r3648022691

gflownet/envs/ctorus.py
```diff
@@ -625,70 +719,329 @@ def get_logprobs(
...
-                -1, self.n_dim, self.n_comp
+            logprobs[do_sample] = distr.log_prob(actions[do_sample])
+            # Start from uniform distribution
+            if self.start_uniform:
+                idx = []
+                for i, state in enumerate(states_from):
+                    if state[-1] == 0 and do_sample[i]:
+                        idx.append(i)
+                if len(idx) > 0:
+                    first_step_idx = torch.tensor(idx, device=self.device)
```
alexhernandezgarcia: Same comments as for `sample_actions_batch()`

gflownet/envs/ctorus.py
```diff
@@ -625,70 +719,329 @@ def get_logprobs(
...
+                            device=self.device,
+                        ),
+                        2
+                        * torch.pi
+                        * torch.ones(
+                            len(first_step_idx),
+                            self.n_dim,
+                            dtype=self.float,
+                            device=self.device,
+                        ),
```
alexhernandezgarcia: Same comments as for `sample_actions_batch()`

gflownet/envs/ctorus.py
```diff
@@ -625,70 +719,329 @@ def get_logprobs(
...
+                mask_inf = ~angles_allclose(actions_bts_input, actions_bts, atol=1e-6)
+                if torch.any(mask_inf):
+                    # weird, but needed to assign walues to a tensor using 2 masks
+                    logprobs_tmp = logprobs[do_bts]
+                    logprobs_tmp[mask_inf] = -torch.inf
+                    logprobs[do_bts] = logprobs_tmp
+                    warnings.warn(
+                        "Warning: logprobs for invalid back-to-source actions set "
+                        "to -inf."
+                    )
```
alexhernandezgarcia: I don't think the code inside this if statement is needed at all - plus part of it looks like debugging code.

Why is it not needed?

Previously, we have
```
do_sample = torch.all(~mask, dim=1)
```
If any element in the mask is True, then `do_sample` is False the corresponding logprob in `logprobs` will remain equal to 0.0.

What the code inside the if statement is doing is essentially checking whether the last element of the mask is True and in that case set the logprob to 0.0, but it should always be zero already, given the code above.

alexhernandezgarcia: I have actually added this:
```
if torch.any(logprobs[do_bts].flatten() != 0.0):
    import ipdb; ipdb.set_trace()
```
before `logprobs[do_bts] = 0.0` and it never stopped at the breakpoint, which confirms my intuition, I would say.

alexhernandezgarcia: Running this:
```
python train.py +experiments=ctorus/sanity logger.do.online=False
```

gflownet/envs/ctorus.py
```diff
@@ -625,70 +719,329 @@ def get_logprobs(
...
+
+            locations : tensor["n_states", "n_dim", "n_comp"]
+                The locations of the von Mises distributions
+        elif Wrapped Normal distribution is used:
+            means: tensor["n_states", "n_dim"]
+                The means of the wrapped normal distributions.
+            stds: tensor["n_states", "n_dim"]
+                The standard deviations of the wrapped normal distributions. This is
+                not learned, and it comes from a predefined noise schedule. See
+                function convert_timesteps_to_stds.
```
alexhernandezgarcia: This is quite messy, no? How about splitting the method into two, one for von Mises and one for wrapped normal?

alexhernandezgarcia (commented):
I am done reviewing! I have made quite a few cosmetic changes, a few re-implementation changes and I have dropped a bunch of comments with suggestions. Some of the are relatively important to improve the readability and performance of the code. I'll be happy to discuss further.

tests/gflownet/envs/test_ctorus.py
AlexandraVolokhova: Agree, removed it here: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/7a93baf9ee40c88b53a0b49c016ea1ac9ce003b2

tests/gflownet/envs/test_ctorus.py
AlexandraVolokhova: Done: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/41d963a55c90ef61e0f94e50f904fe7b651d9518

gflownet/envs/ctorus.py
AlexandraVolokhova: In terms of hard-coding, I think that there's no big difference between having these dictionaries as default parameters or having them the way I've implemented it: either way they are hard-coded, meaning they are defined directly in the code of the method :) 
My version is better in my opinion (i.e. default is `None`, and `if None: assign hardcoded values`) for two reasons:
1. It's a bad practice to use mutable objects as default values of parameters of a python function (se [here](https://dev.to/aaron_rose_0787cc8b4775a0/pythons-most-famous-gotcha-the-mutable-default-argument-jfo) for an example). It leads to subtle hard-to-catch bugs. 
2. These default values are completely different depending on the `distr_type`, so it's not possible to assign something default which will work for both cases. Here again my version with default is `None`, and `if None: assign hardcoded values` works well because the values in the if-clause are conditioned on the `distr_type`.

As for the constants in the beginning of the file -- I'm not sure it will improve readability. Isn't it easier to see the values directly in the code where they are assigned to the corresponding parameters (in this case `random_distr_params` and 'fixed_distr_params')  instead of having to scroll up to see what were these constant values? I don't feel strongly about it, I can make them constants, but I just really don't see how it improves readability

gflownet/envs/ctorus.py
AlexandraVolokhova: Yes, it's unnecessary here, I removed it: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/28877717693312b5002ac1d6ea47b22b98d255d0

gflownet/envs/ctorus.py
AlexandraVolokhova: It was actually not needed for samples, but needed for rewards, I've updated it here: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/6c8820873395342ad57813b22597d572484aaaa6

gflownet/envs/ctorus.py
alexhernandezgarcia: OK, I agree that having a dictionary as a parameter of a method can lead to issues, but at the same time this is very common too. I'll agree to not do it here though.

However, it does seem important to be able to control these parameters no?

gflownet/envs/ctorus.py
AlexandraVolokhova: Yes, here it is needed, I checked in debugger: `samples` come as a torch tensor to this function

AlexandraVolokhova: I didn't update anything in these plotting methods in this PR, so I didn't put much thought into it tbh

gflownet/envs/ctorus.py
alexhernandezgarcia: Are you sure the changes in this commit are correct? According to the [code in the evaluator](https://github.com/alexhernandezgarcia/gflownet/blob/ctorus-update-from-conformer/gflownet/evaluator/base.py#L599), the first input `samples` to `plot_reward_samples()` is expected to be a list, not a numpy array :/

gflownet/envs/ctorus.py
AlexandraVolokhova: Done here: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/fa612efd008c59f64606a2bc8fd93ff8cc57b714

gflownet/envs/ctorus.py
AlexandraVolokhova: Done here: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/fa612efd008c59f64606a2bc8fd93ff8cc57b714

gflownet/envs/ctorus.py
AlexandraVolokhova: well, it is numpy array for sure, I run it with debugger to check. I guess, evaluator is not correct, `x_sampled` is not a list in this case

gflownet/envs/ctorus.py
AlexandraVolokhova: Well, it's possible to pass any non-None`fixed_distr_params` and `random_distr_params` through the config and they will be assigned to these new values. They are still parameters of the `__init__` method, only default values are hard-coded here.

gflownet/envs/ctorus.py
AlexandraVolokhova: I've extended the docstring to make it clearer: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/990217ec60cb6680183466aa72b0692b0144302e

gflownet/envs/ctorus.py
AlexandraVolokhova: Because there states are supposed to be terminating states, so the step number (the final component of the state) should be `== self.length_traj`. It was incorrect before.

gflownet/envs/ctorus.py
AlexandraVolokhova: Agree, done here: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/e3a7b4c2f570dbc7e81ca66f7738b8f277e89bf8

gflownet/envs/ctorus.py
AlexandraVolokhova: After I added this commit addressing your previous comment the tensor `first_step_idx` is not needed anymore, see here: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/e3a7b4c2f570dbc7e81ca66f7738b8f277e89bf8

gflownet/envs/ctorus.py
AlexandraVolokhova: done here: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/b98c67fce9ca1683e5cbfd23fe93d8e1cd374688

gflownet/envs/ctorus.py
AlexandraVolokhova: Done: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/3fb149b28a6b8cfcd16d5582a850c507ebed6998

gflownet/envs/ctorus.py
AlexandraVolokhova: (Same as before) Done: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/3fb149b28a6b8cfcd16d5582a850c507ebed6998

gflownet/envs/ctorus.py
```diff
@@ -48,61 +55,120 @@ class ContinuousTorus(GFlowNetEnv):
...
+        self.distr_type = distr_type
         self.n_dim = n_dim
         self.length_traj = length_traj
+        self.state_space_atol = state_space_atol
         # Policy properties
         self.n_comp = n_comp
         self.policy_encoding_dim_per_angle = policy_encoding_dim_per_angle
-        self.vonmises_min_concentration = vonmises_min_concentration
+        if self.distr_type == "diffusion":
+            # Diffusion has only one parameter per dimention: the mean of the Gaussian
```
alexhernandezgarcia:
```suggestion
            # Diffusion has only one parameter per dimension: the mean of the Gaussian
```

gflownet/envs/ctorus.py
```diff
@@ -48,61 +55,120 @@ class ContinuousTorus(GFlowNetEnv):
...
+                    f"mean and a fixed variance. Received n_comp = {self.n_comp}"
+                )
+            self.sigma_max = np.pi
+            self.sigma_min = 0.01 * np.pi
+            if fixed_distr_params is None:
+                fixed_distr_params = {"means": 0.0, "stds": 0.5}
+            if random_distr_params is None:
+                random_distr_params = {"means": 0.0, "stds": 2 * np.pi}
+        elif self.distr_type == "von_mises":
+            # von Mises has three parameters per dimention: mixture logit, mean, and concentration
```
alexhernandezgarcia:
```suggestion
            # von Mises has three parameters per dimension: mixture logit, mean, and
            # concentration
```

gflownet/envs/ctorus.py
alexhernandezgarcia: Thanks!

gflownet/envs/ctorus.py
alexhernandezgarcia: OK well, let's ignore this and hopefully we'll redo the evaluator soon.

gflownet/envs/ctorus.py
alexhernandezgarcia: Ok, right. This is fine, thanks.

gflownet/envs/ctorus.py
alexhernandezgarcia: Thanks!

gflownet/envs/ctorus.py
alexhernandezgarcia: Got it 👍

gflownet/envs/ctorus.py
alexhernandezgarcia: Indeed

gflownet/envs/ctorus.py
alexhernandezgarcia: Great, thanks!

gflownet/envs/ctorus.py
alexhernandezgarcia: Thanks! I have made an extra change in [this commit](https://github.com/alexhernandezgarcia/gflownet/pull/416/changes/3be55da61c6f30ba36672a5a30575d0f5fdcd3bf) Please take a look if that's fine with you, feel free to resolve.

gflownet/envs/ctorus.py
alexhernandezgarcia: This might actually help integrating @LenaNehale extensions?

gflownet/envs/ctorus.py
AlexandraVolokhova: looks good to me, thank you!

gflownet/envs/ctorus.py
AlexandraVolokhova: Hmm, I'm not sure about two methods, it's nice to be able to call one method `self._extract_distribution_parameters(policy_outputs, timesteps)` and then get the parameters for the policy regardless self.distr_type. If we have two methods here then we'll need to do if/else every time we need to extract the policy parameters, which is more messy imo.

gflownet/envs/ctorus.py
AlexandraVolokhova: Yeah, you're right. I removed it here https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/02ccfc871b797b228ff3a26de173afc39e939934 and adjusted the tests for this change here  https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/e1add53f29937a665f925b1e83e3ea4050a04699 I've also added one more experiment to the [report](https://api.wandb.ai/links/alexandra-volokhova/xda4qpy8), to check that removing this part doesn't change anything in the run

gflownet/envs/ctorus.py
alexhernandezgarcia: Ah no, I would not do an if else here.

I would keep `self._extract_distribution_parameters(policy_outputs, timesteps)`, but create additional `self._extract_vonmises_parameters(policy_outputs, timesteps)` and `self._extract_diffusion_parameters(policy_outputs, timesteps)` to clean up this method.

alexhernandezgarcia: That way, the details and logic is outsourced to the specific method of each distribution. This would be easier to read and extend, IMO.

gflownet/envs/ctorus.py
alexhernandezgarcia: Great, thanks!

Why was this part of the test removed?
```
        actions_rand = torch.tensor(np.random.rand(1, 2).tolist())
        logprobs_rand = env.get_logprobs(
            policy_output, actions_rand, mask, [state], is_backward=True
        )
        assert ~torch.isfinite(logprobs_rand)
```

AlexandraVolokhova: because it was checking that if action backward from a state` [some_angles, 1]` is not equal to back-to-source (which is, not equal to `some_angles`, so any random action basically), then it gets minus infinity logprobs. But we removed it from the code of the get_logrpobs method (assigning -infinity to such actions), so now logbrobs are always zero regardless of the action (for going backward from a state` [some_angles, 1]` ), so I removed this from the test

alexhernandezgarcia: OK, thanks for explaining!

gflownet/envs/ctorus.py
AlexandraVolokhova: Oh, I see what you mean now. Done here: https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/ff4b898a2140b78a74bc60e644f8925e56efd1b8

alexhernandezgarcia: Thank you!

I have edited the docstring and typing of the method. Please take a look at the commit and if it seems fine, feel free to resolve this comment.

https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/dabe14a03dfdff53de4eeca726dce21f42841b5a

alexhernandezgarcia: [This commit](https://github.com/alexhernandezgarcia/gflownet/pull/416/commits/86f64a512ab879faecaeb5b22bb94eaf312d83be) fixes some indent issues too.

gflownet/envs/ctorus.py
```diff
@@ -625,70 +721,370 @@ def get_logprobs(
...
+        logprobs = torch.sum(logprobs, axis=1)
+        return logprobs
+
+    def _get_distr(
+        self,
+        policy_outputs: TensorType["n_states", "policy_output_dim"],
+        timesteps: TensorType["n_states"],
+        is_backward: bool,
+    ):
+        params = self._extract_distribution_parameters(policy_outputs, timesteps)
```
alexhernandezgarcia, line 781: I have noticed that `params` is not used at all if `self.distr_type == diffusion`. That may be related to the issues we have discussed. Please tackle this in a different PR.

gflownet/envs/ctorus.py
AlexandraVolokhova: Looks good to me, thank you!

gflownet/envs/ctorus.py
AlexandraVolokhova: Oh, that's true, I didn't notice this. I fixed this here (in the other PR): https://github.com/alexhernandezgarcia/gflownet/pull/461/commits/62c62b6b8820123092aeea82958df24097dcd88f

AlexandraVolokhova (approved):

PR #447, opened 2026-04-22
logz estimation from trained GFN

douniashaabankabakibo:
Implement estimation of logZ from sampled terminal states and backward trajectories, and integrates it into the Ising evaluation flow.
This implementation works for Ising but can in principle be used for any other environment.


gflownet/evaluator/base.py
```diff
@@ -327,6 +331,56 @@ def compute_log_prob_metrics(self, x_tt, metrics=None):
             "data": lp_data,
         }
 
+    @torch.no_grad()
+    def compute_logz_metrics(self, metrics=None, sampled_states=None):
+        """
+        Compute an estimated logZ from samples generated by the trained GFlowNet.
```
alexhernandezgarcia, line 337:
```suggestion
        Estimate logZ from samples generated by the trained GFlowNet.
```

gflownet/gflownet.py
```diff
@@ -934,6 +934,141 @@ def estimate_logprobs_data(
         pbar2.close()
         return logprobs_estimates, logprobs_std, probs_std
 
+    @torch.no_grad()
+    def estimate_logz_data(
```
alexhernandezgarcia, line 938: Why the name of the method has `_data` at the end? Is it to mimic `estimate_logprobs_data()`? If so, the name of the latter means "estimate log probabilities of data". In the logZ case, that is not the meaning. I would just call it `estimate_logz()` or `estimate_logz_from_samples()`.

PR #448, opened 2026-04-27
Sequence base environment, DNA and SELFIES

alexhernandezgarcia:
This PR implements a basic SequenceBase environment to facilitate the creation of simple sequence-based environments. As an example, this PR also implements a DNA environment based on SequenceBase.

Note: this code is largely based on an old implementation (more than two years) that had not been pushed to this repo. It seems functional, but it should be tested.

### DIFF
--- a/gflownet/envs/sequences/selfies.py
+++ b/gflownet/envs/sequences/selfies.py
@@ -1,10 +1,11 @@
-"""
-Class to represent SELFIES molecules.
-"""
+"""Class to represent SELFIES molecules."""
 
-from typing import List
+from typing import List, Union
+
+from torchtyping import TensorType
 
 from gflownet.envs.sequences.base import SequenceBase
+from gflownet.utils.common import tlong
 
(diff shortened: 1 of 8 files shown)


gflownet/envs/sequences/selfies.py
```diff
@@ -0,0 +1,63 @@
...
+    "[O]",
+    "[P]",
+    "[Ring1]",
+    "[Ring2]",
+    "[S]",
+]
+PAD_TOKEN = "[nop]"
+
+
+class SelfiesEnv(SequenceBase):
```
alexhernandezgarcia: For consistency with other classes, I would just call the class `Selfies` (without `Env`).

pedromsferraz-mila: Good point, done! [cf2e4a7](https://github.com/alexhernandezgarcia/gflownet/pull/448/commits/cf2e4a7fa48eb87c7936e67feeffded66909beab)

PR #455, opened 2026-06-19
[WIP] Create composite env sequence class

timarni:


gflownet/envs/composite/sequence.py
```diff
@@ -0,0 +1,1034 @@
...
+        # Maximum sequence length / maximum number of elements
+        if max_sequence_length is None:
+            if self._fixed_bag is not None:
+                max_sequence_length = len(self._fixed_bag)
+            else:
+                raise ValueError(
+                    "max_sequence_length must be provided in free-growth mode or "
+                    "random fixed-bag mode."
+                )
+        if self._fixed_bag is not None and len(self._fixed_bag) > max_sequence_length:
```
dcbiton, line 188: In this part, why is len(fixed_bag) < max_sequence_length not allowed?

timarni: Because the Sequence composite env can be used in two configurations:
1) The default and expected mode with `do_random_subenvs` = False and `subenvs` = [], here the sequence is in free-growth (adding as many sub-envs from the possible sub-envs as wanted) and the generated sequence can be length 1, 2, ... , max_seq_len
2) With `subenvs` != [] (or `do_random_subenvs` = True for testing), then the sequence has to organize all the subenvs provided in a sequence and the length of the generated sequence is exactly = len(subenvs) = len(fixed_bag)

gflownet/envs/composite/sequence.py
```diff
@@ -0,0 +1,1385 @@
...
+                eos_tensor = tfloat(self.eos, float_type=self.float, device=self.device)
+                # filter out eos actions
+                is_eos_state = torch.zeros_like(is_meta)
+                is_eos_state[is_meta] = torch.any(actions[is_meta] != eos_tensor, dim=1)
+                if torch.any(is_eos_state):
+                    # remove the eos actions
+                    states_stochastic = [
+                        s for s, f in zip(states_from, is_eos_state) if f
+                    ]
+                    # log(n) correction for multiple states of the parent of the same sequences
```
dcbiton, line 1021: @alexhernandezgarcia @timarni please check if this -log(n) correction is implemented correctly. The function is above in [_get_logprobs_of_same_sequences](https://github.com/alexhernandezgarcia/gflownet/pull/455/changes#diff-063c8c694c6be0ded0af5b1c916b82208db1f48654621612bb39e585f6707fb7R966)

gflownet/envs/composite/sequence.py
```diff
@@ -0,0 +1,1387 @@
...
+        if front_only and end_only:
+            raise ValueError("front_only and end_only cannot both be True.")
+        if envs_unique is None and subenvs is None:
+            raise ValueError(
+                "Both envs_unique and subenvs are None. At least one of the two "
+                "variables must contain a set of environments."
+            )
+
+        self.front_only = front_only
+        self.end_only = end_only
```
alexhernandezgarcia: I would call front and end left and right instead, since it is less ambiguous.

gflownet/envs/composite/sequence.py
```diff
@@ -0,0 +1,1387 @@
...
+        self, policy_outputs: TensorType["n_states", "policy_output_dim"]
+    ):
+        """Returns the last ``n_meta_actions`` columns of the policy outputs."""
+        return policy_outputs[:, -self.n_meta_actions :]
+
+    # ------------------------------------------------------------------ #
+    # Parents
+    # ------------------------------------------------------------------ #
+
+    def get_parents(
```
alexhernandezgarcia, line 577: I think there is a bug in this method.

For the `env_grid` fixture in the tests, we can define the following state:

```
state = {'_active': -1, '_dones': [1, 1, 1, 1], '_envs_unique': [0, 0, 0, 0], '_indices': [3, 2, 1, 0], 0: [0, 0], 1: [0, 1], 2: [0, 0], 3: [1, 1]}
```

If we do `env.get_parents(state)`, it only returns one parent even if both `env.front_only` and `env.end_only` are `False. It should return two parents.

gflownet/envs/composite/sequence.py
```diff
@@ -0,0 +1,1387 @@
...
+-----------------------------
+STATE
+-----
+self.state dict used
+
+Required:
+``_envs_unique``: list of GFlowNetEnv objects (from initialization)
+``_active``: int -1/0
+    - value of -1 = composite or sequence level; 0 = subenvironment is active
+``_dones``: list of int; 0/1 boolean done flag refering to the subenvironments
```
alexhernandezgarcia, line 38: I think we do not need the `dones` key since it is redundant given `_active` and `_indices`.

gflownet/envs/composite/sequence.py
```diff
@@ -0,0 +1,1387 @@
...
+``_indices == [3, 2, 0, 1]``.
+
+-----------------------------
+STATE
+-----
+self.state dict used
+
+Required:
+``_envs_unique``: list of GFlowNetEnv objects (from initialization)
+``_active``: int -1/0
```
alexhernandezgarcia, line 36: I think this flag should take three values:
- `0`: meta-environment is active
- `-1`: left subenv is active
- `1`: right subenv is active

gflownet/envs/composite/sequence.py
alexhernandezgarcia: I have added a test case to highlight this.

gflownet/envs/composite/sequence.py
```diff
@@ -0,0 +1,1387 @@
...
+    # ------------------------------------------------------------------ #
+    # Action space, masks and policy structure
+    # ------------------------------------------------------------------ #
+
+    def get_action_space(self) -> List[Tuple]:
+        """
+        Constructs the list with all possible actions.
+
+        The action space consists of, in this order:
+            - The meta-actions to insert a sub-environment: for each direction (first,
```
alexhernandezgarcia: This is a bit confusing to me. The meta-actions, in my view, should just be:
- Insert on the left
- Insert on the right
From the source state, it is enough to simply allow only one of the two, whichever.

dcbiton: - Done changing the meta-actions into left/right

gflownet/envs/composite/sequence.py
dcbiton: changed to left_only / right_only

PR #461, opened 2026-07-31
[Ctorus] Fix bugs for diffusion

AlexandraVolokhova:
In the #416 the diffusion distribution was added but some things were missing to be able to run it. This PR adds these missing pieces

Important notes about Wrapped normal distribution for the future:
1. In the current implementation there's no normalisation constant, we may need to add it.
2. Maybe we want to keep logprobs computation in the logscale

Comparing diffusion run in this branch with the conformer branch in dev: https://wandb.ai/alexandra-volokhova/gfn_sanity_checks/reports/-31-July-26-Ctorus-fixing-diffusion-bugs--VmlldzoxNzYzMTY5Nw

### DIFF
--- a/config/env/ctorus.yaml
+++ b/config/env/ctorus.yaml
@@ -11,11 +11,7 @@ policy_encoding_dim_per_angle: null
 length_traj: 3
 vonmises_min_concentration: 1e-3
 # Parameters of the fixed policy output distribution
+distr_type: von_mises
 n_comp: 3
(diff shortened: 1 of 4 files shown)


alexhernandezgarcia (approved):
I have not done an in-depth review, but superficially it looks good to merge, after pushing a couple of commits, adjusting the docstrings style and applying black to pass CLI tests.

